PR #8, opened 2019-04-23
gg-17381 jdbc compatibility test failure fixed.

sanpwc:
gg-17381 compatibility test failure fixed

partition result is now serialized/deserialized only in case of proper version.


dspavlov:
The issue is resolved, could you please close the PR as well?

PR #17, opened 2019-04-30
GG-17427

sanpwc:
Individual reconnect for jdbc thin in case of best effort affinity mode.


modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1403,7 +1413,7 @@ private JdbcThinTcpIo cliIo(List<UUID> nodeIds) {
             return txIo;
 
         if (nodeIds == null || nodeIds.isEmpty())
-            return iosArr[RND.nextInt(iosArr.length)];
+            return randomIo();
 
         JdbcThinTcpIo io = null;
```
pavlukhin: Is there a need to call `nodeIds.iterator().next()` couple of lines below? Can we call `nodeIds.get(0)` instead?

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1426,7 +1436,18 @@ private JdbcThinTcpIo cliIo(List<UUID> nodeIds) {
...
 
-        return io != null ? io : iosArr[RND.nextInt(iosArr.length)];
+        return io != null ? io : randomIo();
+    }
+
+    /**
+     * @return random tcpIo
+     */
+    private JdbcThinTcpIo randomIo() {
+        UUID randomUUID = UUID.randomUUID();
```
pavlukhin: `UUID.randomUUID` is a static method and uses a globally shared random number generator (RNG). It can bring an unexpected performance impact on random UUID generation. Can we replace it with private RNG (borrowing ideas from `randomUUID` implementation)?

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -204,30 +213,31 @@ public JdbcThinConnection(ConnectionProperties connProps) throws SQLException {
 
         timer = new Timer("query-timeout-timer");
 
+        connHndTimer = new Timer("connection-handler-timer");
```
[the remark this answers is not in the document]
sanpwc: Reworked with Executors.newScheduledThreadPool(2);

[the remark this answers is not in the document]
sanpwc: Obsolete, cause was reworked with Executors.newScheduledThreadPool(2);

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -148,8 +157,8 @@
     /** Connection properties. */
     private final ConnectionProperties connProps;
 
-    /** Connected. */
-    private volatile boolean connected;
+    /** Connections count. */
+    private volatile AtomicInteger connCnt = new AtomicInteger();
```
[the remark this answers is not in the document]
sanpwc: Sure we can.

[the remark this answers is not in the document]
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -996,7 +1009,7 @@ else if (res.status() != ClientListenerResponse.STATUS_SUCCESS)
 
         JdbcResponse res;
 
-        res = cliIo(null).sendRequest(new JdbcCachePartitionsRequest(Collections.singleton(cacheId)), null);
+        res = cliIo(null).sendRequest(new JdbcCachePartitionsRequest(Collections.singleton(cacheId)),null);
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
sanpwc: Yep, I believe It's an atavism of `Collection<UUID>` that was later reworked to `List<UUID>`.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1060,7 +1074,7 @@ else if (affinityCache.version().compareTo(resAffinityVer) > 0) {
      * @throws SQLException On any error.
      */
     void sendQueryCancelRequest(JdbcQueryCancelRequest req, JdbcThinTcpIo cliIo) throws SQLException {
-        if (!connected)
+        if (connCnt.get() == 0)
```
pavlukhin: Also I wonder why we check the counter to determine if we have working connections or not? Cannot we use `ios.isEmpty()` instead?

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
sanpwc: Yep, benchmarks revealed that there's a serious impact of "out-of-scope" random UUID generator. Whats more, `Random` is much faster than `SecureRandom` and cause we don't need cryptographically strong pseudo random number generator in this particular case, it's perfectly fine to use custom randomUUID() with simple `Random` inside. Fixed.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
sanpwc: Nope, after some changes inside onDisconnect(JdbcThinTcpIo cliIo) we might use something like  `if (affinityAwareness && ios.isEmpty() || !affinityAwareness && singleIo == null)`, however `connCnt.get() == 0 `is faster.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1125,24 +1141,21 @@ public String url() {
     /**
      * Called on IO disconnect: close the client IO and opened statements.
      */
-    private void onDisconnect() {
-        if (!connected)
+    private void onDisconnect(JdbcThinTcpIo cliIo) {
+        if (connCnt.get() == 0)
```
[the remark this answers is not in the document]
sanpwc: Hmmm, changed with `assert connCnt.get() > 0`;

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1659,45 +1747,155 @@ private void connectInAffinityAwarenessMode(HostAndPortRange[] srvs) throws SQLE
...
+                                        continue;
+                                    }
+
+                                    Integer delayRemainder = reconnectionDelaysRemainder.get(sockAddr);
+
+                                    if (delayRemainder != null && delayRemainder != 0)
+                                        continue;
+
+                                    if (closed) {
+                                        maintenanceExecutor.shutdown();
```
pavlukhin: Why do we need to call `shutdown` here?

sanpwc: We should prevent rerunning reconnection task if connection itself is in closing state, so we have a choice either call `connectionsHndScheduledFut.cancel()` or `maintenanceExecutor.shutdown()`. Both options are fine here. The only very minor point to favor shutdown() to cancel is the fact that it will also stop `qryTimeoutScheduledFut` a little bit earlier.

dspavlov:

PR #43, opened 2019-05-20
GG-18542: SQL: Introduce quotas for GroupBy aggregates.

AMashenkov:


modules/h2/src/main/java/org/h2/command/dml/SelectGroups.java
```diff
@@ -119,13 +124,16 @@ public void done() {
...
             if (cursor.hasNext()) {
-                Map.Entry<ValueRow, Object[]> entry = cursor.next();
-                currentGroupByExprData = entry.getValue();
+                curEntry = cursor.next();
+                currentGroupByExprData = curEntry.getValue();
                 currentGroupRowId++;
-                return entry.getKey();
+                return curEntry.getKey();
             }
             return null;
```
ygerzhedovich: seems need to set ``curEntry`` as ``null``

modules/h2/src/main/java/org/h2/command/dml/SelectGroups.java
```diff
@@ -135,6 +143,8 @@ public void remove() {
             cursor.remove();
             currentGroupByExprData = null;
             currentGroupRowId--;
+
+            onUpdate(curEntry.getKey(), curEntry.getValue(), null);
```
ygerzhedovich: is it possible NPE here?

modules/h2/src/main/java/org/h2/command/dml/SelectGroups.java
```diff
@@ -439,4 +469,46 @@ public void nextLazyRow() {
...
+            return;
+
+        long size;
+
+        /* TODO: GG-18628: Estimate Aggregate objects size. */
+        if (row != null && old != null)
+            size = (row.length - old.length) * Constants.MEMORY_OBJECT;
+        else if (old == null) {
+            size = groupKey != null && groupKey.track() ? groupKey.getMemory() : 0;
+            size += Constants.MEMORY_ARRAY + row.length * Constants.MEMORY_OBJECT;
```
ygerzhedovich: can be ``row`` ``null`` here?

modules/h2/src/main/java/org/h2/command/dml/SelectGroups.java
AMashenkov: NPE is impossible here as next() method is always called before this.
Moreover, Cursor.remove will fails with IllegalStateException before NPE happens.

modules/h2/src/main/java/org/h2/command/dml/SelectGroups.java
AMashenkov: No. I've add an assertion.

PR #44, opened 2019-05-20
GG-17442

antonovsergey93:


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -747,6 +748,40 @@ private void notifyLifecycleBeansEx(LifecycleEventType evt) {
         }
     }
 
+    /** */
+    private void ackClassPathElement(File clsPathElement, StringBuilder clsPathContent) {
```
antonovsergey93: Use `org.apache.ignite.internal.util.typedef.internal.SB` instead of `StringBuilder`

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -747,6 +748,40 @@ private void notifyLifecycleBeansEx(LifecycleEventType evt) {
...
+        if (clsPathElement.isDirectory()) {
```
antonovsergey93: You should print only files in `clsPathElement` if it's directory.

dspavlov:

PR #46, opened 2019-05-21
GG-18877

sergey-chugunov-1985:
tcp-comm-worker thread should update heartbeats when recovering tcp client in internal loops


dspavlov:

PR #49, opened 2019-05-22
NBL-3: CPP: Implement minimal Cluster API

FedorUporov:


modules/platforms/cpp/core-test/src/teamcity_boost.cpp
```diff
@@ -80,6 +80,7 @@ class TeamcityBoostLogFormatter : public boost::unit_test::unit_test_log_formatt
     virtual void log_start(std::ostream&, boost::unit_test::counter_t);
     virtual void log_finish(std::ostream&);
     virtual void log_build_info(std::ostream&);
+    virtual void log_build_info(std::ostream&, bool);
```
isapego: Why was it added?

FedorUporov: > Why was it added?

I used boost 1.70.0, the signature of function virtual void log_build_info(std::ostream&)
(see here: https://github.com/boostorg/test/blob/boost-1.70.0/include/boost/test/unit_test_log_formatter.hpp#L144) was changed comparing to version 1.69.0:
https://github.com/boostorg/test/blob/boost-1.69.0/include/boost/test/unit_test_log_formatter.hpp#L143

So the compilation error had place and it was needed to fix it.

modules/platforms/cpp/core/src/impl/cluster/cluster_group_impl.cpp
```diff
@@ -48,6 +84,29 @@ namespace ignite
...
+
+                return SP_ClusterGroupImpl(new ClusterGroupImpl(GetEnvironmentPointer(), target));
+            }
+
+            SP_ClusterGroupImpl ClusterGroupImpl::ForDataNodes(std::string cacheName)
+            {
+                return ForCacheNodes(cacheName, Command::FOR_DATA);
+            }
+
             SP_ClusterGroupImpl ClusterGroupImpl::ForServers()
```
isapego: Check this method please - it may really work as "ForCpp" for historical reasons. Now we keep lazy-initialized value in Ignite which this method returns. Review it - maybe we need to change this logic.

modules/platforms/cpp/core/src/impl/cluster/cluster_group_impl.cpp
```diff
@@ -95,6 +181,49 @@ namespace ignite
 
                 return SP_ComputeImpl(new compute::ComputeImpl(GetEnvironmentPointer(), computeProc));
             }
+
+            std::vector<SP_ClusterNodeImpl> ClusterGroupImpl::RefreshNodes()
+            {
+                long oldTopVer = 0;
```
isapego: I believe, we should use `topVer` as initial value of `oldTopVer` here

modules/platforms/cpp/core/src/impl/cluster/ignite_cluster_impl.cpp
```diff
@@ -0,0 +1,56 @@
...
+using namespace ignite::jni::java;
+using namespace ignite::impl::cluster;
+
+namespace ignite
+{
+    namespace impl
+    {
+        namespace cluster
+        {
+            IgniteClusterImpl::IgniteClusterImpl(SP_ClusterGroupImpl impl) :
```
isapego: This is weird. Is it the same in .NET?

modules/platforms/cpp/core/src/impl/ignite_environment.cpp
```diff
@@ -264,6 +290,19 @@ namespace ignite
             return res;
         }
 
+        void  ClusterNodesHolder::AddNode(SP_ClusterNodeImpl node)
+        {
+            nodes[node.Get()->Id()] = node;
```
isapego: Not thread-safe.

modules/platforms/cpp/core/src/impl/ignite_environment.cpp
```diff
@@ -87,6 +91,18 @@ namespace ignite
             };
         };
 
+        class ClusterNodesHolder
```
isapego: I've seen one class like this already

modules/platforms/cpp/core/src/impl/cluster/cluster_group_impl.cpp
```diff
@@ -95,6 +170,48 @@ namespace ignite
...
+                interop::InteropInputStream inStream(memOut.Get());
+                binary::BinaryReaderImpl reader(&inStream);
+
+                bool wasUpdated = reader.ReadBool();
+                if (wasUpdated)
+                {
+                    topVer = reader.ReadInt64();
+                    int cnt = reader.ReadInt32();
+
+                    SP_ClusterNodes newNodes(new std::vector<ClusterNode>());
```
isapego: We should reserve capacity in a vector here as know its full size.

modules/platforms/cpp/core/src/impl/cluster/cluster_group_impl.cpp
```diff
@@ -95,6 +170,48 @@ namespace ignite
...
             }
+
+            std::vector<ClusterNode> ClusterGroupImpl::RefreshNodes()
+            {
+                SharedPointer<interop::InteropMemory> memIn = GetEnvironment().AllocateMemory();
+                SharedPointer<interop::InteropMemory> memOut = GetEnvironment().AllocateMemory();
+                interop::InteropOutputStream out(memIn.Get());
+                binary::BinaryWriterImpl writer(&out, GetEnvironment().GetTypeManager());
+
+                CsLockGuard mtx(nodesLock);
```
isapego: Locking scope is too large here. We only need to hold lock when updating local values - `topVer` and `nodes`.

modules/platforms/cpp/core/src/impl/cluster/cluster_group_impl.cpp
```diff
@@ -95,6 +170,48 @@ namespace ignite
...
+                    for (int i = 0; i < cnt; i++)
+                    {
+                        SP_ClusterNodeImpl impl = GetEnvironment().GetNode(reader.ReadGuid());
+                        if (impl.IsValid())
+                            newNodes.Get()->push_back(ClusterNode(impl));
+                    }
+
+                    nodes = newNodes;
+
+                    return *nodes.Get();
```
isapego: This return is not needed.

PR #58, opened 2019-05-24
GG-18615

sanpwc:
Failover connections support for JDBC thin driver implemented.


modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -865,23 +875,31 @@ JdbcResultWithIo sendRequest(JdbcRequest req) throws SQLException {
...
+
+                while (retryAttemptsLeft > 0) {
+                    JdbcThinTcpIo cliIo = null;
+
+                    ensureConnected();
+
+                    try {
+                        cliIo = (stickyIo == null || !stickyIo.connected()) ? cliIo(calculateNodeIds(req)) : stickyIo;
 
                 if (stmt != null && stmt.requestTimeout() != NO_TIMEOUT) {
```
AMashenkov: Style. Let's make left padding wider.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1698,6 +1731,49 @@ private void updateAffinityCache(JdbcQueryExecuteRequest qryReq, JdbcResponse re
...
+    /**
+     * Calculates query retries count for given {@param req}.
+     *
+     * @param stickyIo sticky connection, if any.
+     * @param req Jdbc request.
+     *
+     * @return retries count.
+     */
+    private int calculateRetryAttemptsCount(JdbcThinTcpIo stickyIo, JdbcRequest req) {
+        if (! affinityAwareness)
```
AMashenkov: Extra space in "!afinityAwareness"

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1698,6 +1731,49 @@ private void updateAffinityCache(JdbcQueryExecuteRequest qryReq, JdbcResponse re
...
+            req.type() == JdbcRequest.META_SCHEMAS ||
+            req.type() == JdbcRequest.CACHE_PARTITIONS)
+            return DFLT_RETRIES_CNT;
+
+        if (req.type() == JdbcRequest.QRY_EXEC) {
+            JdbcQueryExecuteRequest qryExecReq = (JdbcQueryExecuteRequest)req;
+
+            String trimmedQry = qryExecReq.sqlQuery().trim();
+
+            char[] qryCArr = trimmedQry.toCharArray();
```
AMashenkov: There is no need to have a copy char[]. String.indexOf() can be used instead.
BTW, it is possible to get false positive result if ';' is inside some string const.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
sanpwc: >There is no need to have a copy char[]. String.indexOf() can be used instead.
Yep, fixed.
>BTW, it is possible to get false positive result if ';' is inside some string const.
That's true. We either should have fully implemented parser on client side or use very simple rules with such false-positive effects. During the discussion, it was decided to start with the simpler version.

dspavlov:

PR #66, opened 2019-05-27
 GG-19086 Implemented support for SSL for All-in-One mode

akuznetsov-gridgain:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/config/ServerPortCustomizer.java
```diff
@@ -25,8 +26,12 @@
  */
 @Component
 public class ServerPortCustomizer implements EmbeddedServletContainerCustomizer {
+    /** */
+    @Value("${server.port:3000}")
```
nva: Please revert this

modules/web-console/web-console-server/src/main/resources/ignite-config.xml
```diff
@@ -52,6 +52,8 @@
             </bean>
         </property>
 
+        <property name="connectorConfiguration"><null/></property>
```
nva: Please revert this

PR #74, opened 2019-05-30
GG-19136 Metastore update skipped triggering of failure handling on i…

akalash:
…nterrupt exception


dspavlov:

PR #86, opened 2019-06-03
GG-19068 CacheEntryEventType.Removed is not being rised

ashapkin:


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformUtils.java
```diff
@@ -611,6 +612,23 @@ private static void writeCacheEntryEvent(BinaryRawWriterEx writer, CacheEntryEve
...
+     * Write event type to the writer.
+     * @param writer Writer.
+     * @param evtType Type of event.
+     */
+    private static void writeEventType(BinaryRawWriterEx writer, EventType evtType) {
+        switch (evtType){
+            case CREATED: writer.writeByte((byte) 0); break;
+            case UPDATED: writer.writeByte((byte) 1); break;
+            case REMOVED: writer.writeByte((byte) 2); break;
+            case EXPIRED: writer.writeByte((byte) 3); break;
```
ashapkin: What do you think about adding this event to the .NET side? Or should we consider Removed and Expired events as the same one? I created GG-19207 ticket btw.

isapego: I'm not sure. Maybe someone from core team would help here.

ashapkin:
This closes #86

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformUtils.java
ashapkin: We discussed this with @ptupitsyn briefly and decided to port this event too

PR #89, opened 2019-06-04
GG-18720 Implement cluster-wide propagation of RollingUpgadeStatus vi…

ingvard:
…a Distributed Metastorage


modules/core/src/main/java/org/apache/ignite/internal/IgniteNodeAttributes.java
```diff
@@ -26,6 +26,9 @@
     /** Node compound version. */
     public static final String ATTR_BUILD_VER = ATTR_PREFIX + ".build.ver";
 
+    /** Allows to override {@link #ATTR_BUILD_VER} by adding this attribute in the user attributes. */
```
sk0x50: Well, it does not look a good idea to me, adding a new node attribute to `IgniteNodeAttributes` for testing purposes only. I think that required behavior can be achieved by specifying a custom `DiscoverySpi` as follows:
```
public class RollingUpgradeSelfTest extends GridCommonAbstractTest {
    /** {@inheritDoc} */
    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
        IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);

        ...

        if (nextVer != null) {
            final String nextVersion = nextVer;

            TcpDiscoverySpi testSpi = new TcpDiscoverySpi() {
                /** {@inheritDoc} */
                @Override public void setNodeAttributes(Map<String, Object> attrs, IgniteProductVersion ver) {
                    super.setNodeAttributes(attrs, IgniteProductVersion.fromString(nextVersion));
                }
            };

            cfg.setDiscoverySpi(testSpi);
        }

        return cfg;
    }
}
```

modules/core/src/main/java/org/apache/ignite/internal/processors/ru/DistributedRollingUpgradeStatus.java
```diff
@@ -0,0 +1,39 @@
...
+ */
+
+package org.apache.ignite.internal.processors.ru;
+
+import org.apache.ignite.internal.processors.configuration.distributed.DistributedProperty;
+
+/**
+ * The wrapper of {@link RollingUpgradeStatus} for the distributed metastorage binding.
+ */
+public class DistributedRollingUpgradeStatus extends DistributedProperty<RollingUpgradeStatus> {
```
sk0x50: It seems that this class is not used by Ignite code and can be moved to GG.

modules/core/src/main/java/org/apache/ignite/internal/processors/ru/RollingUpgradeProcessor.java
```diff
@@ -26,17 +26,17 @@
      *
      * @param enable {@code true} in order to enable rolling upgrade mode.
      */
-    public RollingUpgradeModeChangeResult setRollingUpgradeMode(boolean enable);
+    RollingUpgradeModeChangeResult setMode(boolean enable);
```
sk0x50: please add 'public' modifier it is required by code style checker.

modules/core/src/main/java/org/apache/ignite/internal/processors/ru/DistributedRollingUpgradeStatus.java
```diff
@@ -0,0 +1,39 @@
...
+    }
+
+    /**
+     * @param name Name of property.
+     * @return Property detached from processor.(Distributed updating are not accessable).
+     */
+    public static DistributedRollingUpgradeStatus detachedProperty(String name) {
+        return new DistributedRollingUpgradeStatus(name);
+    }
+}
```
sk0x50: please add a new line here.

PR #99, opened 2019-06-05
NBL-2: CPP: Demo for Cluster API

FedorUporov:


modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
```diff
@@ -0,0 +1,153 @@
...
+        Ignite client = Ignition::Start(cfgs[3], "Client");
+
+        std::cout << std::endl;
+        std::cout << ">>> Cluster compute example started." << std::endl;
+        std::cout << std::endl;
+
+        // Get binding instances and register our classes as a compute functions.
+        node1.GetBinding().RegisterComputeFunc<PrintMsg>();
+        node2.GetBinding().RegisterComputeFunc<PrintMsg>();
+        node3.GetBinding().RegisterComputeFunc<PrintMsg>();
```
isapego: I believe, this function should be registered for the client node as well.

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
```diff
@@ -0,0 +1,153 @@
...
+
+        // Create cluster groups splitted up by demo attribute value.
+        ClusterGroup localGroup = client.GetCluster().ForAll();
+        ClusterGroup group1 = localGroup.ForAttribute("DemoAttribute", "Value0");
+        ClusterGroup group2 = localGroup.ForAttribute("DemoAttribute", "Value1");
+
+        // Making calls.
+        client.GetCompute(group1).Broadcast(PrintMsg("DemoAttribute=Value0"));
+        client.GetCompute(group2).Broadcast(PrintMsg("DemoAttribute=Value1"));
+
```
isapego: Insert `std::cin.get()` here, to allow user wait for all compute jobs to execute before shutting down a cluster.

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
FedorUporov: Added.

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
FedorUporov: Added additional user message, if I got your idea correctly.

PR #100, opened 2019-06-06
GG-19219 Use stable CE for build web-console.

nva:


modules/web-console/web-console-server/pom.xml
```diff
@@ -36,6 +36,7 @@
     <properties>
         <commons.compress.version>1.18</commons.compress.version>
         <spring.boot.version>1.5.20.RELEASE</spring.boot.version>
+        <swagger2.version>2.9.2</swagger2.version>
```
vveider: Can this properties be moved to `parent/pom.xml` as well?

nva: Moved

vveider:
Looks good now (if it works this way).

PR #120, opened 2019-06-10
GG-19307 Use different ports for internal node.

nva:


modules/web-console/web-console-server/src/main/resources/ignite-config.xml
```diff
@@ -18,11 +18,29 @@
...
        xmlns:xsi="http://www.w3.org/2001/XMLSchema-instance"
-       xsi:schemaLocation="http://www.springframework.org/schema/beans
-                           http://www.springframework.org/schema/beans/spring-beans.xsd">
-    <bean class="org.apache.ignite.configuration.IgniteConfiguration">
-        <property name="clientMode" value="#{new Boolean('${client.mode:false}')}"/>
+       xmlns:util="http://www.springframework.org/schema/util"
+       xsi:schemaLocation="http://www.springframework.org/schema/beans http://www.springframework.org/schema/beans/spring-beans.xsd
+                           http://www.springframework.org/schema/util http://www.springframework.org/schema/util/spring-util.xsd">
+
+    <bean class="org.springframework.beans.factory.config.MethodInvokingFactoryBean">
```
akuznetsov-gridgain: I think we need a comment for this section, like this: "Configure REST API port to 60080".

modules/web-console/web-console-server/src/main/resources/ignite-config.xml
```diff
@@ -45,8 +63,17 @@
             </bean>
         </property>
 
+        <property name="connectorConfiguration">
+            <bean class="org.apache.ignite.configuration.ConnectorConfiguration">
+                <property name="port" value="${random.int[10000,65000]}"/>
```
akuznetsov-gridgain: Why to use random port? I think this can cause "strange" from user point error. Lets just use value that will differ from default one "11211", for example: "22122"?

modules/web-console/web-console-server/src/main/resources/ignite-config.xml
```diff
@@ -45,8 +63,17 @@
...
+            </bean>
+        </property>
+
+        <property name="clientConnectorConfiguration"><null/></property>
+
         <property name="communicationSpi">
             <bean class="org.apache.ignite.spi.communication.tcp.TcpCommunicationSpi">
+                <property name="localPort" value="${random.int[10000,65000]}"/>
```
akuznetsov-gridgain: Same situation, lets simply use not default "47100", but "48100", for example.

modules/web-console/web-console-server/src/main/resources/ignite-config.xml
```diff
@@ -95,5 +122,17 @@
         <property name="failureHandler">
             <bean class="org.apache.ignite.failure.NoOpFailureHandler"/>
         </property>
+
+        <property name="discoverySpi">
+            <bean class="org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi">
+                <property name="localPort" value="${random.int[10000,65000]}"/>
```
akuznetsov-gridgain: Same as above, lets use not default "47500", but "48500" for example?

PR #122, opened 2019-06-11
GG-19376 Fixed sending topology after the cluster was stopped.

nva:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/handlers/ClustersWatcher.java
```diff
@@ -109,6 +108,21 @@
...
+     *
+     * @param ses Websocket session.
+     * @param tops Topologies.
+     */
+    private void sendTopology(Session ses, List<TopologySnapshot> tops) {
+        try {
+            send(ses, new WebSocketEvent(CLUSTER_TOPOLOGY, tops));
+        }
+        catch (Throwable e) {
+            log.error("Failed to send 'Cluster disconnected' event to server", e);
```
akuznetsov-gridgain:
```suggestion
            log.error("Failed to send topology to server", e);
```

PR #128, opened 2019-06-11
GG-19389: Code-generated GridQueryKillResponse and GridQueryKillRequest.

pavel-kuznetsov:
This fixes rolling upgrade issues.


modules/codegen/src/main/java/org/apache/ignite/codegen/MessageCodeGenerator.java
```diff
@@ -43,8 +43,8 @@
 import org.apache.ignite.internal.GridDirectTransient;
 import org.apache.ignite.internal.IgniteCodeGeneratingFail;
 import org.apache.ignite.internal.processors.affinity.AffinityTopologyVersion;
-import org.apache.ignite.internal.processors.cache.mvcc.DeadlockProbe;
-import org.apache.ignite.internal.processors.cache.mvcc.ProbedTx;
+import org.apache.ignite.internal.processors.query.messages.GridQueryKillRequest;
+import org.apache.ignite.internal.processors.query.messages.GridQueryKillResponse;
```
ygerzhedovich: seems it should be reverted

PR #138, opened 2019-06-13
GG-19218 Invalidate from cache old topology snapshots.

nva:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/socket/WebSocketsManager.java
```diff
@@ -55,9 +61,16 @@
  * Web sockets manager.
  */
 @Service
+@EnableScheduling
 public class WebSocketsManager {
     /** */
-    private static final Logger log = LoggerFactory.getLogger(WebSocketsManager.class);
+    private static final IgniteLogger log = new Slf4jLogger(LoggerFactory.getLogger(WebSocketsManager.class));
```
akuznetsov-gridgain: Why do we need IgniteLogger  ?

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/socket/WebSocketsManager.java
```diff
@@ -80,11 +93,13 @@
...
      * Default constructor.
      */
-    public WebSocketsManager() {
+    public WebSocketsManager(Ignite ignite) {
         agents = new ConcurrentLinkedHashMap<>();
         browsers = new ConcurrentHashMap<>();
         clusters = new ConcurrentHashMap<>();
         requests = new ConcurrentHashMap<>();
+
+        new IgniteThread(new ClusterTimeoutWorker(ignite.name(), "session-timeout-worker", log)).start();
```
akuznetsov-gridgain: I do not like this. How about to use Spring "@Scheduled" (or something like this) instead?

modules/web-console/web-console-server/pom.xml
```diff
@@ -136,6 +136,12 @@
             <version>${web-console.ignite.version}</version>
         </dependency>
 
+        <dependency>
```
akuznetsov-gridgain: Do we need this dependency now?

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/Application.java
```diff
@@ -24,7 +24,6 @@
  * Web console launcher.
  */
 @SpringBootApplication
-@EnableScheduling
```
akuznetsov-gridgain: Not sure if it is good idea to move this annotation to service.
We may need scheduling in other places later.
I would keep it here.

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/socket/WebSocketsManager.java
```diff
@@ -55,10 +56,14 @@
  * Web sockets manager.
  */
 @Service
+@EnableScheduling
```
akuznetsov-gridgain: I think it is better to have this annotation on Application.

modules/web-console/web-console-server/pom.xml
nva: no

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/socket/WebSocketsManager.java
nva: Application is entry point. If you create another module that depends this - scheduling will work only after you add this anotation to new Application

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/Application.java
nva: I think will be better to create separate configuration.

PR #145, opened 2019-06-14
GG-19477 WC backend throws IllegalStateException + NPE on closing Demo mode

nva:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/AbstractHandler.java
```diff
@@ -52,6 +52,9 @@
 
             handleEvent(ws, evt);
         }
+        catch (NullPointerException | IOException e) {
+            log.warn("Failed to process message [session=" + ws + ", msg=" + msg + "errMsg=" + e.toString() + "]");
```
akuznetsov-gridgain: should we remove "wc" from various maps?

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/socket/BrowsersHandler.java
```diff
@@ -135,6 +135,11 @@ protected Account extractAccount(WebSocketSession ws) {
                     throw new IllegalStateException("Unknown event: " + evt);
             }
         }
+        catch (IllegalStateException e) {
+            log.warn(e.toString());
```
akuznetsov-gridgain: should we remove "wc" from various maps?

PR #155, opened 2019-06-17
GG-18998 Remove instances of $rootScope use

Klaster1:


Klaster1:
@nva I've merged master in and resolved all conflicts. `User.created$` still has to be used, but I had to move it to another method.

modules/web-console/frontend/app/components/dialog-admin-create-user/controller.ts
```diff
@@ -65,8 +66,8 @@ export class DialogAdminCreateUser {
         this.loading.start('createUser');
 
         this.AdminData.registerUser(this.data)
-            .then(() => {
-                this.$root.$broadcast('userCreated');
+            .then((wtf) => {
```
nva: wtf?

PR #168, opened 2019-06-19
GG-19546 Fixed creation of user by admin.

nva:


modules/web-console/frontend/app/components/dialog-admin-create-user/controller.ts
```diff
@@ -70,10 +70,10 @@ export class DialogAdminCreateUser {
                 this.IgniteMessages.showInfo(`User ${this.data.email} created`);
                 this.close({});
             })
-            .catch((res) => {
+            .catch((data) => {
```
akuznetsov-gridgain: I would rename to "err", not to "data".

modules/web-console/frontend/app/core/admin/Admin.data.js
```diff
@@ -106,6 +106,6 @@ export default class IgniteAdminData {
     registerUser(userInfo) {
         return this.$http.put('/api/v1/admin/users', userInfo)
             .then(({ data }) => data)
-            .catch(this.Messages.showError);
+            .catch(({data}) => {throw data;});
```
akuznetsov-gridgain: I would use "err" instead of "data".

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/repositories/AccountsRepository.java
```diff
@@ -55,7 +55,7 @@ public AccountsRepository(Ignite ignite, TransactionManager txMgr) {
         this.txMgr = txMgr;
 
         accountsTbl = new Table<Account>(ignite, "accounts")
-            .addUniqueIndex(Account::getUsername, (acc) -> "Account with email '" + acc.getUsername() + "' already registered")
+            .addUniqueIndex(a -> a.getUsername().trim().toLowerCase(), (acc) -> "Account with email '" + acc.getUsername() + "' already registered")
```
akuznetsov-gridgain: I think it make sense to introduce function "normalizeEmail(string)" and use this function not only here (in index), but also when we save account.

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/services/AccountsService.java
```diff
@@ -246,16 +246,6 @@ public Account save(UUID accId, ChangeUserRequest changes) {
         try (Transaction tx = txMgr.txStart()) {
             Account acc = accountsRepo.getById(accId);
 
-            String oldEmail = acc.getEmail();
-            String newEmail = changes.getEmail();
-
-            if (!oldEmail.equals(newEmail)) {
-                Account accByEmail = accountsRepo.getByEmail(oldEmail);
```
akuznetsov-gridgain: Why this code was removed? How we can protect that user "A" will not set email as user "B"?

nva: All checks will be done by index on save

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/repositories/AccountsRepository.java
nva: This logic is used only in index.

PR #182, opened 2019-06-21
GG-17407

RealZheka:


RealZheka:
> 
> 
> Do you need this annotation? What happens if you remove it?

It is used to to mark entity as Sql entity in our system. It is necessary to use it to reproduce initial problem described in jira ticket.

RealZheka:
> 
> 
> I don't really understand this change. What's a name() and fullName()?

name() is just a name of the entity, fullName() includes parent's name. Full name will be unique for nested entities so this causes problems described in the ticket.

PR #183, opened 2019-06-21
GG-19776 Send "agent:status" only if cluster changed.

akuznetsov-gridgain:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/socket/WebSocketsManager.java
```diff
@@ -207,7 +207,8 @@ public void sendToNode(WebSocketSession ws, String clusterId, WebSocketEvent evt
     protected void updateTopology(WebSocketSession wsAgent, TopologySnapshot oldTop, TopologySnapshot newTop) {
         AgentDescriptor desc = agents.get(wsAgent);
 
-        updateClusterInBrowsers(desc.accIds);
+        if (newTop.changed(oldTop))
```
nva: This check should be in processTopologyUpdate

PR #186, opened 2019-06-21
GG-19246 Need to reduce stack trace output in case manual change with…

ingvard:
… enable auto-adjust.


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -300,13 +304,13 @@ public int execute(List<String> rawArgs) {
                 return EXIT_CODE_CONNECTION_FAILED;
             }
 
-            if (X.hasCause(e, IllegalArgumentException.class)) {
-                IllegalArgumentException iae = X.cause(e, IllegalArgumentException.class);
+            if (X.hasCause(e, VisorIllegalStateException.class)) {
+                VisorIllegalStateException iae = X.cause(e, VisorIllegalStateException.class);
```
ibessonov: Please rename, "iae" abbreviation is not relevant anymore.

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -300,13 +304,13 @@ public int execute(List<String> rawArgs) {
...
 
-                logger.severe("Check arguments. " + CommandLogger.errorMessage(iae));
+                logger.severe(CommandLogger.errorMessage(iae));
                 logger.info("Command [" + commandName + "] finished with code: " + EXIT_CODE_INVALID_ARGUMENTS);
```
ibessonov: Please log EXIT_CODE_ILLEGAL_SATE_ERROR if you return it later.

PR #188, opened 2019-06-21
GG-19744 Reading of different rest and task result format.

vsisko:


modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -475,7 +478,7 @@ export default class AgentManager {
 
                         const data = await this.pool.postMessage({payload: res.data, useBigIntJson});
 
-                        return data.result ? data.result : data;
+                        return event === EVENT_REST ? data : data.result;
```
akuznetsov-gridgain: May be we should instead always send data from Web Agent as "data.result"?
@nva What do you think?

vsisko: Implemented changes that was supposed by @nva.

vsisko:
Implemented changes that was supposed by @nva.

PR #192, opened 2019-06-24
GG-18500 Possibility to enable loading of example schemas for Oracle database.

vsisko:


vsisko:
We show that flag only for Oracle database.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/handlers/DatabaseHandler.java
```diff
@@ -172,7 +172,7 @@ public DbSchema collectDbSchemas(WebSocketEvent evt) throws SQLException {
                 catalog = parts.length > 0 ? parts[parts.length - 1] : "NONE";
             }
 
-            Collection<String> schemas = dbMetaReader.schemas(conn);
+            Collection<String> schemas = dbMetaReader.schemas(conn, args.getBoolean("importSamples"));
```
akuznetsov-gridgain: Lets use args.getBoolean("importSamples", false) ?

vsisko: Added default argument value.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/db/dialect/MySQLMetadataDialect.java
```diff
@@ -39,12 +39,13 @@
     }
 
     /** {@inheritDoc} */
-    @Override public Collection<String> schemas(Connection conn) throws SQLException {
+    @Override public Collection<String> schemas(Connection conn, boolean importSamples) throws SQLException {
         Collection<String> schemas = new ArrayList<>();
 
         ResultSet rs = conn.getMetaData().getCatalogs();
```
akuznetsov-gridgain: The only difference between  MySQLMetadataDialect  and JdbcMetadataDialect is this line:
JdbcMetadataDialect: ResultSet rs = conn.getMetaData().getSchemas();
MySQLMetadataDialect : ResultSet rs = conn.getMetaData().getCatalogs();

Almost the same for OracleMetadataDialect.

May be it make sense to refactor as method that will return collection strings with schemas names?

vsisko: Removed code duplication on collection of schemas.

PR #193, opened 2019-06-24
NBL-20: Node.js thin: Packet name and description should be changed for GGCE

ragrus-nbl:
Please, note:
1) The install instruction in the main readme is not valid until the renamed package is published
2) The test configuration in TeamCity should be corrected to make tests pass because it contains the old name of the package


ragrus-nbl:
@isapego License has been renamed

PR #196, opened 2019-06-25
GG-20249 Extract message from exception cause stack trace.

vsisko:


modules/web-console/frontend/app/services/ErrorParser.service.js
```diff
@@ -17,6 +17,8 @@
 import isEmpty from 'lodash/isEmpty';
 import {nonEmpty} from 'app/utils/lodashMixins';
 
+const causeStr = 'Caused by: ';
```
akuznetsov-gridgain: Use UPPER_CASE instead.

modules/web-console/frontend/app/services/Messages.service.js
```diff
@@ -42,7 +42,7 @@ export default function factory($alert, errorParser) {
 
         const title = err ? errorMessage(message, err) : errorMessage(null, message);
 
-        msgModal = $alert({type, title, duration});
+        msgModal = $alert({type, title, duration: duration + ((title.match(/<li>/g) || []).length * 3)});
```
akuznetsov-gridgain: Lets think about better solution. I do not like regex parsing.

modules/web-console/frontend/app/services/ErrorParser.service.js
vsisko: Fixed constant name.

modules/web-console/frontend/app/services/Messages.service.js
vsisko: Implemented returning of metadata information of error parsing.

modules/web-console/frontend/app/services/ErrorParser.service.ts
```diff
@@ -58,20 +60,39 @@ export default class {
...
+                    const msgEnd = err.message.indexOf('[', msgStart);
+                    const cause = err.message.substring(msgStart, msgEnd >= 0 && msgEnd < causeEndLine ? msgEnd : causeEndLine);
+
+                    causes.unshift(cause);
+
+                    causeIdx = err.message.indexOf(CAUSE_STR, causeIdx + CAUSE_STR.length);
+                }
+
+                return new ErrorParseResult(
+                    prefix + (lastIdx >= 0 ? msg.substring(lastIdx + 5, msgEndIdx > 0 ? msgEndIdx : traceIndex) : msg)
```
akuznetsov-gridgain: Please refactor to store causes in ErrorParseResult.

PR #203, opened 2019-06-25
GG-19482 Show valid message and close agent when connected to WC with Java backend.

vsisko:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
```diff
@@ -164,6 +164,24 @@
...
+    /**
+     * On error listener.
+     */
+    private static final Emitter.Listener onConnectionError = args -> {
+        Throwable e = (Throwable)args[0];
+
+        IOException ioCause = X.cause(e, IOException.class);
+
+        if (ioCause != null && "404".equals(ioCause.getMessage())) {
+            log.error("You are using outdated version of Web agent. Please download last version of Web agent from Web console");
```
akuznetsov-gridgain:
```suggestion
            log.error("You are using outdated version of Web agent. Please download latest version of Web agent from Web console");
```

PR #214, opened 2019-06-27
GG-20371 Trimmed WC agent arguments.

vsisko:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentUtils.java
```diff
@@ -300,4 +300,17 @@ public static String nid8(Collection<UUID> nids) {
...
     }
+
+    /**
+     * Trim all elements in list and return new list.
+     *
+     * @param lst List of elements to trim.
+     * @return List with trimmed values.
+     */
+    public static List<String> trim(List<String> lst) {
+        if (!F.isEmpty(lst))
```
akuznetsov-gridgain: lets rewrite like: return F.isEmpty(lst) ? lst : lst.stream().map(String::trim).collect(toList());

PR #225, opened 2019-07-01
GG-20445 Web socket connection is broken after navigating to Receiver page

nva:


modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -410,7 +417,7 @@ export default class AgentManager {
         this.wsSubject
             .pipe(
                 filter((evt) => evt.requestId === requestId || evt.eventType === 'disconnected'),
-                take(1)
+                first()
```
akuznetsov-gridgain: There is a difference between take(1) and first(): https://thecompetentdev.com/weeklyjstips/tips/4_rxjs_first_takeone/
It is safe to use first() here?

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -425,7 +432,19 @@ export default class AgentManager {
                     latch.resolve(evt.payload);
             });
 
-        this._sendWebSocketEvent(requestId, eventType, data);
+        this.wsConnection.pipe(
+            filter((isConnected) => isConnected),
```
Klaster1: On my previous project, I moved the wsConnection-filter-first into separate observable called `connectionReady` and used it to wait for it before making all requests. You might want to at least extract it too, this would indicate the significance of it and reuse in other methods, like `_sendWebSocketEvent` if we decide to.

Also, this observable chain introduces new behavior: agent requests now have an explicit 5s timeout while there was none before. This timeout is different from the option in Sockette options, which tells for how long to wait to reconnect after ECONNREFUSED error. I'd at least test all other screens for regressions.

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
nva: I wanted to reduce imports and using of `first` looks more native than `take(1)`

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -242,10 +244,11 @@ export default class AgentManager {
 
         // Open websocket connection to backend.
         this.ws = new Sockette(uri, {
-            timeout: 5000, // Retry every 5 seconds
```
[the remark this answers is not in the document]
nva: I revert timeout to default. I can't find any reason why we should change default behavior.

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
Klaster1: Well, that makes sense, so why not.

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -432,19 +424,11 @@ export default class AgentManager {
...
-                latch.reject({message: 'Connection to web server was lost'});
-
-                return EMPTY;
-            })
-        )
-        .subscribe(() => {});
+        try {
+            this._sendWebSocketEvent(requestId, eventType, data);
+        } catch (ignored) {
+            latch.reject({message: 'Connection to web server was not established yet'});
```
Klaster1: Did you consider `_sendWebSocketEvent` might throw for other reasons too? The error has code you can check against.

nva: Doesn't matter which reason of exception we should return promise. Improved error message.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentUtils.java
```diff
@@ -298,7 +301,16 @@ public static String nid8(Collection<UUID> nids) {
      * @throws Exception If failed to send event.
      */
     public static void send(Session ses, WebSocketEvent evt) throws Exception {
-        ses.getRemote().sendStringByFuture(toJson(evt)).get();
+        Future<Void> fut = ses.getRemote().sendStringByFuture(toJson(evt));
+
+        try {
+            fut.get(10L, TimeUnit.SECONDS);
```
akuznetsov-gridgain: Are you sure that 10 seconds in enough? May be 30 seconds or 1 minutes will be better?

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/handlers/WebSocketRouter.java
```diff
@@ -249,6 +254,13 @@ private void connect() {
         }
     }
 
+    /**
+     * Connect to backend.
+     */
+    private void connect() {
+        connectorPool.submit(this::connect0);
```
akuznetsov-gridgain: What problem we are trying to solve with connectorPool?

nva: Before fix connect will be executed from thread of websocket client that will be closed in connect.
Connect may be called twice from thread of client.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentUtils.java
nva: We need test this under load and increase if needed. Or separate timeouts for user actions and other events.

PR #227, opened 2019-07-01
GG-19390 Improved public API for Rolling Upgrade

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/commandline/ru/RollingUpgradeCommand.java
```diff
@@ -127,15 +128,31 @@ private VisorRollingUpgradeChangeModeTaskArg toVisorArguments(RollingUpgradeArgu
...
+        RollingUpgradeStatus status = res.status();
 
-        if (status.isForcedModeEnabled())
+        log.info("Rolling upgrade is " + (status.enabled()? "enabled" : "disabled"));
+        log.info("Initial version: " + status.initialVersion());
+        log.info("Update version: " + status.updateVersion());
+
+        log.info("Rolling upgrade is " + (status.enabled()? "enabled" : "disabled"));
+        log.info("Initial version: " + status.initialVersion());
+        log.info("Update version: " + status.updateVersion());
```
glukos: Do we repeat the same messages twice?

modules/core/src/main/java/org/apache/ignite/internal/processors/ru/RollingUpgradeStatus.java
```diff
@@ -30,33 +30,35 @@
...
      * Returns the resulting version.
      * The returned value can be {@code null} if Rolling Upgrade is not in progress
      * or resulting version is not determined yet.
      *
+     * This method makes sense only for the case when the {@code forced} mode is disabled.
+     *
      * @return Resulting version.
      */
-    public @Nullable IgniteProductVersion getUpdateVersion();
+    public @Nullable IgniteProductVersion updateVersion();
```
glukos: Name of this method sounds like it will update something.
Maybe updatingVersion / ongoingVersion / pendingVersion / smth else?

modules/core/src/main/java/org/apache/ignite/internal/commandline/ru/RollingUpgradeCommand.java
sk0x50: Thanks for the catch!

modules/core/src/main/java/org/apache/ignite/internal/processors/ru/RollingUpgradeStatus.java
sk0x50: What do you think about the following pair `initialVersion` & `targetVersion`?

glukos: targetVersion sounds good.

PR #232, opened 2019-07-01
Gg 17447

denis-chudov:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -271,6 +272,9 @@
     /** Last time received message from ring. */
     private volatile long lastRingMsgReceivedTime;
 
+    /** */
+    private volatile Boolean clusterSupportsTcpDiscoveryNodeSerializationOptimization = null;
```
antonovsergey93: Server impl it's single threaded class. `volatile` doesn't required. You could use primitive `boolean` type. Initialize its in constructor and recalculate on nodeAdded/nodeLeft/nodeFailed events.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4411,6 +4497,57 @@ private void nodeCheckError(TcpDiscoveryNode node, String errMsg, String sndMsg)
...
+
+            for (InetSocketAddress addr : addrs) {
+                try {
+                    IgniteSpiOperationTimeoutHelper timeoutHelper = new IgniteSpiOperationTimeoutHelper(spi, true);
+
+                    sendMessageDirectly(msg, addr, timeoutHelper);
+
+                    if (node != null)
+                        node.lastSuccessfulAddress(addr);
+
```
antonovsergey93: You should use `return` here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
denis-chudov: fixed, also added nodeAdd event

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -5054,6 +5170,10 @@ private void processNodeLeftMessage(TcpDiscoveryNodeLeftMessage msg) {
                 return;
             }
 
+            //we will need to recalculate this value since the topology changed
```
antonovsergey93: You must do recalculation only for verified messages!

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -5254,6 +5374,10 @@ private void processNodeFailedMessage(TcpDiscoveryNodeFailedMessage msg) {
...
```
antonovsergey93: You must do recalculation only for verified messages!

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -289,6 +295,12 @@
      */
     ServerImpl(TcpDiscoverySpi adapter) {
         super(adapter);
+
+        clusterSupportsTcpDiscoveryNodeSerializationOptimization = nodeSupports(
+            gridKernalContext(),
+            adapter.locNode,
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -333,13 +345,15 @@
         return upcast(ring.visibleRemoteNodes());
     }
 
+    /** */
+    private GridKernalContext gridKernalContext() {
+        return (spi.ignite() instanceof IgniteEx)? ((IgniteEx)spi.ignite()).context(): null;
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -616,12 +630,80 @@ else if (log.isInfoEnabled()) {
...
+     * @param targetNode Target node where message will be sent.
+     * @return new instance of {@link TcpDiscoveryStatusCheckMessage}.
+     */
+    private TcpDiscoveryStatusCheckMessage createTcpDiscoveryStatusCheckMessage(
+        @Nullable TcpDiscoveryNode creatorNode,
+        @Nullable UUID creatorNodeId,
+        UUID failedNodeId,
+        @Nullable TcpDiscoveryNode targetNode
+    ) {
+        TcpDiscoveryStatusCheckMessage msg;
```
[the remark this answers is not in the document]
denis-chudov: done

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4411,6 +4491,55 @@ private void nodeCheckError(TcpDiscoveryNode node, String errMsg, String sndMsg)
...
+         * @param nodeId Node ID to send message to.
+         * @param addrs Node addresses, used when node could not be found in <code>ring</code>.
+         * @param msg Message.
+         * @throws IgniteSpiException Last failure if all attempts failed.
+         */
+        private void trySendMessageDirectly(UUID nodeId, Collection<InetSocketAddress> addrs, TcpDiscoveryAbstractMessage msg) {
+            TcpDiscoveryNode node = ring.node(nodeId);
+
+            if (node == null) {
+                if (addrs != null)
```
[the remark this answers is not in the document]
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -616,12 +630,80 @@ else if (log.isInfoEnabled()) {
         if (!res && node.clientRouterNodeId() == null && nodeAlive(nodeId)) {
             LT.warn(log, "Failed to ping node (status check will be initiated): " + nodeId);
 
-            msgWorker.addMessage(new TcpDiscoveryStatusCheckMessage(locNode, node.id()));
+            msgWorker.addMessage(createTcpDiscoveryStatusCheckMessage(locNode, locNode.id(), node.id(), locNode));
```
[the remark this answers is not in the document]
denis-chudov: targetNode is used only to serialize node addresses in message, and these addresses are used for direct response which is done by coordinator, so in this case creatorNode and targetNode should be same. Tell me pls if I am wrong.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -616,12 +630,80 @@ else if (log.isInfoEnabled()) {
...
+     * @param creatorNode Creator node.
+     * @param creatorNodeId Creator node id, is used when <code>creatorNode</code> is null.
+     * @param failedNodeId Failed node id.
+     * @param targetNode Target node where message will be sent.
+     * @return new instance of {@link TcpDiscoveryStatusCheckMessage}.
+     */
+    private TcpDiscoveryStatusCheckMessage createTcpDiscoveryStatusCheckMessage(
+        @Nullable TcpDiscoveryNode creatorNode,
+        @Nullable UUID creatorNodeId,
+        UUID failedNodeId,
```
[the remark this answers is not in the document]
denis-chudov: failedNodeId is not id of targetNode so there is no reason to change it.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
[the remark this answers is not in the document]
antonovsergey93: It isn't true.  The `targetNode` required only for sameMac check.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryDuplicateIdMessage.java
```diff
@@ -43,6 +46,24 @@ public TcpDiscoveryDuplicateIdMessage(UUID creatorNodeId, TcpDiscoveryNode node)
         assert node != null;
 
         this.node = node;
+
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryDuplicateIdMessage.java
```diff
@@ -43,6 +46,24 @@ public TcpDiscoveryDuplicateIdMessage(UUID creatorNodeId, TcpDiscoveryNode node)
...
+     *
+     * @param creatorNodeId Creator node ID.
+     * @param nodeId ID of node with same ID.
+     */
+    public TcpDiscoveryDuplicateIdMessage(UUID creatorNodeId, UUID nodeId) {
+        super(creatorNodeId);
+
+        assert nodeId != null;
+
+        this.node = null;
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
```diff
@@ -58,18 +63,26 @@ public TcpDiscoveryStatusCheckMessage(TcpDiscoveryNode creatorNode, UUID failedN
         super(creatorNode.id());
 
         this.creatorNode = creatorNode;
+
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
```diff
@@ -58,18 +63,26 @@ public TcpDiscoveryStatusCheckMessage(TcpDiscoveryNode creatorNode, UUID failedN
...
         this.failedNodeId = failedNodeId;
+
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
```diff
@@ -58,18 +63,26 @@ public TcpDiscoveryStatusCheckMessage(TcpDiscoveryNode creatorNode, UUID failedN
...
+        this.creatorNodeAddrs = null;
     }
 
     /**
-     * @param msg Message to copy.
+     * Constructor.
+     *
+     * @param creatorNodeId Creator node ID.
+     * @param failedNodeId Failed node id.
```
antonovsergey93: Missing javadoc for creatorNodeAddrs

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
```diff
@@ -58,18 +63,26 @@ public TcpDiscoveryStatusCheckMessage(TcpDiscoveryNode creatorNode, UUID failedN
...
      */
-    public TcpDiscoveryStatusCheckMessage(TcpDiscoveryStatusCheckMessage msg) {
-        super(msg);
+    public TcpDiscoveryStatusCheckMessage(UUID creatorNodeId, Collection<InetSocketAddress> creatorNodeAddrs, UUID failedNodeId) {
+        super(creatorNodeId);
+
+        this.creatorNodeAddrs = creatorNodeAddrs;
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
```diff
@@ -58,18 +63,26 @@ public TcpDiscoveryStatusCheckMessage(TcpDiscoveryNode creatorNode, UUID failedN
...
 
-        this.creatorNode = msg.creatorNode;
-        this.failedNodeId = msg.failedNodeId;
-        this.status = msg.status;
+        this.creatorNode = null;
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryDuplicateIdMessage.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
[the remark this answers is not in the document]
denis-chudov: I decided to remove targetNode completely as it is the same as creatorNode.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryDuplicateIdMessage.java
```diff
@@ -41,8 +44,25 @@ public TcpDiscoveryDuplicateIdMessage(UUID creatorNodeId, TcpDiscoveryNode node)
...
+
```
antonovsergey93: Empty line not needed here

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -65,7 +65,10 @@
     TX_TRACKING_UPDATE_COUNTER(12),
 
     /** Support new security processor */
-    IGNITE_SECURITY_PROCESSOR(13);
+    IGNITE_SECURITY_PROCESSOR(13),
+
+    /** Support of dereferencing TcpDiscoveryNode field in discovery messages. */
+    TCP_DISCOVERY_MESSAGE_NODE_SERIALIZATION_OPTIMIZATION(14);
```
sergey-chugunov-1985: To me this name isn't descriptive enough. What do you think about NODE_COMPACT_REPRESENTATION_OPTIMIZATION?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -271,6 +274,10 @@
     /** Last time received message from ring. */
     private volatile long lastRingMsgReceivedTime;
 
+    /** */
+    private volatile boolean clusterSupportsTcpDiscoveryNodeSerializationOptimization =
```
sergey-chugunov-1985: `nodeCompactRepresentationSupport` - to be consistent with feature name.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -616,12 +625,80 @@ else if (log.isInfoEnabled()) {
...
+        TcpDiscoveryStatusCheckMessage msg;
+
+        if (clusterSupportsTcpDiscoveryNodeSerializationOptimization) {
+            TcpDiscoveryNode crd = resolveCoordinator();
+
+            boolean sameMacs = creatorNode != null && crd != null && U.sameMacs(creatorNode, crd);
+
+            msg = new TcpDiscoveryStatusCheckMessage(
+                creatorNode == null ? creatorNodeId : creatorNode.id(),
+                spi.getNodeAddresses(creatorNode, sameMacs),
```
sergey-chugunov-1985: Here we have a chance to get an NPE. Take a look at the logic: according to method signature creatorNode is allowed to be null.
But `spi.getNodeAddresses` doesn't expect null to be passed so this call will result in NPE.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -616,12 +625,80 @@ else if (log.isInfoEnabled()) {
...
+        return msg;
+    }
+
+    /**
+     * Creates new instance of {@link TcpDiscoveryDuplicateIdMessage}.
+     *
+     * @param msg Message.
+     * @return new instance of {@link TcpDiscoveryDuplicateIdMessage}.
+     */
+    private TcpDiscoveryStatusCheckMessage createTcpDiscoveryStatusCheckMessage(
```
sergey-chugunov-1985: Not sure we need this method, it doesn't contain any additional logic around call to the method doing actual work.
Please replace with "real" method at the callsite.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4411,6 +4486,55 @@ private void nodeCheckError(TcpDiscoveryNode node, String errMsg, String sndMsg)
...
+
+        /**
+         * Tries to send a message to all node's available addresses.
+         *
+         * @param nodeId Node ID to send message to.
+         * @param addrs Node addresses, used when node could not be found in <code>ring</code>.
+         * @param msg Message.
+         * @throws IgniteSpiException Last failure if all attempts failed.
+         */
+        private void trySendMessageDirectly(UUID nodeId, Collection<InetSocketAddress> addrs, TcpDiscoveryAbstractMessage msg) {
```
sergey-chugunov-1985: First: need to refactor arguments' list as it doesn't fit into one line.

Second: I don't like that first argument of this method is addrs collection when method `trySendMessageDirectlyToAddrs` has the same argument at the first position. So please rearrange arguments list to make it easier to match them visually.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -4827,6 +4934,10 @@ private void processNodeAddFinishedMessage(TcpDiscoveryNodeAddFinishedMessage ms
             if (log.isDebugEnabled())
                 log.debug("Node to finish add: " + node);
 
+            //we will need to recalculate this value since the topology changed
+            clusterSupportsTcpDiscoveryNodeSerializationOptimization &=
```
sergey-chugunov-1985: No need to recalculate this property if node joins to the cluster that doesn't support the feature.
New node supporting the feature won't change the flag if there are nodes not supporting it.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -5069,6 +5180,10 @@ private void processNodeLeftMessage(TcpDiscoveryNodeLeftMessage msg) {
             }
 
             if (msg.verified() && !locNodeId.equals(leavingNodeId)) {
+                //we will need to recalculate this value since the topology changed
+                clusterSupportsTcpDiscoveryNodeSerializationOptimization =
+                    allNodesSupport(TCP_DISCOVERY_MESSAGE_NODE_SERIALIZATION_OPTIMIZATION);
```
sergey-chugunov-1985: No need to recalculate this property if node leaves the cluster that already supports the feature.
The same applies to NODE_FAILED event.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/messages/TcpDiscoveryStatusCheckMessage.java
```diff
@@ -118,7 +137,7 @@ public void status(int status) {
 
         TcpDiscoveryStatusCheckMessage other = (TcpDiscoveryStatusCheckMessage)obj;
 
-        return F.eqNodes(other.creatorNode, creatorNode) &&
+        return F.eqNodes(other.creatorNodeId(), creatorNodeId()) &&
```
sergey-chugunov-1985: `F.eqNodes` cannot be used to check for equality of nodeIds; only instances of `ClusterNode` class are supported.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -616,12 +625,80 @@ else if (log.isInfoEnabled()) {
...
+                creatorNode == null ? creatorNodeId : creatorNode.id(),
+                spi.getNodeAddresses(creatorNode, sameMacs),
+                failedNodeId
+            );
+        }
+        else {
+            if (creatorNode != null)
+                msg = new TcpDiscoveryStatusCheckMessage(creatorNode, failedNodeId);
+            else
+                throw new IgniteSpiException("Cannot create status check message: creator node is unknown.");
```
sergey-chugunov-1985: Also I have a doubt about this exception. In master we never have a situation when we fail to create a StatusCheckMessage.
This exception may mean that we changed semantics of our code. However I believe we cannot construct a scenario that triggers this branch, so please analyse thoroughly all situations when StatusCheck message could be created. My guess is that we can remove `@Nullable` annotation from `creatorNode` parameter and therefore remove branch throwing exception as well.

denis-chudov: We can receive null in creatorNode parameter when we construct TcpDiscoveryStatusCheckMessage from another message, as the point of optimization was not to serialize this creatorNode.
I replaced exception with returning null, but now I should check method result - if it is null, it means that creator node left topology, and we should do nothing, discarding the message.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
denis-chudov: reworked

PR #234, opened 2019-07-02
GG-19573 Initial implementation of Agent inside node.

akuznetsov-gridgain:


modules/core/src/main/java/org/apache/ignite/internal/processors/gmc/ManagementConsoleProcessor.java
```diff
@@ -0,0 +1,40 @@
...
+     */
+    public ManagementConsoleProcessor(GridKernalContext ctx) {
+        super(ctx);
+    }
+
+    /**
+     * @return No-op processor usage exception;
+     */
+    private IgniteException processorException() {
+        return new IgniteException("Current Ignite configuration does not support Management Console.");
```
nva: Why Ignite configuration? 
Why Management Console?

modules/web-console/web-console-common/src/main/java/org/apache/ignite/console/websocket/AgentHandshakeRequest.java
```diff
@@ -41,6 +42,12 @@
     /** */
     private String ver;
 
+    /** Cluster id. */
+    private UUID clusterId;
+
+    /** Last seen. */
+    private long lastSeen;
```
nva: Why we need this? As I remember we need latest metrics timestamp

PR #238, opened 2019-07-03
GG-20751 Code cleanup (unused import, missing comments, ...).

akuznetsov-gridgain:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/common/Utils.java
```diff
@@ -125,4 +126,12 @@ public static String currentRequestOrigin() {
             .build()
             .toString();
     }
+
+    /**
+     * @param p Predicate.
+     * @return Negated predicate.
+     */
+    public static <T> Predicate<T> not(Predicate<T> p) {
```
nva: Why we need this? I can't find any usage

akuznetsov-gridgain: Good catch, I forget to delete it

PR #241, opened 2019-07-03
GG-19572: Add cluster id prototype.

smoldachev:


modules/web-console/web-console-server/src/test/java/org/apache/ignite/console/web/controller/ClusterInfoControllerTest.java
```diff
@@ -0,0 +1,59 @@
+package org.apache.ignite.console.web.controller;
```
akuznetsov-gridgain: Missing license.

modules/web-console/web-console-server/pom.xml
```diff
@@ -162,6 +162,12 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
+            <groupId>org.springframework.security</groupId>
+            <artifactId>spring-security-test</artifactId>
```
akuznetsov-gridgain: Why we need this dependency?

smoldachev: For this amazing annotation @WithMockUser :)

modules/web-console/web-console-server/src/test/java/org/apache/ignite/console/web/controller/ClusterInfoControllerTest.java
smoldachev: Fix

PR #242, opened 2019-07-04
GG-20773 Web Console: possible memory leak in backend after web agent download

nva:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/controller/AgentDownloadController.java
```diff
@@ -78,56 +71,48 @@
...
+                zos.write(content.getBytes(UTF_8));
+                zos.closeArchiveEntry();
+
+                try (ZipFile zip = new ZipFile(latestAgentPath.toFile())) {
+                    // Make a copy of agent ZIP.
+                    zip.copyRawEntries(zos, rawEntry -> true);
+                }
+            }
+            catch (IOException ignored) {
+                // No-op.
```
akuznetsov-gridgain: May be we should log error? I do not like "muting" errors. It will be hard to detect problems without errors in log.

nva: I mute only IOException (closed pipe, ...) I don't see any reason to log closed pipe or send error

PR #247, opened 2019-07-04
GG-19486 IgniteOutOfMemoryException should not prevent cache clearing

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -982,6 +985,63 @@ public void releaseHistoryForPreloading() {
...
+     * For example, removing a row from underlying store may require allocating a new data page
+     * in order to move a tracked page from one bucket to another one which does not have a free space for a new stripe.
+     * See {@link AbstractFreeList#removeDataRowByLink}.
+     * Therefore, inserting a new entry should be prevented in case of some threshold is exceeded.
+     *
+     * @param region Data region to be checked.
+     * @throws IgniteOutOfMemoryException In case of the given data region does not have enough free space
+     * for putting a new entry.
+     */
+    public void ensureFreeSpaceForInsert(DataRegion region)throws IgniteOutOfMemoryException {
```
glukos: space between ) and throws

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -982,6 +985,63 @@ public void releaseHistoryForPreloading() {
...
+
+        long memorySize = regCfg.getMaxSize();
+
+        PageMemory pageMem = region.pageMemory();
+
+        int sysPageSize = pageMem.systemPageSize();
+
+        CacheFreeList freeList = freeListMap.get(regCfg.getName());
+
+        long dirtyPages = (pageMem.loadedPages() - freeList.emptyDataPages());
```
glukos: I guess, "dirty" isn't the best wording for this number as it clashes with "dirty page" abstraction from persistence

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -982,6 +985,63 @@ public void releaseHistoryForPreloading() {
...
+
+        boolean oomThreshold = (memorySize / sysPageSize) <
+            (dirtyPages * (8.0 /*link size*/ / sysPageSize + 1) + 256 /*one page per bucket*/);
```
glukos: A bit risky.
Less than sysPageSize space is used in PagesListMetaIO / PagesListNodeIO (minus headers and so on) for storing links. I'd multiply it by 1.5 to be completely sure.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
sk0x50: Agreed. I will change it to `nonEmptyPages`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
sk0x50: You are right. The space that can be used for storing links should be calculated as follows, I think:
`pageMem.pageSize() - PagesListNodeIO.PAGE_IDS_OFF`.
Perhaps, a constant multiplier would be a good approach in order to eliminate different versions of `PagesListNodeIO/PagesListMetaIO` (that can be added in the future).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
sk0x50: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
sk0x50: I updated the calculation as follows:
```java
        // The maximum number of pages that can be allocated (memorySize / systemPageSize)
        // should be greater or equal to the current number of non-empty pages plus
        // the number of pages that may be required in order to move all pages to a reuse bucket,
        // that is equal to nonEmptyPages * 8 / pageSize, where 8 is the size of a link.
        // Note that not the whole page can be used to storing links,
        // see PagesListNodeIO and PagesListMetaIO#getCapacity(), so we pessimistically multiply the result on 1.5,
        // in any way, the number of required pages is less than 1 percent.
        boolean oomThreshold = (memorySize / pageMem.systemPageSize()) <
            (nonEmptyPages * (8.0 / pageMem.pageSize() + 1) * 1.5 + 256 /*one page per bucket*/);
```

PR #259, opened 2019-07-09
GG-18554 Store last configured refresh rate.

vsisko:


vsisko:
> I don't think we need this at all.

Removed storage service. Stay removing of refresh rate by logout as discussed with @nva.

vsisko:
> I don't think this PR is required at all. You can listen for user changes and reset value in storage as a separate listener, not a part of User service. The User service should arguably not be concerned about refreshRate at all.

Moved as separated effect into GG repository.

PR #260, opened 2019-07-09
NBL-8: CPP: Implement missing methods for ClusterNode class

FedorUporov:


modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
```diff
@@ -44,27 +45,189 @@ namespace ignite
...
+                 * Get a node attribute.
+                 *
+                 * @param name Node attribute name.
+                 * @return Node attribute.
+                 *
+                 * @throw IgniteError in case of attribute name does not exist
+                 * or if template type is not compatible with attribute.
+                 */
+                template<typename T>
+                T GetAttribute(std::string name)
```
isapego: Let's add `bool IsAttributeSet() const` method for user so they will be able check if the attribute present in a simple way.

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
```diff
@@ -44,27 +45,189 @@ namespace ignite
...
+                {
+                    if (attrs.Get()->find(name) == attrs.Get()->end())
+                    {
+                        const char* msg = "The is no Cluster Node attribute with name requested";
```
isapego: Typo: `The is...`

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
```diff
@@ -44,27 +45,189 @@ namespace ignite
...
+                 */
+                std::vector<std::string> GetAttributes();
+
+                /**
+                 * Get Cluster Node consistent ID.
+                 *
+                 * @return Cluster Node consistent ID.
+                 */
+                template<typename T>
+                T GetConsistentId()
```
isapego: We should return `std::string` here, or it will not usable for user.

modules/platforms/cpp/core/include/ignite/ignite_product_version.h
```diff
@@ -0,0 +1,72 @@
...
+        /** SHA1 Length. */
+        static const int SHA1_LENGTH = 20;
+
+        /**
+         * Default constructor.
+         */
+        IgniteProductVersion(int8_t majorNumber, int8_t minorNumber, int8_t maintenance, std::string stage, int64_t releaseDate, std::vector<int8_t> revHash) :
+            majorNumber(majorNumber), minorNumber(minorNumber), maintenance(maintenance), stage(stage), releaseDate(releaseDate), revHash(revHash)
+        {
+            if (revHash.size() != SHA1_LENGTH)
```
isapego: Let's check and throw this exception **before** it's constructed. In constructor it is enough to use `assert` just to make sure it never happens.

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
```diff
@@ -44,27 +45,189 @@ namespace ignite
...
+
+                    return reader.ReadObject<T>();
+                }
+
+                /**
+                 * Get collection of host names this node is known by.
+                 *
+                 * @return Collection of host names this node is known by.
+                 */
+                std::vector<std::string> GetHostNames();
```
isapego: The same as for `GetAddresses()`

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
```diff
@@ -44,27 +45,189 @@ namespace ignite
...
+                 * @return Node order.
+                 */
+                long GetOrder();
+
+                /**
+                 * Get node version.
+                 *
+                 * @return Prodcut version.
+                 */
+                IgniteProductVersion GetVersion();
```
isapego: Why do we return value and not const reference here?

modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
```diff
@@ -373,7 +373,7 @@
     public int getTotalCpus();
 
     /**
-     * Returns the CPU usage usage in {@code [0, 1]} range.
+     * Returns the CPU usage in {@code [0, 1]} range.
```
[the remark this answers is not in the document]
FedorUporov: The "usage" is written two time, not sure that it is correct.

modules/platforms/cpp/binary/include/ignite/impl/interop/interop_input_stream.h
```diff
@@ -202,6 +202,13 @@ namespace ignite
                  */
                 void Position(int32_t pos);
 
+                /**
+                 * Shift stream to the right.
+                 *
+                 * @param cnt Amount of bytes to shift the stream to.
+                 */
+                void Shift(int32_t cnt);
```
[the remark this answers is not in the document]
FedorUporov: Not sure, because we have void CopyAndShift(int8_t* dest, int32_t off, int32_t cnt); below.

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
```diff
@@ -795,6 +823,275 @@ namespace ignite
...
+
+                    return;
+                }
+                else if (typeId == IGNITE_HDR_FULL)
+                {
+                    // Get position back to IGNITE_HDR_FULL
+                    stream->Position(stream->Position() - 1);
+
+                    EmptyType val;
+                    ReadTopObject0<EmptyReader, EmptyType>(val);
```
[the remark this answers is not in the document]
FedorUporov: To skip IGNITE_HDR_FULL correctly we need to pass thru logic from binary_reader_impl.h:
template<typename R, typename T>
void ReadTopObject0(T& res)
switch case IGNITE_HDR_FULL:
It is difficult to calculate offset from it.

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
FedorUporov: It is not possible because sometimes it could be Guid.
It is possible to convert Guid to string, but I am not sure that it is suitable for user.

modules/platforms/cpp/binary/include/ignite/impl/interop/interop_input_stream.h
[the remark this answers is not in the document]
isapego: `CopyAndShift` is internal method, it related to details of internal implementation of stream. Public methods should be more semantic. Semantically `Shift()` is much more confusing to me as a user of this class than `Ignore()` (Shift what? Where? I'd go to the code to check what does this method do if I was a new to this code). What do you think?

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
isapego: Yes it is. AFAIK, this is the only way it is used in Java right now - everyone just call `.toString()` on returned method - because how else you can really use it?

The most widespread usage of this method is comparison and strings are perfectly comparable just as well as GUIDs.

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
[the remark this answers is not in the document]
isapego: The offset is `IGNITE_OFFSET_LEN`, so basically you write here:
```
int32_t len = stream->ReadInt32(stream->Position() + IGNITE_OFFSET_LEN);
stream->Shift(len);
```

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
```diff
@@ -795,6 +823,166 @@ namespace ignite
...
+                    case IGNITE_TYPE_DECIMAL:
+                    case IGNITE_TYPE_ARRAY_DECIMAL:
+                    case IGNITE_TYPE_CLASS:
+                    case IGNITE_TYPE_PROXY:
+                    case IGNITE_TYPE_BINARY_ENUM:
+                        IGNITE_ERROR_FORMATTED_2(IgniteError::IGNITE_ERR_BINARY, "Invalid header", "position", pos,
+                            "unsupported type", static_cast<int>(typeId))
+                    default:
+                        IGNITE_ERROR_FORMATTED_2(IgniteError::IGNITE_ERR_BINARY, "Invalid header", "position", pos,
+                            "invalid type", static_cast<int>(typeId))
```
isapego: I believe, it should be _"unsupported type"_ for all cases, as we can introduce new types in future, which are not _invalid_.

modules/platforms/cpp/binary/include/ignite/impl/interop/interop_input_stream.h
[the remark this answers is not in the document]
FedorUporov: Public method Ignore() added.

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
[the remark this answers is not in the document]
FedorUporov: Fixed.

modules/platforms/cpp/core/include/ignite/impl/cluster/cluster_node_impl.h
FedorUporov: Now the std::string returned in all cases.

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
FedorUporov: Fixed.

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
```diff
@@ -945,40 +918,26 @@ namespace ignite
...
                 case IGNITE_HDR_FULL:
-                    // Get position back to IGNITE_HDR_FULL
-                    stream->Position(stream->Position() - 1);
-                    EmptyType val;
-                    ReadTopObject0<EmptyReader, EmptyType>(val);
+                {
+                    int32_t pos = stream->Position() - 1;
+                    int32_t len = stream->ReadInt32(stream->Position() + IGNITE_OFFSET_LEN - 1);
+                    stream->Position(pos);
+                    stream->Ignore(len);
```
isapego: Looks a little bit too complicated. How about the following?
```
int32_t objectBegin = stream->Position() - 1;
int32_t len = stream->ReadInt32(objectBegin + IGNITE_OFFSET_LEN);
stream->Position(objectBegin + len);
```

modules/platforms/cpp/core/src/impl/cluster/cluster_node_impl.cpp
```diff
@@ -139,8 +144,17 @@ namespace ignite
 
             void ClusterNodeImpl::ReadConsistentId(BinaryReaderImpl& reader)
             {
-                consistentIdOffset = reader.GetStream()->Position();
-                reader.Skip();
+                int8_t typeId = reader.ReadInt8();
```
isapego: Not sure about this part, need to consult with other team members.

isapego: Ok, let's revert this part and leave it the way it was before with `T` as a return type.
We need to mention in doxycomment that be default `T` should be `std::string` though.

FedorUporov: It was decided  to leave not templated function which will return std:string after discussion.

PR #262, opened 2019-07-10
GG-21182: Use reference queue to remove unused Locks in LocalModelStorageProvider

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/inference/storage/model/LocalModelStorageProvider.java
```diff
@@ -48,8 +54,50 @@
...
+            locks.put(key, new WeakReferenceWithCleanUp(key, lock));
+            return lock;
+        }
+        finally {
+            // At this point we already replaced all keys we wanted to replace, so all empty references could be safely
+            // deleted.
+            while ((ref = (WeakReferenceWithCleanUp)refQueue.poll()) != null) {
+                String keyToBeRmv = ref.key;
+                // We double check that we don't replaced the key value already.
+                if (locks.containsKey(keyToBeRmv) && locks.get(keyToBeRmv).get() == null)
```
agoncharuk: can be replaced with locks.remove(ref.key, ref) ?

dmitrievanthony: Good point. Done.

PR #273, opened 2019-07-15
GG-21332 Do not rely on the value returned from user profile update endpoint

Klaster1:
Previously, the '/api/v1/profile/save' endpoint used to return an updated user object. The new Java backend sends an empty response, so the user value assumes the empty string value, which in turn causes conditions that control UI element visibility to return false. I chose to fix the issue on the frontend side.

### DIFF
--- a/modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/controller/AccountController.java
+++ b/modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/controller/AccountController.java
@@ -92,12 +92,14 @@ public class AccountController {
     }
 
     /**
+     * @param req Request wrapper.
      * @param acc Current user.
      * @param changes Changes to apply to user.
      */
     @ApiOperation(value = "Save user.")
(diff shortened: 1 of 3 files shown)


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/web/controller/AccountController.java
```diff
@@ -107,7 +109,7 @@ public AccountController(AuthenticationManager authMgr, AccountsService accounts
 
         SecurityContextHolder.getContext().setAuthentication(authentication);
 
-        return ResponseEntity.ok().build();
+        return ResponseEntity.ok(new UserResponse(acc, req.isUserInRole(ROLE_PREVIOUS_ADMINISTRATOR)));
```
akuznetsov-gridgain: Lets move "req.isUserInRole(ROLE_PREVIOUS_ADMINISTRATOR)" to separate utility method like:
public static boolean isBecomeUsed(req) { return req.isUserInRole(ROLE_PREVIOUS_ADMINISTRATOR)}

and use it in AccountController and GridGainAccountController

Klaster1:
@nva the frontend changes are OK.

PR #280, opened 2019-07-15
GG-19383 Fixed client PME hang on optimistic transactions

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheGateway.java
```diff
@@ -352,7 +355,12 @@ public void onStopped() {
      * @throws IgniteException - in case of atomic operation inside transaction without permission.
      */
     private void checkAtomicOpsInTx(CacheOperationContext opCtx) throws IgniteException {
-        if (ctx.atomic() && !opCtx.allowedAtomicOpsInTx()) {
+        CacheConfiguration cfg = ctx.config();
+
+        if (cfg == null)
```
kreig133: Why do we rely on cfg == null? Looks very fragile.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxOnCachesStopTest.java
```diff
@@ -254,10 +345,182 @@ private void runTxOnCacheStop(TransactionConcurrency conc, TransactionIsolation
...
+
+    /**
+     * Creates a list of cache configurations.
+     *
+     * @return List of cache configurations.
+     */
+    private List<CacheConfiguration> createCacheConfigurations() {
+        String GRP_NAME = "test-destroy-group";
+
+        List<CacheConfiguration> cacheCfgs = new ArrayList<>(50);
```
kreig133: What is 50? You have already CACHE_CNT, could we replace?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxOnCachesStopTest.java
```diff
@@ -254,10 +345,182 @@ private void runTxOnCacheStop(TransactionConcurrency conc, TransactionIsolation
...
+     * @param cacheIdxToBeDestroyed Variable that allows to get an index of destroyed cache.
+     * @param clients Client nodes that are used for initiating transactions.
+     * @param concurrency Transaction concurrency mode.
+     * @param isolation Transaction isolation.
+     * @return TxLoad future.
+     */
+    private IgniteInternalFuture startTxLoad (
+        final AtomicBoolean stopTxLoad,
+        final AtomicInteger cacheIdxToBeDestroyed,
+        final ArrayList<Ignite> clients,
```
kreig133: ArrayList? Maybe, just list? What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheGateway.java
sk0x50: Unfortunately, this method is called from `GridCacheGateway#enter(CacheOperationContext)` before the `readLock` is acquired, and therefore there is no guarantee that the cache context was not cleaned. Perhaps, it makes sense to move this check under the mentioned lock.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxOnCachesStopTest.java
sk0x50: Nice catch!

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxOnCachesStopTest.java
sk0x50: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -3145,6 +3145,63 @@ void forceCloseCaches(ExchangeActions exchActions) {
...
     }
 
+    /**
+     * Blocks cache gateways before rolling back transactions that were affected by Exchange,
+     * which in turn was caused by the destroyCache call for these caches.
+     * This should be done in order to prohibit enlisting new entries into optimistic transactions.
+     *
+     * @param exchActions Change requests.
+     */
+    private void blockCacheGatewaysOnExchangeDone(ExchangeActions exchActions) {
```
glukos: This is old version of logic, also this method is not used. We should remove it.

sk0x50: Shame on me! Will remove.

PR #283, opened 2019-07-16
GG-21333

ingvard:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheOffheapManager.java
```diff
@@ -2230,11 +2230,9 @@ private Metas getOrAllocatePartitionMetas() throws IgniteCheckedException {
         /** {@inheritDoc} */
         @Override public void updateInitialCounter(long start, long delta) {
             try {
-                CacheDataStore delegate0 = init0(true);
-
-                if (delegate0 == null)
-                    throw new IllegalStateException("Should be never called.");
+                CacheDataStore delegate0 = init0(false);
```
ingvard: Could you define the behavior of init0 with checkExists? Why this change is correct?

ascherbakoff: This is a fix for flaky test which I figured out while working on green visa. See my comment at line 2235. It has no direct relation to the fix, but is caused by GG-17348.

glukos: @ascherbakoff ok, please describe semantic of GridCacheDataStore#init0 (explain its contract and what does it initialize) and add proper doc for checkExists argument (what exists? partition file, instance of class or what? it's not obvious from the code).

PR #299, opened 2019-07-19
GG-19184: Offload to disk for distinct and plain values

rkondakov:
This patch introduces disk spilling (or disk offloading) for intermediate query results. These intermediate results are stored in `H2ManagedLocalResult`, which is an almost certain copy of H2's `LocalResultImpl` but with a more fine-grained memory control. When stored rows exceed established memory limit, all these rows are spilled to the disk using one of the implementations of `ResultExternal` interface: 

- `PlainExternalResult` - is used for storing unsorted results. It has a quite straightforward implementation: each new row is serialized and spilled to the disk. Once rows are stored on the disk, they can be read from it consequently.

- `SortedExternalResult` - is used for storing sorted or distinct results. Upon adding a new row it is stored in in-memory buffer first. When the buffer exceeds the memory limit, it is first sorted and then flushed to the disk. This flushed buffer is called `Chunk`. After adding all needed rows we'll have a list of this sorted chunks where we can read sorted data from using merge sort algorithm. In distinct cases we have to use internal index `ExternalResultHashIndex` to determine whether the external result contains added row or not. This index is built as a disk-based open addressing hash table: when the table bucket is not empty  - we put entry in the next empty bucket. Hash map is resized on demand.


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1979,6 +1986,26 @@ public QueryMemoryManager memoryManager() {
...
+        try {
+            File spillDir = U.resolveWorkDirectory(
+                ctx.config().getWorkDirectory(),
+                DISK_SPILL_DIR,
+                false);
+
+            U.delete(spillDir);
+        }
+        catch (Exception e) {
+            log.debug("Failed to delete spill directory.", X.getFullStackTrace(e));
```
AMashenkov: Logging this at warning level may be useful for user in case an app has no rights to write to spill directory.
Also let's change msg to smth "Failed to cleanup SQL temporary directory with intermediate SQL query results from previous node run."

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryTracker.java
```diff
@@ -104,47 +132,32 @@
...
-            long reserved = release0(size);
+            long res = reserved - size;
+
+            if (res < 0)
+                throw new IllegalStateException("Try to free more memory that ever be reserved: [" +
+                    "reserved=" + reserved + ", toFree=" + size + ']');
+
+            long reserved = this.reserved = res;
 
             assert reserved >= 0;
```
AMashenkov: This assert looks useless

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,374 @@
...
+        try {
+            String fileName = "spill_" + ctx.localNodeId() + "_" + idGen.incrementAndGet();
+
+            file = new File(U.resolveWorkDirectory(
+                ctx.config().getWorkDirectory(),
+                DISK_SPILL_DIR,
+                false
+            ), fileName);
+
+            file.deleteOnExit();
```
AMashenkov: This causes put into shared LinkedHashSet inside synchronized block.
Let's remove it as 
1. directory will be cleaned on next node start
2. it is designed that fileName will never be reused (filename contains node uuid).

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,374 @@
...
+
+            throw new IgniteException("Failed to read query result the from spill file.", e);
+        }
+    }
+
+    /**
+     * @return New data buffer.
+     */
+    @NotNull protected Data createDataBuffer() {
+        return Data.create(null, INIT_BUF_SIZE, true); // TODO init size
```
AMashenkov: Let's remove "TODO" or specify a ticket number.
BTW, H2 uses Constants.DEFAULT_PAGE_SIZE = 4096 as default buffer size for rows.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/PlainExternalResult.java
```diff
@@ -0,0 +1,100 @@
...
+        Value[] row = readRowFromFile();
+
+        return row;
+    }
+
+    /** {@inheritDoc} */
+    @Override public int addRow(Value[] row) {
+        Data buff = createDataBuffer();
+
+        addRowToBuffer(row, buff);
```
AMashenkov: Looks like required buffer size can be easily calculated for this particular case.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+        return size++;
+    }
+
+    /**
+     * Checks if current result contains given row with sort order check.
+     *
+     * @param row Row.
+     * @return {@code True} if current result does not contain th given row.
+     */
+    private boolean containsRowWithOrderCheck(Value[] row) { // TODO merge removeRow and getPreviousRow
```
AMashenkov: Let's remove TODO.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+        return true;
+    }
+
+    /**
+     * Returns the previous row.
+     *
+     * @param row Row.
+     * @return Previous row.
+     */
+    @Nullable private Value[] getPreviousRow(Value[] row) { // TODO merge removeRow and getPreviousRow
```
AMashenkov: Let's remove TODO

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+
+            if (unsortedRowsBuf == null)
+                unsortedRowsBuf = new ArrayList<>();
+
+            unsortedRowsBuf.add(row);
+        }
+
+        long delta = H2Utils.calculateMemoryDelta(null, null, row);
+
+        memTracker.reserved(delta);
```
AMashenkov: Let's make reservation before creating\adding row into buffer.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+
+    /** {@inheritDoc} */
+    @Override protected void onClose() {
+        super.onClose();
+
+        U.closeQuiet(hashIdx);
+    }
+
+    /** {@inheritDoc} */
+    @Override public int removeRow(Value[] values) { // TODO merge removeRow and getPreviousRow
```
AMashenkov: TODO

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+            this.curPos = start;
+            this.end = end;
+        }
+
+        /**
+         * @return {@code True} if next row is available within a chunk.
+         */
+        boolean next() {
+            while (curPos < end) {
+                curRow = readRowFromFile(curPos); // TODO read multiple rows and cache it if possible.
```
AMashenkov: TODO with no ticket number

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -933,6 +933,70 @@ public void testCeilPow2() throws Exception {
...
+        assertEquals(1 << 27, U.ceilPow2Long((1 << 26) + 100));
+
+        assertEquals(1L << 61, U.ceilPow2Long((1L << 61) - 100));
+        assertEquals(1L << 61, U.ceilPow2Long(1L << 61));
+        assertEquals(1L << 62, U.ceilPow2Long((1L << 61) + 100));
+
+        assertEquals(1L << 62, U.ceilPow2Long((1L << 62) - 1));
+        assertEquals(1L << 62, U.ceilPow2Long(1L << 62));
+
+        try {
```
AMashenkov: Is it possible to replace this with assertThrows() ?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -98,6 +109,9 @@ public QueryMemoryManager(GridKernalContext ctx, long globalQuota) {
 
         this.blockSize = Long.getLong(IgniteSystemProperties.IGNITE_SQL_MEMORY_RESERVATION_BLOCK_SIZE, DFLT_MEMORY_RESERVATION_BLOCK_SIZE);
         this.globalQuota = globalQuota;
+        // TODO GG-18629 - get from configuration.
+        this.failOnMemLimitExceed = Boolean.getBoolean(IgniteSystemProperties.IGNITE_SQL_FAIL_ON_QUERY_MEMORY_LIMIT_EXCEED);
```
AMashenkov: Would IGNITE_SQL_USE_DISK_OFFLOAD be more clear to user?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
rkondakov: Done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryTracker.java
rkondakov: Done.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
rkondakov: Done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
rkondakov: Done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/PlainExternalResult.java
rkondakov: Done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
rkondakov: Done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
rkondakov: Done

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/LocalQueryMemoryTrackerSelfTest.java
```diff
@@ -16,9 +16,12 @@
 
 package org.apache.ignite.internal.processors.query.oom;
 
+import org.apache.ignite.testframework.junits.WithSystemProperty;
+
 /**
  * Query memory manager for local queries.
  */
+@WithSystemProperty(key = "IGNITE_SQL_FAIL_ON_QUERY_MEMORY_LIMIT_EXCEED", value = "true")
```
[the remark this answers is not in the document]
rkondakov: Memory tracking and disk spilling tests have different testing strategies. IMO it is not a good idea of reusing the same test framework for both cases.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/QueryMemoryTrackerSelfTest.java
```diff
@@ -34,6 +35,7 @@
 /**
  * Query memory manager test for distributed queries.
  */
+@WithSystemProperty(key = "IGNITE_SQL_FAIL_ON_QUERY_MEMORY_LIMIT_EXCEED", value = "true")
```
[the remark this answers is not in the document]
rkondakov: See my answer in the previous comment.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
rkondakov: Done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
rkondakov: Agree. This name looks more meaningful. Done.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,546 @@
...
+        cp.rows = this.rows;
+        cp.sort = this.sort;
+        cp.distinctRows = this.distinctRows;
+        cp.distinct = distinct;
+        cp.distinctIndexes = distinctIndexes;
+        cp.currentRow = null;
+        cp.offset = 0;
+        cp.limit = -1;
+        cp.containsNull = containsNull;
+        cp.external = e2;
```
AMashenkov: Do we need to copy memTracker as well?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,546 @@
...
+        }
+    }
+
+    private void addRowsToDisk() {
+        if (external == null) {
+            createExternalResult();
+        }
+
+        if (memTracker != null)
+            memTracker.released(memReserved);
```
AMashenkov: Memory released before row.clear()

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,546 @@
...
+                ValueRow array = getDistinctRow(values);
+                Value[] previous = distinctRows.get(array);
+                if (previous == null || sort != null && sort.compare(previous, values) > 0) {
+                    distinctRows.put(array, values);
+                }
+                rowCount = distinctRows.size();
+                if (!isEnoughMemoryAfterReservation(array, previous, values)) {
+                    createExternalResult();
+                    rowCount = external.addRows(distinctRows.values());
+                    distinctRows = null;
```
AMashenkov: should we release memory here?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -102,29 +110,43 @@ public QueryMemoryManager(GridKernalContext ctx, long globalQuota) {
...
+    @Override public boolean reserved(long size) {
         if (size == 0)
-            return; // Nothing to do.
+            return true; // Nothing to do.
 
-        assert size > 0;
+        long reserved0 = reserved.addAndGet(size);
 
-        reserved.accumulateAndGet(size, reserveOp);
+        if (reserved0 >= globalQuota) {
```
AMashenkov: Let's revert back accumulateAndGet() call for disabled diskOffload.
E.g. smth like this:

if (failOnMemLimitExceed)
   reserved.accumulateAndGet(size, reserveOp);
else {
    long reserved0 = reserved.addAndGet(size);
    return reserved0 < globalQuota;
}

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -102,29 +110,43 @@ public QueryMemoryManager(GridKernalContext ctx, long globalQuota) {
...
 
     /** {@inheritDoc} */
-    @Override public void reserve(long size) {
+    @Override public boolean reserved(long size) {
         if (size == 0)
-            return; // Nothing to do.
+            return true; // Nothing to do.
 
-        assert size > 0;
+        long reserved0 = reserved.addAndGet(size);
```
AMashenkov: Can we use GridAtomicLong.checkAndSet() here.
A predicate can fails with exception if offload is disabled (failOnMemLimitExceed==false) as it works before,
and return true\false if offload is enabled

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+    private long findFreeSlotForInsert(int hashCode) {
+        long slot = slot(hashCode);
+
+        Entry entry;
+
+        while ((entry = readEntryFromIndexFile(slot)) != null && !entry.isRemoved() && !entry.isEmpty()) {
+            slot = slot((int)(slot + 1)); // Check next slot.
+
+            if (slot == slot(hashCode))
+                throw new RuntimeException("No free space left in the hash map.");
```
AMashenkov: Let's throw some meaningful exception. E.g. IOOM or SQL exception with OOM code.

AMashenkov: Also, seems there is no need to calculate slot(hashcode) more than once.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+     *
+     * @param hashCode Hash code.
+     * @return Free slot number.
+     */
+    private long findFreeSlotForInsert(int hashCode) {
+        long slot = slot(hashCode);
+
+        Entry entry;
+
+        while ((entry = readEntryFromIndexFile(slot)) != null && !entry.isRemoved() && !entry.isEmpty()) {
```
AMashenkov: Actually, there is no need to read entry and create Entry instance, a check will be enough.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+
+                if (row != null) {
+                    ValueRow keyFromDisk = rowStore.getRowKey(row);
+
+                    if (key.equals(keyFromDisk)) // 2. Check the actual row equals the given key.
+                        break;
+                }
+            }
+
+            slot = slot((int)(slot + 1)); // Check the next slot.
```
AMashenkov: Does you mean slot++ ?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+import static java.nio.file.StandardOpenOption.CREATE_NEW;
+import static java.nio.file.StandardOpenOption.READ;
+import static java.nio.file.StandardOpenOption.WRITE;
+import static org.apache.ignite.internal.processors.query.h2.IgniteH2Indexing.DISK_SPILL_DIR;
+
+/**
+ * Basic class for external result. Contains common methods for file IO.
+ */
+@SuppressWarnings({"MissortedModifiers", "WeakerAccess", "ForLoopReplaceableByForEach"})
+public abstract class AbstractExternalResult implements ResultExternal {
```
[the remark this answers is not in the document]
rkondakov: Could you explain the advantages of this splitting? It is not clear for me. `AbstractExternalResult` encapsulates all file-related operations and it's child classes are responsible for specific serialization and dumping to disk.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+    /** File name generator. */
+    private static final AtomicLong idGen = new AtomicLong();
```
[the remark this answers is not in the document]
rkondakov: This class is single threaded. `volatile` declaration should be replaced with a `final` one - it looks like a rudiment, so I'll fix it. `AtomicLong` here is used for unique sequence generation. Multiple external results may be created simultaneously and we need to take care they have unique file names.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+
+            fileIo = fileIOFactory.create(file, CREATE_NEW, READ, WRITE);
+
+            if (log.isDebugEnabled())
+                log.debug("Created spill file "+ file.getName());
+
+            this.memTracker = memTracker;
+
+            if (memTracker != null)
+                memTracker.registerCloseListener(this::close);
```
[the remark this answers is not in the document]
rkondakov: Agree.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+     *
+     * @param rowLen Data length to read.
+     * @return Byte buffer with data.
+     */
+    @NotNull protected ByteBuffer readDataFromFile(int rowLen) {
+        try {
+            ByteBuffer rowBytes = ByteBuffer.allocate(rowLen); // 2 x integer for length in bytes and columns count.
+
+            while (rowBytes.hasRemaining()) {
+                int bytesRead = fileIo.read(rowBytes);
```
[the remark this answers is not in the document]
rkondakov: Good point. Thank you. I'll fix it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+
+    /**
+     * Reads data from file.
+     *
+     * @param rowLen Data length to read.
+     * @return Byte buffer with data.
+     */
+    @NotNull protected ByteBuffer readDataFromFile(int rowLen) {
+        try {
+            ByteBuffer rowBytes = ByteBuffer.allocate(rowLen); // 2 x integer for length in bytes and columns count.
```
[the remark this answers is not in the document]
rkondakov: Agree. Going to fix it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+        return Data.create(null, cap, true);
+    }
+
+    /**
+     * Adds row to buffer.
+     *
+     * @param row Row.
+     * @param buff Buffer.
+     */
+    protected void addRowToBuffer(Value[] row, Data buff) {
```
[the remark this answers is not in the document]
rkondakov: There is no recovery needed. External result is needed as long as the parent query is alive. When node is failed, all queries are stopped, so all external results can easily be discarded.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+
+    /**
+     * Marks row in the file as removed by setting {@code -1} on it's {@code colCount} position.
+     *
+     * @param addr Row absolute address in the file.
+     */
+    protected void markRowRemoved(long addr) {
+        setFilePosition(addr);
+
+        ByteBuffer hdr = readRowHeaderFromFile();
```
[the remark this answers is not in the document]
rkondakov: Could you please point out where I  read row from disk? I read only header.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+    protected final File file;
+
+    /** Spill file IO.*/
+    protected volatile FileIO fileIo;
+
+    /** Memory tracker. */
+    protected final H2MemoryTracker memTracker;
+
+    /** Parent result. */
+    protected final AbstractExternalResult parent;
```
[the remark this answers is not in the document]
rkondakov: This construction is needed for query results caching. Sometimes H2 caches the query results instead of running a new query. And for this cases child results are created. These child instances share the same file for reading the result. Despite multiple children can be created and each this child is used by it's own thread, this construction looks thread safe because file is effectively final - no more data can be written to it when child is allowed to be created. When we've done with all children, the underlying file should be closed - so we need to have some bookkeeping of the results which use the same offload file.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+import static java.nio.file.StandardOpenOption.CREATE_NEW;
+import static java.nio.file.StandardOpenOption.READ;
+import static java.nio.file.StandardOpenOption.WRITE;
+
+/**
+ * Disk-based hash index for the sorted external result.
+ * Open addressing hashing scheme is used.
+ */
+@SuppressWarnings("TypeMayBeWeakened")
+public class ExternalResultHashIndex implements AutoCloseable {
```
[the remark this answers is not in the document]
rkondakov: I was thinking about it when designed the solution. Two main reasons why I decided to  introduce additional index `ExternalResultHashIndex`:

1. It was undesirable to change H2 native code more than it is needed for solving the current tasks. It is caused by the difficulties of ensuring that H2 code works as expected. If we break something in H2 code, it is very difficult to detect the problem. So, I decided to implement `org.h2.result.ResultExternal` interface without changing it at all and with a minors changes to 'H2ManagedLocalResult' and other classes which are used by H2 engine. And in this case some methods like `org.h2.result.ResultExternal#addRow` can not be implemented without fast duplicates detection which is done by  `ExternalResultHashIndex`.
2. Methods like 'org.h2.result.ResultExternal#removeRow'  and `org.h2.result.ResultExternal#contains` should work fast, and `ExternalResultHashIndex` contributes to. If we use LSM style for finding duplicates, we should check all sorted chunks on the diskm which not be as fast as  `ExternalResultHashIndex`. So it can lead to performance issues for some types of queries as well.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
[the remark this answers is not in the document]
rkondakov: But it looks like the same the same FileIO shouldn't be shared between all children - each child should open it's own FileIO. I'll fix it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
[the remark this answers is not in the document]
akalash: It doesn't matter what you read. My question is why do you read at least something from disk? As I can see you don't use this loaded data at all. It would make sense if you are doing the consistency checking of data but you don't have such a feature.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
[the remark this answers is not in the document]
akalash: 1. As I understand you need to guarantee that org.h2.result.ResultExternal#next wouldn't return duplication but you must not do it exactly in org.h2.result.ResultExternal#addRow
2. org.h2.result.ResultExternal#removeRow it has at least two fetches from disk so it certainly will be relatively slow.
LSM style does the same thing which you already did in your SortedExternalResult#next, doesn't it? You need just added one more check that 'next != current'.

My point of view - you should do as simpler as possible if you don't have any proofs to do more complicated. 
LSM way will be slower only in one scenario when we have a lot of same keys which fall into different chunks but how real is this scenario is? I mean now you have class ExternalResultHashIndex which complex enough and involve one more IO operation with the rollover of file. I concerned that it can lead to new bugs(usually this lead).
Anyway, it depends on your expected load profile and maybe it makes sense to do benchmark on both approach(as I understand the implementation of LSM way should not be difficult on current codebase).

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1126,6 +1126,32 @@ public static int ceilPow2(int v) {
...
 
+    /**
+     * Gets nearest power of 2 larger or equal than v.
+     *
+     * Copied from <a href="https://graphics.stanford.edu/~seander/bithacks.html#RoundUpPowerOf2">bithacks</a>.
+     *
+     * @param v Value.
+     * @return Nearest power of 2.
+     */
+    public static long ceilPow2Long(long v) {
```
akalash: Long#highestOneBit

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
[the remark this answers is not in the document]
akalash: I only now figured out your comment about org.h2.result.ResultExternal#contains so it is a good point. Maybe hash table it is better variant. I try to figure out it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+
+    /** {@inheritDoc} */
+    @Override public int removeRow(Value[] values) {
+        ValueRow key = getRowKey(values);
+
+        if (sortedRowsBuf != null) {
+            Object prev = sortedRowsBuf.remove(key);
+
+            if (prev != null)
+                return size--;
```
ibessonov: Here and in line 247 you should write "--size".

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+            spillRowsBufferToDisk();
+
+        return size++;
+    }
+
+    /**
+     * Checks if current result contains given row with sort order check.
+     *
+     * @param row Row.
+     * @return {@code True} if current result does not contain th given row.
```
ibessonov: typo

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+    }
+
+    /**
+     * Spills rows to disk from the in-memory buffer.
+     */
+    private void spillRowsBufferToDisk() {
+        if (F.isEmpty(sortedRowsBuf) && F.isEmpty(unsortedRowsBuf))
+            return;
+
+        ArrayList<Value[]> rows = isAnyDistinct() ? new ArrayList<>(sortedRowsBuf.values()) : unsortedRowsBuf;
```
ibessonov: Type of the variable should be just "List<Value[]>"

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+            this.curPos = start;
+            this.end = end;
+        }
+
+        /**
+         * @return {@code True} if next row is available within a chunk.
+         */
+        boolean next() {
+            while (curPos < end) {
+                curRow = readRowFromFile(curPos);
```
ibessonov: Removed entry will be counted as the end of file, right?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1982,6 +1989,27 @@ public QueryMemoryManager memoryManager() {
...
+     * offloading of the intermediate results of heavy queries.
+     */
+    private void cleanSpillDirectory() {
+        try {
+            File spillDir = U.resolveWorkDirectory(
+                ctx.config().getWorkDirectory(),
+                DISK_SPILL_DIR,
+                false);
+
+            U.delete(spillDir);
```
pavlukhin: I remember we discussed multiple nodes sharing same work directory case. Will it work as expected in such case? Should not we add a describing comment?

modules/h2/src/main/java/org/apache/ignite/internal/processors/query/h2/H2MemoryTracker.java
```diff
@@ -16,21 +16,68 @@
...
+import java.util.ArrayList;
+import java.util.Collection;
+
 /**
- * Abstract memory tracker.
+ * Memory tracker.
  */
 public abstract class H2MemoryTracker implements AutoCloseable {
+    /** Objects to be closed along with the current tracker. */
+    private Collection<Closeable> closeList;
```
pavlukhin: Won't `AutoCloseable` be enough here instead of `Closeable`?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,546 @@
...
+     */
+    private boolean isAnyDistinct() {
+        return distinct || distinctIndexes != null;
+    }
 
-            for (int i = 0; i < oldRow.length; i++)
-                memory -= oldRow[i].getMemory();
+    /** {@inheritDoc} */
+    @Override public void removeDistinct(Value[] values) {
+        if (!distinct) {
```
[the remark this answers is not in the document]
pavlukhin: The code here is almost copied from h2 `org.h2.result.LocalResultImpl`. We need to decide whether it should be close to an original or not.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,546 @@
...
+        if (isAnyDistinct()) {
+            if (distinctRows != null) {
+                ValueRow array = getDistinctRow(values);
+                Value[] previous = distinctRows.get(array);
+                if (previous == null || sort != null && sort.compare(previous, values) > 0) {
+                    distinctRows.put(array, values);
+                }
+                rowCount = distinctRows.size();
+                if (!isEnoughMemoryAfterReservation(array, previous, values)) {
+                    createExternalResult();
```
pavlukhin: I suppose the code is borrowed from h2 but I suspect that external result could be recreated here if one already exists. I would good to ensure somehow that there is no mistake. Simple thing coming to mind is `assert external == null`.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+            long oldSize = cap;
+
+            initNewIndexFile(oldSize * 2);
+
+            copyDataFromOldFile(oldFileIo, oldIdxFile, oldSize);
+        }
+    }
+
+    /**
+     * Copies data from the ols file to the extended new one.
```
pavlukhin: ols -> old

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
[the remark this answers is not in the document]
pavlukhin: I tend to agree with Anton here. Concrete external result implementations use _file-related_ operations defined in this class. Implementations does override _file-related_ methods and there is no _template_ methods. Decomposition itself would be a benefit here. If _file-related_ operations were placed in a separate class (e.g. `ExternalResultIO`) it would be obvious that it does not share a state with any `ResultExternal` implementation. Otherwise a code reader should spend a time to figure out mentioned independence (like I did).

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/PlainExternalResult.java
```diff
@@ -0,0 +1,100 @@
...
+    @Override public int addRows(Collection<Value[]> rows) {
+        if (rows.isEmpty())
+            return size;
+
+        Data buff = createDataBuffer(rowSize(rows));
+
+        for (Value[] row : rows)
+            addRowToBuffer(row, buff);
+
+        writeBufferToFile(buff);
```
pavlukhin: It seems that `buff` filled with all rows can double an amount of memory a query can use (`H2ManagedLocalResult#rows` + same rows in `buff` here). Is not it a problem?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+    @Override public int addRows(Collection<Value[]> rows) {
+        for (Value[] row : rows)
+            addRow(row);
+
+        return size;
+    }
+
+    /** {@inheritDoc} */
+    @Override public int addRow(Value[] row) {
+        if (isAnyDistinct()) {
```
pavlukhin: ` if (isAnyDistinct() && containsRowWithOrderCheck(row)) { return size; }`

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+
+    /**
+     * @return {@code True} if it is need to spill rows to disk.
+     */
+    private boolean needToSpill() {
+        return !memTracker.reserved(0);
+    }
+
+    /**
+     * Adds row in-memory row buffer.
```
pavlukhin: _Adds row to..._

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+
+    /**
+     * Result queue.
+     */
+    private Queue<Chunk> resQueue;
+
+    /**
+     * Comparator for {@code sortedRowsBuf}.
+     */
+    private Comparator<Value> cmp;
```
pavlukhin: It would be great to leave a comment describing a difference between `cmp` and `sort` fields use. Both are Comparators  but as a reader I cannot figure out why do we need two.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+        long initFilePos = lastWrittenPos;
+
+        for (Value[] row : rows) {
+            if(isAnyDistinct()) {
+                long rowPosInBuff = buff.length();
+
+                addRowToHashIndex(row, initFilePos + rowPosInBuff);
+            }
+
+            addRowToBuffer(row, buff);
```
pavlukhin: This call confused me: here we are spilling rows to disk but current line adds them back! I spent some time to realize that there are different `addRowToBuffer` methods. I suggest to rename one of them, e.g. `AbstractExternalResult#addRowToBuffer(org.h2.value.Value[], org.h2.store.Data)` to `addRowToWriteBuffer`.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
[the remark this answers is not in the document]
pavlukhin: @akalash, are you *definitely* sure that LSM approach fits significantly better here? Are the some crucial flaws in a current design? Are there any blockers to come up with a better solution some time later? I suggest to proceed with a current design if it is good enough.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+    private void addRowToBuffer(Value[] row) {
+        long delta = H2Utils.calculateMemoryDelta(null, null, row);
+
+        memTracker.reserved(delta);
+
+        if (isAnyDistinct()) {
+            assert unsortedRowsBuf == null;
+
+            if (sortedRowsBuf == null)
+                sortedRowsBuf = new TreeMap<>(cmp);
```
pavlukhin: Clarifying question: why do we use `TreeMap` here? Is it only in order to provide custom `Comparator`? I think about renaming `sortedRowsBuf` to `mapRowsBuf` if a sorting order of the map is not used.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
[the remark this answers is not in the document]
akalash: According to the requirement of using 'contains' method current solution is good enough for me. Although I still have some concerns about it cause it makes code more difficult but currently, I don't have any other solution.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+    private File idxFile;
+
+    /** Index file channel. */
+    private FileIO fileIo;
+
+    /** Row store. */
+    private final SortedExternalResult rowStore;
+
+    /** Reusable byte buffer for reading and writing. We use this only instance just for prevention GC pressure. */
+    private final ByteBuffer reusableBuff = ByteBuffer.allocate(Entry.ENTRY_BYTES);
```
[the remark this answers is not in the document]
rkondakov: This is because the entry size is fixed in ExternalResultHashIndex. Entries in AbstractExternalResult may have arbitrary sizes.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
rkondakov: Agree

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,546 @@
...
+        if (memory < 0) {
+            memTracker.released(-memory);
+
+            return true;
+        }
+        else
+            return memTracker.reserved(memory);
+    }
+
+    @Override public boolean isLazy() {
```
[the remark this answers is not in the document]
rkondakov: ok

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+    ExternalResultHashIndex(GridKernalContext ctx, File spillFile, SortedExternalResult rowStore, long initSize) {
+        this.ctx = ctx;
+        dir = spillFile.getParent();
+        spillFileName = spillFile.getName();
+        this.rowStore = rowStore;
+
+        if (initSize <= MIN_CAPACITY)
+            initSize = MIN_CAPACITY;
+
+        long initCap = U.ceilPow2Long(initSize * 2);
```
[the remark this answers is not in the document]
rkondakov: Technically you are right. But I am not expect tables with 1^59 entries in the real life. But ok, I'll add check for this.
To minimize the number collision  we need to have HashTable capacity at least twice bigger than the entries number. This is the reason of multiplying by 2. But it's needed to be long initCap = U.ceilPow2Long(initSize) * 2

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+        initNewIndexFile(initCap);
+    }
+
+    /**
+     * Puts row into the hash index.
+     *
+     * @param key Row distinct key.
+     * @param rowAddr Row address in the rows file.
+     */
+    public void put(ValueRow key, Long rowAddr) {
```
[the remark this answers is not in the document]
rkondakov: You're right, this is the wrong type of method parameter. I didn't replaced it with the primitive while I did it in other places. So, I'll fix it.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+     * @param key Row distinct key
+     * @return Row address in the rows file.
+     */
+    public Long remove(ValueRow key) {
+        Entry entry = findEntry(key);
+
+        if (entry == null)
+            return null;
+
+        writeEntryToIndexFile(entry.slot(), key.hashCode(), -2); // row addr == -2 means removed row.
```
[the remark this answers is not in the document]
rkondakov: Agree.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+     * @param hashCode Hash code.
+     * @return Free slot number.
+     */
+    private long findFreeSlotForInsert(int hashCode) {
+        long slot = slot(hashCode);
+
+        Entry entry;
+
+        while ((entry = readEntryFromIndexFile(slot)) != null && !entry.isRemoved() && !entry.isEmpty()) {
+            slot = slot((int)(slot + 1)); // Check next slot.
```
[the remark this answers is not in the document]
rkondakov: ok

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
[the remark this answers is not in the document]
rkondakov: I'm not sure I got you right. Why do you think it is lie? If you didn't find a free slot in the hashmap it means that there is no free space left there. So, it is not a lie.
I may agree that it is better to use `assert` instead of `RuntimeException`.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+    }
+
+    /**
+     * Ideal slot for the given hashcode: <code>slot = hashcode % capacity</code>.
+     *
+     * @param hashCode Hashcode.
+     * @return Ideal slot.
+     */
+    private long slot(int hashCode) {
+        return hashCode & (cap - 1);
```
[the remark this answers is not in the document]
rkondakov: Agree. Fixed.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+     * Checks if the capacity of the hashtable is enough and extends it if needed.
+     */
+    private void ensureCapacity() {
+        if (entriesCnt > LOAD_FACTOR * cap) {
+            FileIO oldFileIo = fileIo;
+            File oldIdxFile = idxFile;
+
+            long oldSize = cap;
+
+            initNewIndexFile(oldSize * 2);
```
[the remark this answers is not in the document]
rkondakov: ok

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
rkondakov: H2 API requires ArrayList. See `org.h2.result.SortOrder#sort(java.util.ArrayList<org.h2.value.Value[]>)`

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
rkondakov: No, removed entries will be skipped until we reach non removed entry or the end of chunk.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
rkondakov: `U.delete(spillDir)` tries to remove all files and directories from spillDir. If it is not succeed (because file is being used by somebody), it ignores IOException and stops removing the files. So, it cleans as much outdated files as possible.

modules/h2/src/main/java/org/apache/ignite/internal/processors/query/h2/H2MemoryTracker.java
rkondakov: Agree.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
rkondakov: Actually it cannot be recreated because there is of invariant `distinctRows != null XOR external != null`. So only either of them is not null.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/PlainExternalResult.java
rkondakov: Looks like it might be a problem. Good point. Thank you.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
rkondakov: ok

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
rkondakov: No, because this result is already fully spilled on disk and it doesn't consume memory.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/AbstractExternalResult.java
```diff
@@ -0,0 +1,392 @@
...
+            }
+
+            rowBytes.flip();
+
+            return rowBytes;
+        }
+        catch (IOException e) {
+            close();
+
+            throw new IgniteException("Failed to read query result the from spill file.", e);
```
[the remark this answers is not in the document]
rkondakov: Because this exception can cot be handled by H2 code. We need to give up an entire operation if something went wrong.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,469 @@
...
+        Entry entry = readEntryFromIndexFile(slot);
+
+        while (!entry.isEmpty()) {
+            if (hashCode == entry.hashCode()) { // 1. Check hashcode equals.
+                Value[] row = rowStore.readRowFromFile(entry.rowAddress());
+
+                if (row != null) {
+                    ValueRow keyFromDisk = rowStore.getRowKey(row);
+
+                    if (key.equals(keyFromDisk)) // 2. Check the actual row equals the given key.
```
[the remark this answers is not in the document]
rkondakov: It happens only when hashcodes are equal - we don't read entire row for each comparison.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/SortedExternalResult.java
```diff
@@ -0,0 +1,473 @@
...
+        long written = writeBufferToFile(buff);
+
+        lastWrittenPos = initFilePos + written;
+
+        chunks.add(new Chunk(initFilePos, lastWrittenPos));
+
+        for (Value[] row : rows) {
+            long delta = H2Utils.calculateMemoryDelta(null, row, null);
+
+            memTracker.released(-delta);
```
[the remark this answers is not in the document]
rkondakov: done

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryTracker.java
```diff
@@ -54,47 +66,63 @@
     QueryMemoryTracker(H2MemoryTracker parent, long maxMem, long blockSize) {
         assert maxMem > 0;
 
+        // TODO GG-18629 - get from configuration.
+        failOnMemLimitExceed = !Boolean.getBoolean(IgniteSystemProperties.IGNITE_SQL_USE_DISK_OFFLOAD);
```
AMashenkov: This lookup to HashTable may costs.
Let's move to static field or outside constructor (e.g. to memory manager).

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
```diff
@@ -0,0 +1,512 @@
...
+        this.fileIOFactory = fileIOFactory;
+        dir = spillFile.getParent();
+        spillFileName = spillFile.getName();
+        this.rowStore = rowStore;
+
+        if (initSize <= MIN_CAPACITY)
+            initSize = MIN_CAPACITY;
+
+        // We need at least the half of hash map be empty to minimize collisions number.
+        long initCap = Long.highestOneBit(initSize) * 2;
```
ibessonov: With this code you'll have at least half space being used (not at most). I think you should multiply it by 4 to get what you want.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/DiskSpillingBasicTest.java
```diff
@@ -0,0 +1,229 @@
...
+        for (int i = 0; i < NUM_THREADS; i++)
+            assertTrue("Thread with not executed queries found.", cntrs.get(i) > 0);
+
+        assertWorkDirClean();
+    }
+
+    /** */
+    @Test
+    public void testFilesDeletedOnError() throws IOException {
+        String query = "SELECT DISTINCT *, 1 / (p.id - 800) " + // Query should throw NPE during execution.
```
mshonichev: Division by zero: "1"; SQL statement

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultData.java
```diff
@@ -0,0 +1,500 @@
...
+    Collection<Chunk> chunks() {
+        return chunks;
+    }
+
+    /** {@inheritDoc} */
+    @Override public void close() {
+        U.closeQuiet(fileIo);
+        file.delete();
+
+        U.closeQuiet(hashIdx);
```
akalash: Maybe better to surround this code by try-finally because first lines of code can throw IoExeption.
```
try {
        U.closeQuiet(fileIo);
        file.delete();
} finally {
      U.closeQuiet(hashIdx);
}
```

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryTracker.java
rkondakov: Agree. Moved to memory manager.

modules/h2/src/main/java/org/h2/expression/Subquery.java
```diff
@@ -51,6 +53,11 @@ public Value getValue(Session session) {
             }
             return v;
         }
+        finally {
+            // Do not the very first last result in the case when caching is turned on. Otherwise close all results.
```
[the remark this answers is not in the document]
rkondakov: Fixed. The word 'close' was missed.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,575 @@
...
+            memTracker.released(-memory);
+
+            return true;
+        }
+        else {
+            try {
+                return memTracker.reserved(memory);
+            }
+            catch (Throwable e) {
+                memReserved -= memory;
```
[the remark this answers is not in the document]
rkondakov: Yes, the insufficient memory budget can be the reason of this exception. I've refactored this snippet to avoid using subtraction in the 'catch' clause.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,575 @@
...
+    @Override public boolean isLazy() {
+        return false;
+    }
 
-        // Replace old row.
-        if (oldRow != null) {
-            memory = (row.length - oldRow.length) * Constants.MEMORY_POINTER;
+    /** {@inheritDoc} */
+    @Override public void setMaxMemoryRows(int maxValue) {
+        // No-op.
```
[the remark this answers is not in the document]
rkondakov: We do not use rowCount-based memory tracking in this class as H2 does. 'H2MemoryTracker' is used instead. I've added a comment about this.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultHashIndex.java
rkondakov: Yes. Good point, thank you. Fixed.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/DiskSpillingBasicTest.java
rkondakov: You are right. Fixed.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/ExternalResultData.java
rkondakov: Agree. I moved 'U.closeQuiet(hashIdx)' to the second line right after 'U.closeQuiet(fileIo)' which can not throw 'IOException".

modules/h2/src/main/java/org/h2/command/dml/Call.java
```diff
@@ -66,7 +65,7 @@ public ResultInterface query(int maxrows) {
         if (isResultSet) {
             return v.getResult();
         }
-        LocalResult result = LocalResultFactory.DEFAULT.create(session, expressions, 1);
+        LocalResult result = session.getDatabase().getResultFactory().create(session, expressions, 1, true);
```
AMashenkov: @tledkov-gridgain, please re-check this.
Looks like this cause unneeded MemoryTracker creation and memory allocation.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2ManagedLocalResult.java
```diff
@@ -43,71 +86,568 @@
...
+    /**
+     * Adds rows to disk.
+     */
+    private void addRowsToDisk() {
+        if (external == null) {
+            createExternalResult();
+        }
+
+        // We need to release memory before adding to external result to prevent immediate rows spilling.
+        if (memTracker != null)
```
AMashenkov: Why you free memory before actual 'rows.clear()' call?

rkondakov: This is because we are about to put these rows into the external result. If we not release memory before putting row into external result, external result is thinking that we have no available memory at all and start spilling rows to disk one-by-one instead of grouping them into the batches.

rkondakov: This is because we are about to put these rows into the external result. If we not release memory before putting row into external result, external result would be thinking that we have no available memory at all and would start spilling rows to disk one-by-one instead of grouping them into the batches.

PR #303, opened 2019-07-22
GG-20804 PME test optinization

ingvard:


modules/core/src/test/java/org/apache/ignite/cache/store/jdbc/CacheJdbcPojoWriteBehindStoreWithCoalescingTest.java
```diff
@@ -552,7 +557,7 @@ public void readWithCoalescing() throws Exception {
             @Override public void run() {
                 try {
                     while (t2Count.get() > 0) {
-                        for (int i = 200000; i < 400000; i++) {
+                        for (int i = stepSize; i < 2* stepSize; i++) {
```
dspavlov: 2 * <-spac is missing between 2 and *

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/CacheLateAffinityAssignmentTest.java
```diff
@@ -2134,7 +2134,7 @@ public void testNoForceKeysRequests() throws Exception {
 
         final AtomicInteger nodeIdx = new AtomicInteger();
 
-        final long stopTime = System.currentTimeMillis() + 60_000;
+        final long stopTime = System.currentTimeMillis() + GridTestUtils.SF.applyUB(400_000, 60_000);
```
dspavlov: It seems test stop time is raised here to 400 seconds. Should we reduce it?

ingvard: No, this means that max time 60 seconds and min 400_000 * 0.1 miles = 40 seconds.

modules/core/src/test/java/org/apache/ignite/cache/store/jdbc/CacheJdbcPojoWriteBehindStoreWithCoalescingTest.java
ingvard: fixed

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/CacheLateAffinityAssignmentTest.java
ingvard: fixed

PR #312, opened 2019-07-23
GG 21272

antonovsergey93:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridTransactionsSystemUserTimeMetricsTest.java
```diff
@@ -0,0 +1,263 @@
...
+    /** */
+    private static String longOpTimeoutCommon;
+
+    /** */
+    private volatile boolean slowPrepare;
+
+    /** {@inheritDoc} */
+    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        ListeningTestLogger testLog =
+            new ListeningTestLogger(false, super.getConfiguration(igniteInstanceName).getGridLogger());
```
antonovsergey93: You could use `GridAbstractTest.log` field

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/LongRunningTxTimeDumpSettingsClosure.java
```diff
@@ -0,0 +1,68 @@
...
+    private final Float samplesPercentage;
+
+    /**
+     * Auto-inject Ignite instance
+     */
+    @IgniteInstanceResource
+    private Ignite ignite;
+
+    /** */
+    public LongRunningTxTimeDumpSettingsClosure(Long timeoutThreshold, Float samplesPercentage) {
```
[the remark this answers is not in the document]
denis-chudov: Are you sure about that? In current usages thay will never be both null, but if this theoretically happens, no actions will be performed, but also there will be no errors in log.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
```diff
@@ -3744,10 +3804,91 @@ private void readyNearLock(IgniteTxEntry txEntry,
...
+    @Override public boolean state(TransactionState state) {
+        boolean res = super.state(state);
+
+        if (state == COMMITTED || state == ROLLED_BACK) {
+            leaveSystemSection();
+
+            commitOrRollbackTime.compareAndSet(0, System.nanoTime() - commitOrRollbackStartTime.get());
+
+            long systemTimeMillis = this.systemTime.get() / 1_000_000;
+            long totalTimeMillis = System.currentTimeMillis() - startTime();
```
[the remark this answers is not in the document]
denis-chudov: startTime() returns timestamp of the transaction start in milliseconds,  so it is not an error.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
```diff
@@ -3744,10 +3804,91 @@ private void readyNearLock(IgniteTxEntry txEntry,
...
+            long userTimeMillis = totalTimeMillis - systemTimeMillis;
+
+            writeTxMetrics(systemTimeMillis, userTimeMillis);
+
+            if (cctx.tm().longTransactionTimeDumpThreshold() > 0
+                && totalTimeMillis > cctx.tm().longTransactionTimeDumpThreshold()
+                && RANDOM.nextDouble() <= cctx.tm().longTransactionTimeDumpSampleLimit()) {
+                DateFormat timeFormat = new SimpleDateFormat("HH:mm:ss.SSS");
+                timeFormat.setTimeZone(TimeZone.getTimeZone("UTC"));
```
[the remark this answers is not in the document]
denis-chudov: If I will not, zero timestamp will be formatted as "03:00:00" for Moscow, as formatting depends on time zone.

denis-chudov:
> Why you use AtomicLong#compareAndSet(long, long) method and don't check returned value?

I rethinked one usage of this, now I check the result of compareAndSet in GridNearTxLocal#state() but in other places there is only reason to set this values only once and then reset in other place, no need to check the result of CAS.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridTransactionsSystemUserTimeMetricsTest.java
```diff
@@ -0,0 +1,262 @@
...
+
+        cfg.setCommunicationSpi(new TestCommunicationSpi());
+
+        return cfg;
+    }
+
+    /**
+     * Setting long op timeout to small value to make this tests faster
+     */
+    @Override protected void beforeTestsStarted() throws Exception {
```
antonovsergey93: Use @WithSystemProperty for setting system property before test(s) and removing it after test(s)

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridTransactionsSystemUserTimeMetricsTest.java
```diff
@@ -0,0 +1,262 @@
...
+        longOpTimeoutCommon = System.getProperty(IGNITE_LONG_OPERATIONS_DUMP_TIMEOUT);
+
+        System.setProperty(IGNITE_LONG_TRANSACTION_TIME_DUMP_THRESHOLD, String.valueOf(LONG_TRAN_TIMEOUT));
+        System.setProperty(IGNITE_LONG_TRANSACTION_TIME_DUMP_SAMPLE_LIMIT, String.valueOf(1.0f));
+        System.setProperty(IGNITE_LONG_OPERATIONS_DUMP_TIMEOUT, String.valueOf(LONG_OP_TIMEOUT));
+    }
+
+    /**
+     * Returning long operations timeout to its former value.
+     */
```
antonovsergey93: Use @WithSystemProperty for setting system property before test(s) and removing it after test(s)

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/GridCommonAbstractTest.java
```diff
@@ -2357,4 +2361,32 @@ protected void assertCountersSame(int partId, boolean withReserveCntr) throws As
             cntr0 = cntr.get2();
         }
     }
+
+    /**
+     * Returns MX bean by specified group name and class.
```
antonovsergey93: Please move method to GridAbstractTest, if it's possible

modules/core/src/test/java/org/apache/ignite/internal/processors/metrics/impl/MetricUtilsTest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import org.apache.ignite.internal.processors.metric.impl.HistogramMetric;
+import org.junit.Test;
+
+import static org.apache.ignite.internal.processors.metric.impl.MetricUtils.toJson;
+
+/**
+ *
+ */
+public class MetricUtilsTest {
```
antonovsergey93: I don't see any assert in this test class.

modules/core/src/test/java/org/apache/ignite/testframework/junits/common/GridCommonAbstractTest.java
denis-chudov: It is possible, but likely makes no sense as GridAbstractTest has only one child (GridSpiAbstractTest) and for now I see no reason that it (or it's children) needs this method

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -6031,6 +6032,45 @@ public void setTxOwnerDumpRequestsAllowed(boolean allowed) {
...
+    public void longTransactionTimeDumpThreshold(long threshold) {
+        assert threshold >= 0 : "Threshold timeout must be greater than or equal to 0.";
+
+        ClusterGroup grp = ctx.grid()
+                .cluster()
+                .forPredicate(node -> IgniteFeatures.nodeSupports(ctx, node, LRT_SYSTEM_USER_TIME_DUMP_SETTINGS));
+
+        IgniteCompute compute = ctx.grid().compute(grp);
+
+        compute.broadcast(new LongRunningTxTimeDumpSettingsClosure(threshold, null));
```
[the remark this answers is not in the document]
denis-chudov: There is no propagation of any jmx settings from TransactionMxBean to new joining nodes. After discussion we decided to leave it as is and create a new ticket for this problem.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -6031,6 +6032,45 @@ public void setTxOwnerDumpRequestsAllowed(boolean allowed) {
...
+    }
+
+    /**
+     * Sets the percentage of samples of long running transactions that will be dumped in log, if
+     * {@link #longTransactionTimeDumpThreshold} is set to non-zero value."
+     *
+     * @param percentage Percentage, must be value between 0.0 and 1.0 inclusively.
+     */
+    public void longTransactionTimeDumpSampleLimit(double percentage) {
+        assert percentage >= 0.0 && percentage <= 1.0 : "Percentage value must be between 0.0 and 1.0 inclusively.";
```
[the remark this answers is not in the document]
denis-chudov: There is no propagation of any jmx settings from TransactionMxBean to new joining nodes. After discussion we decided to leave it as is and create a new ticket for this problem.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearTxLocal.java
```diff
@@ -3744,10 +3809,121 @@ private void readyNearLock(IgniteTxEntry txEntry,
...
+
+            long transactionTimeDumpThreshold = cctx.tm().longTransactionTimeDumpThreshold();
+
+            double transactionTimeDumpSamplesCoefficient = cctx.tm().transactionTimeDumpSamplesCoefficient();
+
+            boolean isLong = transactionTimeDumpThreshold > 0 && totalTimeMillis > transactionTimeDumpThreshold;
+
+            boolean randomlyChosen = transactionTimeDumpSamplesCoefficient >= 0.0
+                && RANDOM.nextDouble() <= transactionTimeDumpSamplesCoefficient
+                && transactionDumpHitRateCounter != null
```
antonovsergey93: Why `transactionDumpHitRateCounter` could be null? You should add `@Nullable` to costructor args or remove null check from condition.

denis-chudov: Because there is a default constructor and also mock child class that uses multiple null arguments for constructor.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1284,6 +1284,30 @@
      */
     public static final String INDEX_REBUILDING_PARALLELISM = "INDEX_REBUILDING_PARALLELISM";
 
+    /**
+     * Threshold timeout for long transactions, if transaction exceeds it, it will be dumped in log with
+     * information about how much time did it spent in system time (time while aquiring locks, preparing,
+     * commiting, etc) and user time (time when client node runs some code while holding transaction and not
+     * waiting it). Equals 0 if not set. No long transactions are dumped in log if nor this parameter
+     * neither {@link #IGNITE_TRANSACTION_TIME_DUMP_SAMPLES_COEFFICIENT} is set.
```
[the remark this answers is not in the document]
denis-chudov: No long transactions are dumped in log if nor this parameter neither {@link #IGNITE_TRANSACTION_TIME_DUMP_SAMPLES_COEFFICIENT} is set.

PR #316, opened 2019-07-24
GG-21710 Added processing of additional connection errors.

vsisko:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
```diff
@@ -172,10 +181,18 @@
...
-            log.error("You are using outdated version of Web agent. Please download latest version of Web agent from Web console");
+        if (ioCause != null) {
+            if ("404".equals(ioCause.getMessage())) {
+                log.error("You are using outdated version of Web agent. Please download latest version of Web agent from Web console.");
 
-            System.exit(1);
+                System.exit(1);
+            }
+
+            if ("504".equals(ioCause.getMessage())) {
```
nva: Why this check not in `onError`

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
```diff
@@ -172,10 +181,18 @@
 
         IOException ioCause = X.cause(e, IOException.class);
 
-        if (ioCause != null && "404".equals(ioCause.getMessage())) {
-            log.error("You are using outdated version of Web agent. Please download latest version of Web agent from Web console");
+        if (ioCause != null) {
+            if ("404".equals(ioCause.getMessage())) {
```
nva: Why this check not in `onError`

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
```diff
@@ -172,10 +181,18 @@
...
+                log.error("You are using outdated version of Web agent. Please download latest version of Web agent from Web console.");
```
nva: This error also can happens when you are trying to connect to different application instead of web console

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
vsisko: Added check of 504 error in case of error event. We show different message for error and connection error event.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
vsisko: We show different message for error and connection error event.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
vsisko: Message changed.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
nva: I don't see any reason for separate error message for this case please remove

vsisko: Removed.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentLauncher.java
```diff
@@ -152,14 +153,30 @@
                 return;
             }
 
-            IOException ignore = X.cause(e, IOException.class);
+            EOFException eof = X.cause(e, EOFException.class);
 
-            if (ignore != null && "404".equals(ignore.getMessage())) {
-                log.error("Failed to receive response from server (connection refused).");
+            if (eof != null) {
```
nva: Use `X.hasCause` instead.

vsisko: Used X.hasCause.

PR #319, opened 2019-07-24
NBL-43: CPP: Implement missing methods for IgniteCluster class

FedorUporov:


modules/platforms/cpp/core-test/src/cache_test.cpp
```diff
@@ -132,6 +132,43 @@ struct CacheTestSuiteFixture
...
+    cache::Cache<int, int> Cache()
+    {
+        return grid0.GetCache<int, int>("partitioned");
+    }
+
+    /*
+     * Constructor.
+     */
+    CacheNativePersistenceTestSuiteFixture()
+    {
```
isapego: I think we should use `ClearLfs()` utility method here. You can copy it from thin client tests or ODBC tests.

FedorUporov: Added.

modules/platforms/cpp/core-test/src/cache_test.cpp
```diff
@@ -729,3 +766,26 @@ BOOST_AUTO_TEST_CASE(TestPutGetStructWithEnumField)
...
+
+    BOOST_REQUIRE(cluster.IsWalEnabled(cache.GetName()));
+    cluster.DisableWal(cache.GetName());
+    BOOST_REQUIRE(!cluster.IsWalEnabled(cache.GetName()));
+    cluster.EnableWal(cache.GetName());
+    BOOST_REQUIRE(cluster.IsWalEnabled(cache.GetName()));
+
+    BOOST_CHECK_THROW(cluster.IsWalEnabled("foo"), IgniteError);
+    BOOST_CHECK_THROW(cluster.DisableWal("foo"), IgniteError);
+    BOOST_CHECK_THROW(cluster.EnableWal("foo"), IgniteError);
```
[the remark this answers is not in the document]
FedorUporov: The positive cases are above:
   BOOST_REQUIRE(cluster.IsWalEnabled(cache.GetName()));
    cluster.DisableWal(cache.GetName());
    BOOST_REQUIRE(!cluster.IsWalEnabled(cache.GetName()));
    cluster.EnableWal(cache.GetName());
    BOOST_REQUIRE(cluster.IsWalEnabled(cache.GetName()));

Or, you want to add something more?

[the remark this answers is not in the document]
isapego: Ok, it seems I have missed `(Disable|Enable)Wal` part. Sorry for that

modules/platforms/cpp/core-test/src/cache_test.cpp
```diff
@@ -166,6 +166,8 @@ struct CacheNativePersistenceTestSuiteFixture
         grid0 = Ignite();
 
         Ignition::StopAll(true);
+
+        ignite_test::ClearLfs();
```
isapego: We should also do it in constructor, just in case someone forgot to clean LFS previously.

PR #320, opened 2019-07-25
GG-20545: Introduce partition data TTL

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/PartitionDataStorage.java
```diff
@@ -61,4 +79,43 @@
...
+         *
+         * @param part Partition number.
+         */
+        public Cleaner(int part) {
+            this.part = part;
+        }
+
+        /** {@inheritDoc} */
+        @Override public void run() {
+            Object data = storage.remove(part);
```
avplatonov: What about locking here?

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -54,8 +57,11 @@
         loggingFactory = constant(NoOpLogger.factory());
         seed = constant(new Random().nextLong());
         rngSupplier = p -> new Random();
+        dataTtl = -1;
```
avplatonov: named constant please

modules/ml/src/main/java/org/apache/ignite/ml/dataset/impl/cache/util/PartitionDataStorage.java
```diff
@@ -43,22 +60,135 @@
...
+
+        /**
+         * Reschedules this cleaner using the last access time as a reference point. The next attepmt to clean up will
+         * be {@link #ttl} milliseconds after the last access.
+         *
+         * @param lastAccessTime Last access time in milliseconds (difference between the last access time and midnight,
+         *                       January 1, 1970 UTC).
+         */
+        private void reschedule(long lastAccessTime) {
+            long delay = lastAccessTime - System.currentTimeMillis() + ttl;
```
avplatonov: lets add check "delay < 0"

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -54,8 +60,11 @@
         loggingFactory = constant(NoOpLogger.factory());
         seed = constant(new Random().nextLong());
         rngSupplier = p -> new Random();
+        dataTtl = DEFAULT_TTL;
```
avplatonov: When I read "dataTtl = INFINITY_TTL" I can understand what you mean. In case of "DEFAULT_TTL"-name I can't

examples/src/main/java/org/apache/ignite/examples/ml/knn/KNNClassificationExample.java
```diff
@@ -60,7 +60,8 @@ public static void main(String[] args) throws IOException {
                 KNNClassificationTrainer trainer = new KNNClassificationTrainer()
                     .withK(3)
                     .withDistanceMeasure(new EuclideanDistance())
-                    .withWeighted(true);
+                    .withWeighted(true)
+                    .withDataTtl(60);
```
zaleslaw: I suggest to move it to DataSet builder or to DataConfiguration (it seems strange in Trainer)

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -34,6 +34,9 @@
     /** Serial version id. */
     private static final long serialVersionUID = 8502532880517447662L;
 
+    /** Default partition data TTL (infinite). */
+    private static final long INFINITE_TTL = -1;
```
zaleslaw: ENUM - ENUM, it sounds like ENUM, you love old C style)))

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
```diff
@@ -114,7 +130,7 @@ private static ParallelismStrategy strategyByType(ParallelismStrategy.Type stgyT
     @Override public LearningEnvironment buildForWorker(int part) {
         Random random = rngSupplier.apply(part);
         random.setSeed(seed.apply(part));
-        return new LearningEnvironmentImpl(part, random, parallelismStgy.apply(part), loggingFactory.apply(part));
+        return new LearningEnvironmentImpl(part, dataTtl, random, parallelismStgy.apply(part),  loggingFactory.apply(part));
```
zaleslaw: So long so large the constructor of LEI, maybe we should refactor to setters

modules/ml/src/main/java/org/apache/ignite/ml/knn/KNNTrainer.java
```diff
@@ -97,6 +97,17 @@ public Self withIdxType(SpatialIndexType idxType) {
         return self();
     }
 
+    /**
+     * Sets up {@code dataTtl} parameter.
+     *
+     * @param dataTtl Partition data time-to-live in seconds (-1 for an infinite lifetime).
+     * @return This instance.
+     */
+    public Self withDataTtl(long dataTtl) {
```
zaleslaw: Maybe we should tune LE externally and pass it via withLE setter?

examples/src/main/java/org/apache/ignite/examples/ml/knn/KNNClassificationExample.java
dmitrievanthony: By default, it's part of the learning environment. In the case of kNN it's also been added into trainer API because it's critical for this algorithm.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
dmitrievanthony: No, it doesn't look like a `enum`.

modules/ml/src/main/java/org/apache/ignite/ml/environment/DefaultLearningEnvironmentBuilder.java
dmitrievanthony: We introduced `LearningEnvironmentBuilder` to eliminate this problem. This constructor isn't in client API, it's used only in builder.

modules/ml/src/main/java/org/apache/ignite/ml/knn/KNNTrainer.java
dmitrievanthony: Didn't get what you mean.

PR #332, opened 2019-07-29
GG-21944 Logger configuration for WC backend.

vsisko:


modules/web-console/web-console-server/src/main/resources/log4j2.xml
```diff
@@ -0,0 +1,33 @@
+<?xml version="1.0" encoding="UTF-8"?>
+
+<!--
+    Copyright (C) GridGain Systems. All Rights Reserved.
+    _________        _____ __________________        _____
+    __  ____/___________(_)______  /__  ____/______ ____(_)_______
+    _  / __  __  ___/__  / _  __  / _  / __  _  __ `/__  / __  __ \
```
akuznetsov-gridgain: I think this is not correct header.

vsisko: Fixed license.

PR #348, opened 2019-08-01
GG-20511 Improve message parse in browser.

nva:


modules/web-console/frontend/app/modules/agent/decompress.worker.js
```diff
@@ -14,14 +14,12 @@
...
 
-import bigIntJSON from 'json-bigint';
-
 /** This worker parse to JSON from string. */
 // eslint-disable-next-line no-undef
 onmessage = function(e) {
     const data = e.data;
 
-    const res = bigIntJSON({storeAsString: true}).parse(data);
+    const res = JSON.parse(data.replace(/([:,\[])([0-9]{15,}(?:\.\d+)?)/g, '$1"$2"'));
```
akuznetsov-gridgain: Please add unit test for all possible combinations with big numbers, quotes, square bracket and so on.
For example: {a: 12345678901234567890, b: "12345678901234567890", c: [12345678901234567890, "12345678901234567890"]} 

Also it will be interesting to see a benchmarks: JSON.parse vs bigIntJSON vs JSON.parse(replace_reg_exp) to see the difference.

nva: Added tests. 
Benchmark results can be found in ticket https://ggsystems.atlassian.net/browse/GG-20511

PR #351, opened 2019-08-02
GG-22227 Invalid topology version, Added topVer in PartitionCountersN…

ingvard:
…eighborcastRequest.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/mvcc/msg/PartitionCountersNeighborcastRequest.java
```diff
@@ -38,15 +39,27 @@
...
     /** */
     public PartitionCountersNeighborcastRequest() {
     }
 
     /** */
     public PartitionCountersNeighborcastRequest(
-        Collection<PartitionUpdateCountersMessage> updCntrs, IgniteUuid futId) {
+        Collection<PartitionUpdateCountersMessage> updCntrs,
+        IgniteUuid futId,
+        AffinityTopologyVersion topVer
```
ascherbakoff: Better mark arg as @NotNull

ingvard: fixed.

PR #359, opened 2019-08-05
GG-17432 Excessive use of memory in continuous queries

pavlukhin:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
```diff
@@ -913,7 +925,7 @@ private void onEntryUpdate(CacheContinuousQueryEvent evt, boolean notify, boolea
                     notifyLocalListener(evts, trans);
 
                     if (!internal && !skipPrimaryCheck)
-                        sendBackupAcknowledge(ackBuf.onAcknowledged(entry), routineId, ctx);
+                        sendBackupAcknowledge(ackBufBackup.onAcknowledged(entry), routineId, ctx);
```
pavlukhin: Why do not we call `cleanupBuffers` here? Should we cleanup buffer here? If yes it worth adding a test for that case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryEventBuffer.java
```diff
@@ -322,6 +332,21 @@ private Object process0(long cntr, CacheContinuousQueryEntry entry, boolean back
...
         }
 
+        /**
+         * @param updateCntr Acknowledged counter.
+         */
+        synchronized void cleanupEntries(Long updateCntr) {
+            if (entries == null)
+                return;
+
+            for (int i = 0; i < entries.length; i++) {
```
pavlukhin: I suppose we do not need to visit each item in `entries`. Can we remember last cleaned index for a batch and start a cleanup from it? Also we can stop the loop as long as first entry with greater update counter is met.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
ibelyakov: We shouldn't clear the buffer for local deployed listener since we're not using "entryBufs" inside "handleEntry" for this case. I have added additional test for this scenario: "testMultipleNodesWithBackupsWithoutClient".

PR #360, opened 2019-08-05
NBL-68: Fix connection failure detection in `Client.connect()`.

dmelnichuk:


isapego (commented):
Looks OK, but can we add tests for this case? Reproducer seems simple enough.

dmelnichuk:
True, I added a test case.

> Looks OK, but can we add tests for this case? Reproducer seems simple enough.

PR #363, opened 2019-08-06
GG-21824 Implement changing of GMC url by control script

nva:


modules/core/src/main/java/org/apache/ignite/internal/visor/management/ChangeManagementConfigurationTask.java
```diff
@@ -0,0 +1,74 @@
...
+    @Override protected VisorJob<ManagementConfiguration, ManagementConfiguration> job(ManagementConfiguration arg) {
+        return new ChangeManagementConfigurationTask.ChangeManagementConfigurationJob(arg, debug);
+    }
+
+    /**
+     * Job that actually gets the status of rolling upgrade.
+     */
+    private static class ChangeManagementConfigurationJob extends VisorJob<ManagementConfiguration, ManagementConfiguration> {
+        /** */
+        private static final long serialVersionUID = 0L;
```
smoldachev: Why we use same serialVersionUID for all Serialisable classes?

nva: That because we need to disable standard java checks to execute job on different versions nodes. Job class should not have any serialized field except argument/result with portable serialization.

PR #369, opened 2019-08-07
GG-22376 Move agent source to CE repository

nva:


modules/gmc-agent/src/main/java/org/gridgain/agent/AgentConfiguration.java
```diff
@@ -0,0 +1,252 @@
...
+
+import static java.nio.charset.StandardCharsets.UTF_8;
+import static org.gridgain.agent.AgentUtils.secured;
+
+/**
+ * Agent configuration.
+ *
+ * TODO GG-20641 This class should be replaced with manipulation with metastore.
+ */
+public class AgentConfiguration implements Serializable {
```
smoldachev: We don't need this configuration.

PR #380, opened 2019-08-09
GG-21027 Support array types as keys.

ingvard:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -5380,6 +5380,37 @@ public static void writeMap(ObjectOutput out, Map<?, ?> map) throws IOException
...
+            if (obj instanceof float[])
+                return Arrays.hashCode((float[])obj);
+            if (obj instanceof double[])
+                return Arrays.hashCode((double[])obj);
+            if (obj instanceof char[])
+                return Arrays.hashCode((char[])obj);
+            if (obj instanceof boolean[])
+                return Arrays.hashCode((boolean[])obj);
+
+            return Arrays.hashCode((Object[])obj);
```
sanpwc: Seems that this won't work properly if the key is array of arrays.

ingvard: Yes. I suppose it's expected:)

modules/core/src/test/java/org/apache/ignite/cache/store/StoreArrayKeyTest.java
```diff
@@ -0,0 +1,222 @@
...
+        return Arrays.asList(
+            new byte[][] {new byte[] {1, 2, 3}, new byte[] {3, 2, 1}, new byte[] {1, 2, 3}},
+            new short[][] {new short[] {1, 2, 3}, new short[] {3, 2, 1}, new short[] {1, 2, 3}},
+            new int[][] {new int[] {1, 2, 3}, new int[] {3, 2, 1}, new int[] {1, 2, 3}},
+            new long[][] {new long[] {1, 2, 3}, new long[] {3, 2, 1}, new long[] {1, 2, 3}},
+            new float[][] {new float[] {1, 2, 3}, new float[] {3, 2, 1}, new float[] {1, 2, 3}},
+            new double[][] {new double[] {1, 2, 3}, new double[] {3, 2, 1}, new double[] {1, 2, 3}},
+            new char[][] {new char[] {1, 2, 3}, new char[] {3, 2, 1}, new char[] {1, 2, 3}},
+            new boolean[][] {new boolean[] {true, false, true}, new boolean[] {false, true, false}, new boolean[] {true, false, true}},
+            new String[][] {new String[] {"a", "b", "c"}, new String[] {"c", "b", "a"}, new String[] {"a", "b", "c"}}
```
EdShangGG: What's about Object[]?

modules/core/src/test/java/org/apache/ignite/cache/store/StoreArrayKeyTest.java
```diff
@@ -0,0 +1,222 @@
...
+    public void shouldReadFromBackup() {
+        IgniteCache<Object, Object> cache = node2.getOrCreateCache(CACHE_WITH_BACKUPS);
+
+        cache.put(firstKey, 1);
+        cache.put(secondKey, 2);
+        cache.put(likeFirstKey, 3);
+
+        assertEquals(2, cache.size());
+
+        stopGrid(0);
```
EdShangGG: Why do we think that the 0th node was primary for these partitions?

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteCacheTestSuite.java
```diff
@@ -215,6 +216,8 @@
         GridTestUtils.addTestIfNeeded(suite, IgniteCachePrivateExecutionContextTest.class, ignoredTests);
         GridTestUtils.addTestIfNeeded(suite, IgniteCacheSharedExecutionContextTest.class, ignoredTests);
 
+        GridTestUtils.addTestIfNeeded(suite, StoreArrayKeyTest.class, ignoredTests);
```
EdShangGG: Why was addTestIfNeeded used? In which cases should it be filtered out?

modules/yardstick/src/main/java/org/apache/ignite/yardstick/cache/IgniteGetAllPutAllTxBytesKeyBenchmark.java
```diff
@@ -0,0 +1,74 @@
...
+                int base = r.nextRandom();
+
+                byte[] key = String.valueOf(base).getBytes();
+
+                vals.put(key, base);
+            }
+
+            IgniteCache<byte[], Integer> cache = cacheForOperation();
+
+            cache.getAll(vals.keySet());
```
EdShangGG: There is no test which covers doAll operation for array keys.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
ingvard: I fixed it. Added deep hash code.

modules/core/src/test/java/org/apache/ignite/cache/store/StoreArrayKeyTest.java
ingvard: Added.

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteCacheTestSuite.java
ingvard: I used it because other tests in this class use it.

modules/core/src/test/java/org/apache/ignite/cache/store/StoreArrayKeyTest.java
ingvard: It's a mistake. I don't have any idea how to fix it. (It was removed).

PR #385, opened 2019-08-13
GG-22625 Add "help" menu

Klaster1:


modules/web-console/e2e/testcafe/fixtures/help-menu.js
```diff
@@ -0,0 +1,50 @@
...
+        await t.useRole(user);
+    })
+    .afterEach(async(t) => {
+        await dropTestDB();
+    });
+
+test.page(resolveUrl('/'))('Help menu items', async(t) => {
+    await t
+        .hover(helpMenu.trigger)
+        .expect(helpMenu.item('Documentation').getAttribute('href')).eql('https://docs.gridgain.com/docs/web-console')
```
nva: Possibly we should also check that target attribute equals "_blank"

Klaster1: Good point, I'll add this.

PR #393, opened 2019-08-14
GG-22688: Collaborative filtering (recommendation system) using SGD

dmitrievanthony:


modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationModel.java
```diff
@@ -0,0 +1,59 @@
...
+ * @param <O> Type of an object of recommendation.
+ * @param <S> Type of a subject of recommendation.
+ */
+public class RecommendationModel<O extends Serializable, S extends Serializable>
+    implements IgniteModel<ObjectSubjectPair<O, S>, Double> {
+    /** */
+    private static final long serialVersionUID = -3664382168079054785L;
+
+    /** Object of recommendation matrix (result of factorization of rating matrix). */
+    private final Map<O, Vector> objMatrix;
```
avplatonov: Please be careful with size of this matrices
I suppose that this model can be very big in terms of memory using. Lets create additional ticket for another implementation where model is distributed on cluster.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationTrainer.java
```diff
@@ -0,0 +1,319 @@
...
+    private LearningEnvironmentBuilder environmentBuilder = LearningEnvironmentBuilder.defaultBuilder();
+
+    /** Trainer learning environment. */
+    private LearningEnvironment trainerEnvironment = environmentBuilder.buildForTrainer();
+
+    /** Batch size of stochastic gradient descent. The size of a dataset used on each step of SGD. */
+    private int batchSize = 1000;
+
+    /** Regularizer. */
+    private double regularizer = 0.0;
```
avplatonov: regularizer is function that try to setup model with some restrictions. This is the strength parameter of regularizer

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationTrainer.java
```diff
@@ -0,0 +1,319 @@
...
+            // Collect total set of objects and subjects (their identifiers).
+            Set<O> objects = dataset.compute(RecommendationDatasetData::getObjects, RecommendationTrainer::join);
+            Set<S> subjects = dataset.compute(RecommendationDatasetData::getSubjects, RecommendationTrainer::join);
+
+            // Generate initial model (object and subject matrices) initializing them with random values.
+            Map<O, Vector> objMatrix = generateRandomVectorForEach(objects, trainerEnvironment.randomNumbersGenerator());
+            Map<S, Vector> subjMatrix = generateRandomVectorForEach(subjects, trainerEnvironment.randomNumbersGenerator());
+
+            // SGD steps.
+            for (int i = 0; i < maxIterations; i++) {
```
avplatonov: Why do you skip the gradient descent convergence calculation? If a gradient is converged we can stop SGD iterations.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/util/MatrixFactorizationGradient.java
```diff
@@ -0,0 +1,68 @@
...
+import java.util.Map;
+import org.apache.ignite.ml.math.primitives.vector.Vector;
+
+/**
+ * Gradient of matrix factorization loss function.
+ *
+ * @param <O> Type of an object of recommendation.
+ * @param <S> Type of a subject of recommendation.
+ */
+public class MatrixFactorizationGradient<O extends Serializable, S extends Serializable> implements Serializable {
```
avplatonov: I'm not sure but maybe you move gradient application to vector to MatrixFactorizationGradient class? IMHO it looks more clearly and size of RecommendationTrainer class will be decreased

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/util/RecommendationDatasetData.java
```diff
@@ -0,0 +1,153 @@
...
+     * Constructs a new instance of recommendation dataset data.
+     *
+     * @param ratings All ratings related to the partition.
+     */
+    public RecommendationDatasetData(List<? extends ObjectSubjectRatingTriplet<O, S>> ratings) {
+        this.ratings = Collections.unmodifiableList(ratings);
+    }
+
+    /**
+     * Calculates gradient of the loss function of recommendation system SGD training.
```
avplatonov: I think it will be more convenient if you write gradient formulas here in javadoc explicitly. They are unreadable in code.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationModel.java
dmitrievanthony: Correspondent task created: https://ggsystems.atlassian.net/browse/GG-22331.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationTrainer.java
dmitrievanthony: Correspondent task created: https://ggsystems.atlassian.net/browse/GG-22916.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/util/MatrixFactorizationGradient.java
dmitrievanthony: Done.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/util/RecommendationDatasetData.java
dmitrievanthony: I haven't found a convenient way to add a formula into Javadoc, but I can add a link to an article about it. Do you think it makes sense?

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/RecommendationTrainer.java
dmitrievanthony: Done.

modules/ml/src/main/java/org/apache/ignite/ml/recommendation/util/RecommendationDatasetData.java
avplatonov: Yes, it will be great!

PR #409, opened 2019-08-19
GG-22618 Collect DR events.

nva:


modules/core/src/main/java/org/apache/ignite/internal/visor/util/VisorTaskUtils.java
```diff
@@ -421,14 +421,14 @@ public static boolean boolValue(String propName, boolean dflt) {
      * @param ignite Grid instance.
      * @return {@code true} if all task events explicitly specified in configuration.
      */
-    public static boolean checkExplicitTaskMonitoring(Ignite ignite) {
-        int[] evts = ignite.configuration().getIncludeEventTypes();
+    public static boolean checkExplicitEvents(Ignite ignite, int[] evts) {
```
akuznetsov-gridgain: param `evts` is missing in javadocs.

modules/core/src/main/java/org/apache/ignite/internal/visor/util/VisorTaskUtils.java
```diff
@@ -421,14 +421,14 @@ public static boolean boolValue(String propName, boolean dflt) {
...
+        int[] inclEvtTypes = ignite.configuration().getIncludeEventTypes();
```
akuznetsov-gridgain: Lets rename inclEvtTypes  -> curEvts ?

PR #421, opened 2019-08-22
GG-22986 Set correct default value for Web Console logs destination if IGNITE_HOME property is not set.

nva:


modules/web-console/bin/web-console.bat
```diff
@@ -82,6 +82,18 @@ goto checkIgniteHome2
...
+::
+:: ADD YOUR/CHANGE ADDITIONAL OPTIONS HERE
+::
+if "%JVM_OPTS%" == "" set JVM_OPTS=-Xms1g -Xmx1g -server -XX:MaxMetaspaceSize=256m
+
+::
+:: Disable update notifier.
+::
+set JVM_OPTS=-DIGNITE_UPDATE_NOTIFIER %JVM_OPTS%
+
```
vsisko: Should -DIGNITE_UPDATE_NOTIFIER be equal false as in sh file?

nva: Nice catch, fixed

PR #423, opened 2019-08-22
GG-23047: Add limited queue for traces.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
```diff
@@ -93,12 +113,27 @@ public GmcSpanExporter(GridKernalContext ctx) {
...
+
+    /**
+     * @return Worker which send messages from queue to topic.
+     */
+    private RetryableSender<SpanBatch> createSenderWorker() {
+        return new RetryableSender<SpanBatch>(log, QUEUE_CAP) {
+            @Override protected void send(SpanBatch list) {
                 IgniteMessaging msg = ctx.grid().message(ctx.grid().cluster().forOldest());
-                msg.send(TRACING_TOPIC, new SpanList(spans));
+                msg.send(TRACING_TOPIC, list);
```
akuznetsov-gridgain: Can we inline `msg` ?

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
```diff
@@ -0,0 +1,89 @@
...
+package org.gridgain.service.tracing;
+
+import org.apache.ignite.IgniteLogger;
+
+import java.util.concurrent.LinkedBlockingDeque;
+
+/**
+ * Retryable sender with limited queue.
+ */
+public abstract class RetryableSender<T> implements Runnable {
```
akuznetsov-gridgain: May be instead of `abstract  class` we can implement class that accept Runnable / Closure as argument?
Usage will be like: `new RetryableSender<SpanBatch>(log, QUEUE_CAP, () -> {some_code});`

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/TracingService.java
```diff
@@ -54,39 +61,39 @@
...
-    private void send(SpanList spans) {
-        buf.addAll(spans.getList());
-
-        if (mgr.send(buildSaveSpanDest(ctx.cluster().get().id()), Lists.newArrayList(buf)))
-            buf.clear();
+    private RetryableSender<SpanBatch> createSenderWorker() {
+        return new RetryableSender<SpanBatch>(log, QUEUE_CAP) {
+            @Override protected void send(SpanBatch list) {
+                if (!mgr.send(buildSaveSpanDest(ctx.cluster().get().id()), list.getList()))
+                    throw new IgniteException("Message with spans doesn't send");
```
akuznetsov-gridgain: "Message with spans doesn't send" -> "Failed to send message with spans"

smoldachev: Fix.

modules/gmc-agent/src/main/java/org/gridgain/dto/span/SpanBatch.java
```diff
@@ -39,13 +39,13 @@
     /**
      * Default constructor.
      */
-    public SpanList() {
+    public SpanBatch() {
```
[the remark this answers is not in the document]
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/MetricsService.java
```diff
@@ -46,6 +46,7 @@
     /** Logger. */
     private IgniteLogger log;
 
+    /** Listener. */
```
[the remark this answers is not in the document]
smoldachev: Done.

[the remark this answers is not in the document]
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
```diff
@@ -0,0 +1,89 @@
...
+                send(e);
+
+                retryCnt = 0;
+            }
+            catch (InterruptedException ex) {
+                break;
+            }
+            catch (Exception ex) {
+                if (retryCnt == 0)
+                    log.warning("Can't send message with spans, retry!", ex);
```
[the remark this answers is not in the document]
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/TracingService.java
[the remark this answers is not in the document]
smoldachev: Fix.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/TracingService.java
```diff
@@ -34,6 +33,12 @@
  * Tracing service.
  */
 public class TracingService implements AutoCloseable {
+    /** Queue capacity. */
+    private static final int QUEUE_CAP = 100;
+
+    /** Logger. */
+    private IgniteLogger log;
```
akuznetsov-gridgain: Can `log`, `ctx` and `onNodeTraces` be final?

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
```diff
@@ -93,16 +113,28 @@ public GmcSpanExporter(GridKernalContext ctx) {
...
+                        .forEach(worker::addToSendQueue);
             }
         };
     }
 
+    /**
+     * @return Worker which send messages from queue to topic.
+     */
+    private RetryableSender<SpanBatch> createSenderWorker() {
+        Consumer<SpanBatch> sndFn = (b) -> ctx.grid().message(ctx.grid().cluster().forOldest()).send(TRACING_TOPIC, b);
```
akuznetsov-gridgain: Can we inline `sndFn` ?

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
```diff
@@ -0,0 +1,89 @@
...
+
+    /** {@inheritDoc} */
+    @Override public void run() {
+        while (true) {
+            T e = null;
+
+            try {
+                Thread.sleep(retryCnt * 1000);
+
+                e = queue.takeFirst();
```
nva: Why not take()?

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
```diff
@@ -0,0 +1,89 @@
...
+                retryCnt = 0;
+            }
+            catch (InterruptedException ex) {
+                break;
+            }
+            catch (Exception ex) {
+                if (retryCnt <= MAX_SLEEP_TIME_SECONDS)
+                    retryCnt++;
+
+                if (retryCnt == 0)
```
nva: retryCnt will be always >= 1

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
```diff
@@ -0,0 +1,89 @@
...
+                e = queue.takeFirst();
+                sndFn.accept(e);
+
+                retryCnt = 0;
+            }
+            catch (InterruptedException ex) {
+                break;
+            }
+            catch (Exception ex) {
+                if (retryCnt <= MAX_SLEEP_TIME_SECONDS)
```
nva: You can use MAth.min(MAX_SLEEP_TIME_SECONDS, retryCnt) in sleep

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
```diff
@@ -63,6 +72,12 @@
     /** Exporter. */
     private OpenCensusTraceExporter exporter;
 
+    /** Executor service. */
+    private final ExecutorService exSrvc = Executors.newSingleThreadExecutor();
```
nva: Why executor not incapsulated in sender?

modules/gmc-agent/src/main/java/org/gridgain/dto/span/SpanBatch.java
```diff
@@ -27,9 +27,9 @@
 import java.util.List;
 
 /**
- * DTO for span list.
+ * DTO for batch of span list.
  */
-public class SpanList implements Externalizable {
+public class SpanBatch implements Externalizable {
```
nva: I don't see any reason of using this wrapper over List. Let's remove it

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
```diff
@@ -93,16 +113,28 @@ public GmcSpanExporter(GridKernalContext ctx) {
...
             @Override public void timeLimitedExport(Collection<SpanData> spanDataList) {
-                List<Span> spans = spanDataList.stream()
+                List<Span> spans = spanDataList
+                        .stream()
                         .map(GmcSpanExporter::fromSpanDataToSpan)
                         .collect(Collectors.toList());
 
-                IgniteMessaging msg = ctx.grid().message(ctx.grid().cluster().forOldest());
-                msg.send(TRACING_TOPIC, new SpanList(spans));
+                Lists.partition(spans, BATCH_SIZE)
```
nva: Why sender doesn't split data into batch?

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
```diff
@@ -0,0 +1,89 @@
...
+
+                addToSendQueue(e);
+            }
+        }
+    }
+
+    /**
+     * @param e Span list.
+     */
+    public synchronized void addToSendQueue(T e) {
```
nva: I think `send` will be better. And why we need synchronized over concurrent collection

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/TracingService.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
smoldachev: Change on take.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
smoldachev: Fix

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
smoldachev: Thanks :)

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
smoldachev: Agree, move to sender.

modules/gmc-agent/src/main/java/org/gridgain/dto/span/SpanBatch.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/GmcSpanExporter.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/service/tracing/RetryableSender.java
smoldachev: Rename method. 

We need synchronization, because we must provide guarantees that we will not lose an element in the second offerLast call (for example, the second thread can add an element to the queue, while the first thread calls removeFirst and falls to sleep before calling offerLast, and after waking up it cannot add an element because the queue is full).

modules/gmc-agent/src/main/java/org/gridgain/agent/RetryableSender.java
```diff
@@ -0,0 +1,133 @@
...
+            catch (InterruptedException ex) {
+                break;
+            }
+            catch (Exception ex) {
+                retryCnt++;
+
+                if (retryCnt == 1)
+                    log.warning("Failed to send message with spans, will retry in " + retryCnt * 1000 + " ms", ex);
+
+                addToQueue(e);
```
nva: Why we return value into beginning of the queue? Use peek + remove on success instead of take + pushLast on fail

modules/gmc-agent/src/main/java/org/gridgain/agent/RetryableSender.java
```diff
@@ -0,0 +1,133 @@
...
+            res.add(batch);
+
+        return res;
+    }
+
+    /**
+     * @param batch Batch.
+     */
+    private synchronized void addToQueue(List<T> batch) {
+        if (!queue.offerLast(batch)) {
```
nva: Possibly we should create queue with capacity = target capacity + 1. Check after offer if target capacity exceeded then remove first

modules/gmc-agent/src/main/java/org/gridgain/agent/RetryableSender.java
```diff
@@ -0,0 +1,133 @@
...
+        exSrvc.submit(this);
+    }
+
+    /** {@inheritDoc} */
+    @Override public void run() {
+        while (true) {
+            List<T> e = null;
+
+            try {
+                Thread.sleep(Math.min(MAX_SLEEP_TIME_SECONDS, retryCnt) * 1000);
```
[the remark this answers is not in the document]
smoldachev: How we affect messaging here? We run sender in another thread and try to sleep only when we get exception.

modules/gmc-agent/src/main/java/org/gridgain/agent/RetryableSender.java
smoldachev: Because peek is not a иlocking method, and we will waste the cpu when the queue is empty. We can try using offerFirst if the sending failed, and if we have enough capacity in the queue, we put the failed packet on top and try again.

modules/gmc-agent/src/main/java/org/gridgain/agent/RetryableSender.java
```diff
@@ -0,0 +1,133 @@
...
+
+                e = queue.take();
+                sndFn.accept(e);
```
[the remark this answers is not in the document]
smoldachev: Why? We are using the generic logic with retry. What happened if we'll wait additional 1 second for coordinator case?

modules/gmc-agent/src/main/java/org/gridgain/agent/RetryableSender.java
smoldachev: If we will check size then we will block all queue for this, i think it's not good idea.

PR #425, opened 2019-08-22
GG-21615 .NET: StreamTransformer is not invoked for duplicate entries

ptupitsyn:
* Root cause: `StreamTransformer` used `InvokeAll`, while in Java it is `Invoke` in a loop, causing different semantics.
* Related issue: local processor allocation is not used, causing performance degradation for `Invoke` calls on local cache entries (which is exactly what's going on with `StreamTransformer`)


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/PlatformCache.java
```diff
@@ -744,16 +748,16 @@ public IgniteCache rawCache() {
 
                 case OP_INVOKE: {
                     Object key = reader.readObjectDetached();
-
-                    CacheEntryProcessor proc = platformCtx.createCacheEntryProcessor(reader.readObjectDetached(), 0);
+                    long ptr = reader.readLong();
+                    CacheEntryProcessor proc = platformCtx.createCacheEntryProcessor(reader.readObjectDetached(), ptr);
```
ashapkin: Extract a private method to avoid code duplicates?

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
```diff
@@ -981,23 +992,39 @@ private Task<long> SizeAsync0(int? part, params CachePeekMode[] modes)
...
+
+            try
+            {
+                return DoOutOpAsync(CacheOp.InvokeAsync, writer =>
+                    {
+                        writer.WriteObjectDetached(key);
+                        writer.WriteLong(ptr);
+                        writer.WriteObjectDetached(holder);
+                    },
+                    r =>
```
ashapkin: Lets be consistent with the first arg, rename to "reader" (or "writer" to "w")
Here and below.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
```diff
@@ -981,23 +992,39 @@ private Task<long> SizeAsync0(int? part, params CachePeekMode[] modes)
...
-            return DoOutOpAsync(CacheOp.InvokeAsync, writer =>
-                {
-                    writer.WriteObjectDetached(key);
-                    writer.WriteObjectDetached(holder);
-                },
-                r =>
-                {
-                    if (r == null)
-                        return default(TRes);
+            var ptr = AllocateNoTx(holder);
```
ashapkin: Can you describe this a little bit, please. It's kinda confusing to read  `AllocateNoTx` right after `StartTx`. Would not `AllocateNoTx` always return 0? Cause at first glance the original issue was caused by PlatformsUtils.readSet method.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Dataload/DataStreamerTest.cs
```diff
@@ -502,6 +502,27 @@ public void TestStreamTransformer()
             TestStreamReceiver(new StreamTransformer<int, int, int, int>(new EntryProcessorBinarizable()));
         }
 
+        [Test]
```
ashapkin: Dont we need a separate test for local DataStreamer (with pointer allocation)?

modules/platforms/cpp/core/include/ignite/impl/operations.h
```diff
@@ -252,6 +252,40 @@ namespace ignite
...
+             */
+            InCacheInvokeOperation(const T1& val1, const T2& val2) : val1(val1), val2(val2)
+            {
+                // No-op.
+            }
+
+            virtual void ProcessInput(ignite::impl::binary::BinaryWriterImpl& writer)
+            {
+                writer.WriteTopObject<T1>(val1);
+                writer.GetStream()->WriteInt64(0);
```
isapego: You can just type `writer.WriteInt64(0);`, no need to get stream here.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/PlatformCache.java
ptupitsyn: I don't think it will improve readability, and it won't save LoC, so let's keep this as is.

ptupitsyn: I don't think it will improve readability, and it won't save LoC, so let's keep this as is.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
ptupitsyn: Agree, fixed.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Dataload/DataStreamerTest.cs
ptupitsyn: This is covered by existing tests.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
ptupitsyn: I've updated the PR description (so this gets into the commit message too).

`StartTx` starts a transaction only when there is a `TransactionScope` and the cache is `Transactional`, in all other cases it just does nothing. `AllocateNoTx` checks if there is an active Ignite transaction, which can be created in two ways - with a `TransactionScope` or with `ITransactions` API.

I agree that naming is confusing here, let me try to improve it.

ptupitsyn: Changed names to `StartTxIfNeeded` and `AllocateIfNoTx`, hopefully this is less confusing.

modules/platforms/cpp/core/include/ignite/impl/operations.h
ptupitsyn: Good point, fixed.

PR #436, opened 2019-08-25
Prototype of the distributed SQL with Apache Calcite

rkondakov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/scanner/WalScanner.java
```diff
@@ -79,7 +79,7 @@ private static WALIterator standaloneWalIterator(
     ) throws IgniteCheckedException {
         return iteratorFactory.iterator(
             parametersBuilder.copy().addFilter((type, pointer) ->
-                // PHYSICAL need fo page shanpshot or delta record.
+                // INSTANCE need fo page shanpshot or delta record.
```
AMashenkov: Looks like inadvertent renaming.

rkondakov: Oops!

PR #439, opened 2019-08-26
GG-19739 UUID in the binary type name is replaced by hash of columns.

rkondakov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -1107,6 +1112,45 @@ public static boolean isCommandNoOp(Prepared cmd) {
...
+
+            String fieldsDigest = String.join(", ", fieldDigests);
+
+            MessageDigest md = MessageDigest.getInstance("MD5");
+
+            md.update(fieldsDigest.getBytes());
+
+            StringBuilder sb = new StringBuilder();
+
+            for (byte md5Byte : md.digest())
```
AMashenkov: Do you mean Integer.toHexString(md5Byte) here?
You can use Hex.encodeHex(byte[]) from apache-commons which look less tricky and generate much less garbage.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -1355,8 +1355,8 @@ public static String createTableCacheName(String schemaName, String tblName) {
      * @param tblName Table name.
      * @return Value type name.
      */
-    public static String createTableValueTypeName(String schemaName, String tblName) {
-        return createTableCacheName(schemaName, tblName) + "_" + UUID.randomUUID().toString().replace("-", "_");
+    public static String createTableValueTypeName(String schemaName, String tblName, String tblDigest) {
```
korlov42: please add new param to javadoc.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -1107,6 +1112,45 @@ public static boolean isCommandNoOp(Prepared cmd) {
         return cmd instanceof NoOperation;
     }
 
+    /**
+     * Creates table digest as MD5 hash from sorted list of column names and types.
+     *
+     * @param tbl Create table command
```
korlov42: dot at the end is missing

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
korlov42: also _IgniteUtils_ already has `public static String byteArray2HexString(byte[] arr)`. It may suits.

rkondakov: Unfortunately, we do not have dependency on apache-commons library in Ignite's core module and therefore I can not use `Hex.encodeHex(byte[])` here. Moreover, this is not a hot path, so garbage generation impact can be neglected here.

AMashenkov: There is commons-codec dependency in ignite-indexing.
CommandProcessor is a part of indexing module, not core.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
rkondakov: Done.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
rkondakov: Done.

PR #442, opened 2019-08-26
GG-21926: Export node configuration to gmc.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/service/config/NodeConfigurationService.java
```diff
@@ -0,0 +1,81 @@
...
+    /** Manager. */
+    private final WebSocketManager mgr;
+
+    /** Sender. */
+    private final GmcSender<VisorGridConfiguration> snd;
+
+    /** On node traces listener. */
+    private final IgniteBiPredicate<UUID, Object> lsnr = this::onNodeConfiguration;
+
+    public NodeConfigurationService(GridKernalContext ctx, WebSocketManager mgr) {
```
akuznetsov-gridgain: Missing javadocs.

modules/gmc-agent/src/main/java/org/gridgain/service/config/NodeConfigurationService.java
```diff
@@ -0,0 +1,81 @@
...
+    @Override public void close() {
+        ctx.grid().message().stopLocalListen(NODE_CONFIGURATION_TOPIC, lsnr);
+        U.closeQuiet(snd);
+    }
+
+    /**
+     * @param uuid Uuid.
+     * @param cfgList Config list.
+     */
+    boolean onNodeConfiguration(UUID uuid, Object cfgList) {
```
akuznetsov-gridgain: maybe 'boolean' -> 'Void'  + return null ?

modules/gmc-agent/src/main/java/org/gridgain/service/sender/RetryableSender.java
```diff
@@ -22,6 +22,7 @@
 import java.util.concurrent.BlockingQueue;
 import java.util.concurrent.ExecutorService;
 import java.util.concurrent.Executors;
+import com.google.common.collect.Lists;
```
akuznetsov-gridgain: Do we really need guava ?

modules/gmc-agent/src/main/java/org/gridgain/service/config/NodeConfigurationExporter.java
```diff
@@ -0,0 +1,64 @@
...
+
+    /** Status description. */
+    public static final String NODE_CONFIGURATION_TOPIC = "gmc-node-configuration-topic";
+
+    /** Sender. */
+    private final CoordinatorSender<VisorGridConfiguration> snd;
+
+    /** Context. */
+    private GridKernalContext ctx;
+
```
nva: not needed line

modules/gmc-agent/src/main/java/org/gridgain/service/sender/RetryableSender.java
```diff
@@ -73,6 +74,14 @@ protected RetryableSender(int cap) {
         exSrvc.shutdown();
     }
 
+    /**
+     * @param element Element to send.
+     */
+    public void send(T element) {
+        if (element != null)
+            addToQueue(Lists.newArrayList(element));
```
ingvard: Collections.singleton()

modules/gmc-agent/src/main/java/org/gridgain/service/config/NodeConfigurationService.java
smoldachev: Fix.

modules/gmc-agent/src/main/java/org/gridgain/service/config/NodeConfigurationService.java
smoldachev: The "onNodeConfiguration" is IgniteBiPredicate with "boolean apply" method.

modules/gmc-agent/src/main/java/org/gridgain/service/sender/RetryableSender.java
smoldachev: Nope, change on Collections.singletonList.

modules/gmc-agent/src/main/java/org/gridgain/service/config/NodeConfigurationExporter.java
smoldachev: Fix.

modules/gmc-agent/src/main/java/org/gridgain/service/sender/RetryableSender.java
smoldachev: Thanks :)

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -103,12 +111,14 @@ public Agent(GridKernalContext ctx) {
...
     @Override public void onKernalStart(boolean active) {
         spanExporter = new GmcSpanExporter(ctx);
+        nodeConfigurationExporter = new NodeConfigurationExporter(ctx);
         metaStorage = ctx.cache().context().database().metaStorage();
 
         launchAgentListener(null, ctx.discovery().discoCache());
 
         // Listener for coordinator changed.
         ctx.event().addDiscoveryEventListener(this::launchAgentListener, EVTS_DISCOVERY);
+        nodeConfigurationExporter.export();
```
akuznetsov-gridgain: I think it is not good idea to do potentially "heavy" operations in onKernalStart method. 
May be backend should "poll" for configurations from cluster?

smoldachev: I don't think what putting element in queue is "heavy" operation (yes, we are wrapping configuration but in the future we will have a serializable IgniteConfiguration). Maybe we can suggest an additional type of callbacks (afterStart for example) to core team?

PR #443, opened 2019-08-26
GG-17341 review

ascherbakoff:


modules/clients/src/test/java/org/apache/ignite/internal/jdbc2/JdbcStreamingSelfTest.java
```diff
@@ -160,6 +163,59 @@ protected Connection createStreamedConnection(boolean allowOverwrite, long flush
         super.afterTest();
     }
 
+    /**
```
ascherbakoff: Do we need a test for INSERT INTO SELECT FROM scenario ? I would prefer to add.

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
```diff
@@ -453,6 +453,21 @@
      */
     public void active(boolean active);
 
+    /**
+     * Checks Ignite grid in read-only mode or not.
+     *
+     * @return {@code true} if grid is in read-only mode and {@code false} If grid allows data modification operations.
+     */
+    public boolean readOnly();
```
ascherbakoff: Do we need throwing an exception here ? What happens if the method is called when cluster is not activated ?

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
```diff
@@ -453,6 +453,21 @@
      */
     public void active(boolean active);
 
+    /**
+     * Checks Ignite grid in read-only mode or not.
```
ascherbakoff: Javadoc is not very clear. Consider consulting someone english speaking to improve doc quality in public interfaces.

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
```diff
@@ -453,6 +453,21 @@
...
+     *
+     * @return {@code true} if grid is in read-only mode and {@code false} If grid allows data modification operations.
```
ascherbakoff: Here and below check for letters case (must be @code True) and same issues as above - quality of documentation.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/ApiParity/ClusterParityTest.cs
```diff
@@ -37,7 +37,8 @@ public class ClusterParityTest
             "baselineAutoAdjustEnabled",
             "baselineAutoAdjustTimeout",
             "baselineAutoAdjustStatus",
-            "clientReconnectFuture"
+            "clientReconnectFuture",
+            "readOnly"
```
ascherbakoff: Why this property marked as unneeded ?

modules/clients/src/test/java/org/apache/ignite/internal/jdbc2/JdbcStreamingSelfTest.java
antonovsergey93: `java.sql.SQLException: Streaming mode supports only INSERT commands without subqueries.`

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
antonovsergey93:
```

modules/core/src/main/java/org/apache/ignite/IgniteCluster.java
antonovsergey93: done

modules/platforms/dotnet/Apache.Ignite.Core.Tests/ApiParity/ClusterParityTest.cs
antonovsergey93: I moved to MissingMembers and create ticket https://ggsystems.atlassian.net/browse/GG-23363 for implementation.

PR #449, opened 2019-08-27
GG-23175 Web Console should check IgniteFeatures instead of cluster version.

nva:


modules/web-console/frontend/app/types/Agent.ts
```diff
@@ -69,3 +72,15 @@ export type VisorQueryResult =
 export type QuerySqlX2Response = VisorQueryResult
 
 export type QueryFetchFirstPageResult = VisorQueryResult
+
+export enum IgniteFeatures {
+	INDEXING= 15,
+	WC_ROLLING_UPGRADE_STATUS = 20,
+	WC_SNAPSHOT_CHAIN_MODE = 21
```
akuznetsov-gridgain: Where WC_DR_STATE_EVENTS ?

modules/web-console/web-console-common/src/main/java/org/apache/ignite/console/utils/Utils.java
```diff
@@ -197,4 +198,26 @@ public static String extractErrorMessage(String prefix, Throwable e) {
...
     }
+
+    /**
+     * Checks that feature supported by cluster.
+     *
+     * @param featuresAttrBytes Byte array value of supported features.
+     * @param featureId Feature to check.
+     * @return {@code True} if feature is declared to be supported by cluster.
+     */
+    public static boolean gridSupports(byte[] featuresAttrBytes, int featureId) {
```
akuznetsov-gridgain: May be `featureSupported` will be a better name than `gridSupports`?

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -82,7 +82,23 @@
     /**
      * A mode when data nodes throttle update rate regarding to DR sender load
      */
-    DR_DATA_NODE_SMART_THROTTLING(19);
+    DR_DATA_NODE_SMART_THROTTLING(19),
```
akuznetsov-gridgain: FEATURES NOT RESERVED AT:
https://ggsystems.atlassian.net/wiki/spaces/GG/pages/1192198276/Community+edition+features+list

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -176,8 +177,16 @@ class ConnectionState {
 export default class AgentManager {
     static $inject = ['Demo', '$q', '$transitions', '$location', 'AgentModal', 'UserNotifications', 'IgniteVersion', 'ClusterLoginService'];
 
+    currentCluster$: Observable<ConnectionState>;
+    clusterIsActive$: Observable<boolean>;
+    clusterIsAvailable$: Observable<boolean>;
+
     clusterVersion: string;
 
+    indexing: boolean;
```
akuznetsov-gridgain: Do we need  `indexing`, `ruStaus` and `snapshoChainMode`? I think we should have only `supportedFeatures` and enum or bit numbers to later check with method like `isFeatureSupported(features, FEATURE)`

akuznetsov-gridgain (changes requested):

modules/web-console/frontend/app/components/page-queries/components/queries-notebook/controller.ts
```diff
@@ -2208,13 +2208,8 @@ export class NotebookCtrl {
...
             if (this._hasRunningQueries(paragraphs)) {
-                try {
-                    return Confirm.confirm('You have running queries. Are you sure you want to cancel them?')
-                        .then(() => this._closeOpenedQueries(paragraphs));
-                }
-                catch (err) {
-                    return Promise.reject(new CancellationError());
-                }
+                return Confirm.confirm('You have running queries. Are you sure you want to cancel them?')
+                    .then(() => this._closeOpenedQueries(paragraphs));
```
Klaster1: I don't see how this is related to the task. Commit messages tell nothing useful about this change either.

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -118,8 +118,13 @@ const getLastActiveClusterId = () => {
 };
 
 class ConnectionState {
-    constructor(cluster) {
-        this.cluster = cluster;
+    cluster: AgentTypes.ClusterStats;
```
Klaster1:
```suggestion
    cluster: AgentTypes.ClusterStats | null;
```

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -165,9 +169,6 @@ class ConnectionState {
     }
 
     disconnect() {
-        if (this.cluster)
```
Klaster1: Why did you remove this?

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -176,8 +177,16 @@ class ConnectionState {
 export default class AgentManager {
     static $inject = ['Demo', '$q', '$transitions', '$location', 'AgentModal', 'UserNotifications', 'IgniteVersion', 'ClusterLoginService'];
 
+    currentCluster$: Observable<ConnectionState>;
```
Klaster1: 👍

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
Klaster1: I agree with @akuznetsov-gridgain, there's too much going on AgentManager already, we can at least reduce the amount of properties it exposes.

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -871,4 +890,42 @@ export default class AgentManager {
...
+    }
+
+    /**
+     * Decode base64 string to byte array.
+     *
+     * @param base64 Base64 string.
+     * @return {Uint8Array} Result byte array.
+     * @private
+     */
+    _base64ToArrayBuffer(base64: string) {
```
Klaster1:
```suggestion
    _base64ToArrayBuffer(base64: string): Uint8Array {
```

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -871,4 +890,42 @@ export default class AgentManager {
...
+        const bitIdx = feature & 0x7;
+
+        return (bytes[byteIdx] & (1 << bitIdx)) !== 0;
+    }
+
+    /**
+     * Decode base64 string to byte array.
+     *
+     * @param base64 Base64 string.
+     * @return {Uint8Array} Result byte array.
```
Klaster1:
```suggestion
     * @return Result byte array.
```

Klaster1 (changes requested):

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
nva: not found usage of this.cluster.disconnect

modules/web-console/frontend/app/components/page-queries/components/queries-notebook/controller.ts
nva: This code don't throw exceptions, just return promise.

modules/web-console/web-console-common/src/main/java/org/apache/ignite/console/utils/Utils.java
nva: Remove

vsisko (changes requested):
A am agree with Aleksey and Ilya. Also please check that e2e tests are not broken.

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -82,7 +82,23 @@
...
+
+    /**
+     * Support rolling upgrade status task for Web Console.
+     */
+    WC_ROLLING_UPGRADE_STATUS(20),
+
+    /**
+     * Support for chain parameter in snapshot delete task for Web Console.
+     */
+    WC_SNAPSHOT_CHAIN_MODE(21),
```
akuznetsov-gridgain: One more comment - you must add unit test for new features

akuznetsov-gridgain (changes requested):

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -891,6 +885,21 @@ export default class AgentManager {
...
     }
 
+    /**
+     * Collect features supported by the cluster.
+     *
+     * @param cluster Cluster.
+     * @return all supported features.
+     */
+    allFeatures(cluster: AgentTypes.ClusterStats): AgentTypes.IgniteFeatures[] {
+        return _.reduce(AgentTypes.IgniteFeatures, (acc, featureId) => {
```
Klaster1: The new API looks awkward, but it's better than before. We should probably iterate on this later.

Klaster1 (approved):

Klaster1 (approved):

modules/web-console/frontend/app/types/Agent.ts
nva: Will do this in separated ticket

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
nva: We already have e2e test

modules/web-console/frontend/app/types/Agent.ts
Klaster1: @nva please make a task for this so we don't forget.

modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
nva: I tried to investigate this and found that I could not define an object with keys as in enumeration keys, only as values.

```
enum A {
  A1 = 10
}

type B = {
  [x in A]: boolean
}

const features: B = {
  10: true
}
```

akuznetsov-gridgain (approved):

PR #450, opened 2019-08-27
GG-23154 Cross-cache tx is mapped on wrong primary when enlisted caches have incompatible assignments.

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/GridDistributedTxMapping.java
```diff
@@ -276,7 +276,7 @@ private void evictReaders(Collection<IgniteTxKey> keys, @Nullable Collection<Ign
         if (entries == null || entries.isEmpty())
             return;
 
-        for (Iterator<IgniteTxEntry> it = entries.iterator(); it.hasNext();) {
+        for (Iterator<IgniteTxEntry> it = entries.iterator(); it.hasNext(); ) {
```
sk0x50: Looks like, this whitespace is not necessary.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/colocated/GridDhtColocatedLockFuture.java
```diff
@@ -1702,94 +1702,117 @@ void onResult(GridNearLockResponse res) {
...
-                    }
-                    finally {
-                        cctx.shared().txContextReset();
-                    }
-                });
-            }
-            else {
-                int i = 0;
+                        // Use remapped version for all subsequent mappings.
+                        synchronized (this) {
```
sk0x50: It seems that `mappings` should be guarded by `synchronized (GridDhtColocatedLockFuture.this)`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearLockFuture.java
```diff
@@ -1669,143 +1670,167 @@ void onResult(GridNearLockResponse res) {
...
 
-                                remap();
-                            }
-                            catch (IgniteCheckedException e) {
-                                onDone(e);
-                            }
-                            finally {
-                                cctx.shared().txContextReset();
+                        // Use remapped version for all subsequent mappings.
+                        synchronized (this) {
```
sk0x50: It seems that `mappings` should be guarded by `synchronized (GridNearLockFuture.this)`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearOptimisticTxPrepareFuture.java
```diff
@@ -155,30 +154,8 @@ private void onError(Throwable e, boolean discoThread) {
...
-
-                            onComplete();
-                        }
-                    });
-
-                    return;
-                }
-            }
-
+        if (ERR_UPD.compareAndSet(this, null, e))
```
sk0x50: Could you please briefly describe why the following calls are no longer relevant? I mean `tx.setRollbackOnly()` and `tx.rollbackAsync()`.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/transactions/TxCrossCacheMapOnInvalidTopologyTest.java
```diff
@@ -0,0 +1,315 @@
...
+            IgniteEx g1 = startGrid(1);
+
+            awaitPartitionMapExchange();
+
+            Ignite client = startGrid("client");
+            assertNotNull(client.cache(CACHE1));
+            assertNotNull(client.cache(CACHE2));
+
+            try (IgniteDataStreamer<Object, Object> streamer = crd.dataStreamer(CACHE1)) {
+                // Put one key per partition.
```
sk0x50: I think this is a typo. It should be `500 keys per partition`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearOptimisticTxPrepareFuture.java
ascherbakoff: Removal of this block fixes the inability to rollback optimistic transaction if failed during preparing. New behavior and reproducing test is introduced by GG-22420. Otherwise deadlock happens while trying to rollback preparing transaction on local node. Removal of this block doesn't seem to break anything visible to TC tests.

PR #451, opened 2019-08-27
GG-23213: Binary metadata should be registered during start of cache

AMashenkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -947,6 +954,80 @@ public void skipFieldLookup(boolean skipFieldLookup) {
         this.skipFieldLookup = skipFieldLookup;
     }
 
+    /**
+     * Register metadata locally for already registered caches.
+     */
+    private void registerMetadataForRegisteredCaches() {
+        for (DynamicCacheDescriptor cacheDescriptor : ctx.cache().cacheDescriptors().values()) {
```
AMashenkov: Braces.

PR #457, opened 2019-08-28
GG-22290 Implemented request of credentials on authentication error.

vsisko:


modules/web-console/frontend/app/modules/agent/AgentManager.service.ts
```diff
@@ -509,7 +509,7 @@ export default class AgentManager {
                     case SuccessStatus.AUTH_FAILED:
                         this.clustersSecrets.get(cluster.id).resetCredentials();
 
-                        throw new Error('Cluster authentication failed. Incorrect user and/or password.');
+                        return this._restOnCluster(event, params);
```
akuznetsov-gridgain: Not sure that I like this change. User will not receive error message if he entered wrong password.

vsisko: Implemented showing of error on repeat showing of login dialog.

PR #462, opened 2019-08-28
GG-23386 .NET: Fix env var handling in PersistenceTest.TestBaselineTopology

ptupitsyn:
Restore environment variable to the previous value in `finally` block


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/PersistenceTest.cs
```diff
@@ -224,68 +227,74 @@ public void TestGridActivationNoPersistence()
         [Test]
         public void TestBaselineTopology()
         {
-            Environment.SetEnvironmentVariable("IGNITE_BASELINE_AUTO_ADJUST_ENABLED", "false");
+            var oldAutoAdjustEnabled = Environment.GetEnvironmentVariable(BaselineAutoAdjustEnabledVar);
```
ashapkin: Just a suggestion. I noticed at least one more usage with the same case - TestIgniteHome. 
Probably it's a good point for refactoring.

Extract a single IDIsposable class,
setup new var in ctor
resolve old in dispose

Thoughts?

ptupitsyn: Agree, done

PR #464, opened 2019-08-29
GG-23450 Add IgniteFeature indicating that Web Console can collect DR events from this grid

nva:
… events from this grid.


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -82,7 +82,13 @@
     /**
      * A mode when data nodes throttle update rate regarding to DR sender load
      */
-    DR_DATA_NODE_SMART_THROTTLING(19);
+    DR_DATA_NODE_SMART_THROTTLING(19),
+
+    /**
```
akuznetsov-gridgain: You should implement unit test for new WC_DR_EVENTS feature, in ggprivate repository in pair branch

PR #468, opened 2019-08-30
GG-22327 Long stacktrace on WC start instead of clear message

nva:


modules/web-console/web-console-server/src/main/resources/META-INF/spring.factories
```diff
@@ -0,0 +1,21 @@
...
+# WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+# See the License for the specific language governing permissions and
+# limitations under the License.
+#
+
+org.springframework.boot.diagnostics.FailureAnalyzer=\
+  org.apache.ignite.console.config.BeanCreationFailureAnalyzer
+
+org.springframework.boot.diagnostics.FailureAnalysisReporter=\
+  org.apache.ignite.console.config.CustomFailureAnalysisReporter
```
akuznetsov-gridgain: Missing empty line in the end of file.

modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/config/BeanCreationFailureAnalyzer.java
```diff
@@ -0,0 +1,38 @@
...
+package org.apache.ignite.console.config;
+
+import org.apache.ignite.internal.util.typedef.X;
+import org.springframework.beans.factory.BeanCreationException;
+import org.springframework.boot.diagnostics.AbstractFailureAnalyzer;
+import org.springframework.boot.diagnostics.FailureAnalysis;
+import org.springframework.boot.diagnostics.analyzer.AbstractInjectionFailureAnalyzer;
+
+/**
+ * An {@link AbstractInjectionFailureAnalyzer} that performs analysis of failures caused
```
akuznetsov-gridgain: Looks like copy-paste.
Lets use "Custom failures analyzer that will report failure reason without polluting logs."

nva: He don't report failure reason, just extract error. 
Look at javadoc of FailureAnalyzer or it implementation in spring framework.
I don't think that we should describe something more.

PR #469, opened 2019-08-30
GG-23453 Use Cluster Id, Tag from cluster to display in Web Console

nva:


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/top/GridTopologyCommandHandler.java
```diff
@@ -315,6 +322,11 @@ else if (e.getValue().getClass().isArray() && !ATTR_IGNITE_FEATURES.equals(e.get
                 }
             }
 
+            IgniteClusterEx cluster = ctx.grid().cluster();
+
+            attrs.put(IGNITE_CLUSTER_ID, cluster.id());
+            attrs.put(IGNITE_CLUSTER_NAME, cluster.tag());
+
```
akuznetsov-gridgain: May be it make sense to add check for new attributes to `org.apache.ignite.internal.processors.rest.JettyRestProcessorAbstractSelfTest#testTopology()` ?

nva: Added test

PR #470, opened 2019-08-30
2

alamar:


modules/core/src/main/java/org/apache/ignite/internal/util/GridDebug.java
```diff
@@ -360,8 +360,15 @@ public static long getCommittedVirtualMemorySize() {
     private static void initHotspotMBean() {
         if (hotspotMBean == null) {
             synchronized (GridDebug.class) {
-                if (hotspotMBean == null)
-                    hotspotMBean = getMBean(HOTSPOT_BEAN_NAME, HotSpotDiagnosticMXBean.class);
+                if (hotspotMBean == null) {
+                    try {
+                        hotspotMBean = getMBean(HOTSPOT_BEAN_NAME,
+                            Class.forName("com.sun.management.HotSpotDiagnosticMXBean"));
```
slukyano: Feels like it should be a constant, same as the bean name.

modules/core/src/main/java/org/apache/ignite/internal/util/GridDebug.java
```diff
@@ -360,8 +360,15 @@ public static long getCommittedVirtualMemorySize() {
...
+                    }
+                    catch (ClassNotFoundException e) {
+                        throw new RuntimeException(e);
```
slukyano: Let's use IgniteException instead. I believe this is our alternative to a generic RuntimeException.

PR #474, opened 2019-09-02
GG-23469 Improve STDOUT logging for WC

nva:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/common/Utils.java
```diff
@@ -16,26 +16,41 @@
...
+import static org.springframework.boot.Banner.Mode.OFF;
 import static org.springframework.security.web.authentication.switchuser.SwitchUserFilter.ROLE_PREVIOUS_ADMINISTRATOR;
 
 /**
  * Utilities.
  */
 public class Utils {
-    /** */
+    /**
+     *
```
akuznetsov-gridgain: should be single line: `/** */`

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/top/GridTopologyCommandHandler.java
```diff
@@ -315,6 +322,11 @@ else if (e.getValue().getClass().isArray() && !ATTR_IGNITE_FEATURES.equals(e.get
                 }
             }
 
+            IgniteClusterEx cluster = ctx.grid().cluster();
```
akuznetsov-gridgain: Why this class was changed? Looks like code from another issue.

PR #475, opened 2019-09-02
NBL-71: CPP: Implement additional missing methods for IgniteCluster class

FedorUporov:


modules/platforms/cpp/core/src/impl/ignite_environment.cpp
```diff
@@ -39,7 +39,47 @@ namespace ignite
...
+                }
+
+                ~IgniteEnvironmentSPHolder()
+                {
+                    std::set<SharedPointer<IgniteEnvironment>*>::iterator it;
+                    for (it = ptrs.begin(); it != ptrs.end(); ++it)
+                        delete* it;
+                }
+
+            } EnvSPHolder;
```
isapego: I'd avoid global states. Let's discuss, if we can store it in other place.

FedorUporov: Discussed.
The fix on java side is required.

PR #488, opened 2019-09-04
GG-23233: Add annotations to spans.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/dto/tracing/Annotation.java
```diff
@@ -0,0 +1,83 @@
...
+     */
+    public Annotation(long ts, String val) {
+        this.ts = ts;
+        this.val = val;
+    }
+
+    /**
+     * @return Timestamp.
+     */
+    public long getTs() {
```
akuznetsov-gridgain: Lets use "full" name for methods: `get(set)Timestamp()`

smoldachev: Fix.

PR #490, opened 2019-09-05
GG-23584

sanpwc:
Data types coverage for cache to jdbc operations.


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/AbstractDataTypesCoverageTest.java
```diff
@@ -534,4 +540,200 @@ public void nestedObjectField(ObjectBasedOnPrimitivesAndCollections nestedObjFie
...
+        private String sqlStrVal;
+
+        /**
+         * Constructor.
+         *
+         * @param byteArr Original value.
+         */
+        public ByteArrayed(byte[] byteArr) {
+            val = byteArr;
+            sqlStrVal = "x'" + Hex.encodeHexString(byteArr) + "'";
```
AMashenkov: There is no need to add a new dependency just for Hex function as
we already have U.byteArray2HexString()

sanpwc: Fixed.

PR #491, opened 2019-09-05
GG-23681

sanpwc:
Data types coverage for jdbc to cache operations.


modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinCacheToJdbcDataTypesCoverageTest.java
```diff
@@ -0,0 +1,620 @@
...
+    @SuppressWarnings("unchecked")
+    private void checkQuery(Function converterToSqlExpVal, BiFunction<Object, Object, Boolean> equalsProcessor,
+        Class<?> dataType, Object originalKey, Object originalVal, String selectQry)
+        throws IgniteCheckedException, SQLException {
+        try {
+            if (writeSyncMode == CacheWriteSynchronizationMode.FULL_ASYNC &&
+                !waitForCondition(new GridAbsPredicateX() {
+                                      @Override public boolean applyx() throws IgniteCheckedException {
+                                          try {
+                                              return stmt.executeQuery(selectQry).next();
```
AMashenkov: Cursor is not closed here.

sanpwc: Seems that it's not a problem cause they are closed either inside stmt.execute() or within afterTest()

PR #494, opened 2019-09-05
GG-23711: Introduce action controllers.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/action/ActionControllerAnnotationProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+    /**
+     * Find the action methods by specific package.
+     *
+     * @param basePkg Base package.
+     */
+    public static Map<String, ActionMethod> findActionMethods(String basePkg) {
+        if (!methods.isEmpty())
+            return Collections.unmodifiableMap(methods);
+
+        Reflections reflections = new Reflections(basePkg);
```
nva: Why not use spring-context for this?

```
ClassPathScanningCandidateComponentProvider scanner =
            new ClassPathScanningCandidateComponentProvider(false);

        scanner.addIncludeFilter(new AnnotationTypeFilter(ActionController.class));

        List<String> collect = scanner
            .findCandidateComponents(basePkg)
```

modules/gmc-agent/src/main/java/org/gridgain/action/ActionControllerAnnotationProcessor.java
```diff
@@ -0,0 +1,59 @@
...
+    }
+
+    /**
+     * Find the action methods by specific package.
+     *
+     * @param basePkg Base package.
+     */
+    public static Map<String, ActionMethod> findActionMethods(String basePkg) {
+        if (!methods.isEmpty())
+            return Collections.unmodifiableMap(methods);
```
nva: Why we copy unmodifiable on each call?

modules/gmc-agent/src/main/java/org/gridgain/action/ActionDispatcher.java
```diff
@@ -0,0 +1,75 @@
...
+     * @param req Request.
+     * @return Completable future with action result.
+     */
+    public CompletableFuture dispatch(Request req) throws Exception {
+        String actName = req.getActionName();
+
+        ActionMethod mtd = findActionMethods().get(actName);
+
+        if (mtd == null)
+            throw new IgniteException("Action not found");
```
nva: use i18 message

modules/gmc-agent/src/main/java/org/gridgain/action/controller/BaselineActionsController.java
```diff
@@ -0,0 +1,56 @@
...
+
+import org.apache.ignite.internal.GridKernalContext;
+import org.gridgain.action.ActionController;
+
+import java.util.concurrent.CompletableFuture;
+
+/**
+ * Baseline actions controller.
+ */
+@ActionController
```
nva: May we should add some base name or prefix to reduce method name?

modules/gmc-agent/src/main/java/org/gridgain/action/ActionDispatcher.java
```diff
@@ -0,0 +1,75 @@
...
+        return invoke(mtd, req.getArgument());
+    }
+
+    /**
+     * Invoke action method.
+     *
+     * @param mtd Method.
+     * @param arg Argument.
+     */
+    private CompletableFuture invoke(ActionMethod mtd, Object arg) throws Exception {
```
nva: I think we invoke actions in separated poll, not in web socket workers?

modules/gmc-agent/src/main/java/org/gridgain/action/ActionMethod.java
```diff
@@ -0,0 +1,60 @@
...
+     */
+    public ActionMethod(Method mtd, Class<?> controllerCls) {
+        this.mtd = mtd;
+        this.controllerCls = controllerCls;
+    }
+
+    /**
+     * @return Controller class.
+     */
+    public Class<?> getControllerCls() {
```
akuznetsov-gridgain: Method name should not be abbreviated.

modules/gmc-agent/src/main/java/org/gridgain/action/controller/BaselineActionsController.java
```diff
@@ -0,0 +1,56 @@
...
+     */
+    public BaselineActionsController(GridKernalContext ctx) {
+        this.ctx = ctx;
+    }
+
+    /**
+     * @param isAutoAdjustEnabled Is auto adjust enabled.
+     */
+    public CompletableFuture<Void> updateAutoAdjustEnabled(boolean isAutoAdjustEnabled) {
+        ctx.grid().cluster().baselineAutoAdjustEnabled(isAutoAdjustEnabled);
```
akuznetsov-gridgain: I think that everything should be executed via compute.
Otherwise we must implement support for security, cancellation, progress, .... features one more time.

modules/gmc-agent/src/main/java/org/gridgain/action/ActionControllerAnnotationProcessor.java
smoldachev: Reflections are faster than spring, and also, if we remove Spring dependencies on the agent, this will help us.

modules/gmc-agent/src/main/java/org/gridgain/action/ActionControllerAnnotationProcessor.java
smoldachev: Fix.

modules/gmc-agent/src/main/java/org/gridgain/action/ActionDispatcher.java
smoldachev: We don't have i18 support in agent.

modules/gmc-agent/src/main/java/org/gridgain/action/controller/BaselineActionsController.java
smoldachev: Add

modules/gmc-agent/src/main/java/org/gridgain/action/ActionDispatcher.java
smoldachev: Agree, add.

modules/gmc-agent/src/main/java/org/gridgain/action/ActionMethod.java
smoldachev: Fix.

modules/gmc-agent/src/main/java/org/gridgain/action/controller/BaselineActionsController.java
smoldachev: Discussed with @agoncharuk offline.

PR #498, opened 2019-09-06
GG-23714 Introduce SCHEDULING feature, when ignite-schedule in classpath

nva:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -95,7 +96,10 @@
     WC_SNAPSHOT_CHAIN_MODE(22),
 
     /** Support of baseline auto adjustment for Web Console. */
-    WC_BASELINE_AUTO_ADJUSTMENT(23)
+    WC_BASELINE_AUTO_ADJUSTMENT(23),
+
+    /** Scheduling disabled. */
+    WC_SCHEDULING_NOT_AVAILABLE(24)
```
akuznetsov-gridgain: I do not like to have "negative" feature. Lets refactor to "WC_SCHEDULING" to make it looks like other features. You should also update Wiki.

nva: It not possible, I tried. You can distinct new node without 'ignite-schedule' module and old node without this feature

PR #511, opened 2019-09-11
GG-22666 Add DR monitoring and management capabilities to control.sh

ibessonov:


modules/core/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrAbstractSubCommand.java
```diff
@@ -0,0 +1,111 @@
...
+            log.severe(CommandLogger.errorMessage(e));
+
+            throw e;
+        }
+
+        return null;
+    }
+
+    /** {@inheritDoc} */
+    protected VisorResultDto execute0(
```
akuznetsov-gridgain: Missing `@Override`;

modules/core/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrCacheCommand.java
```diff
@@ -0,0 +1,301 @@
...
+            }
+
+            return null;
+        }
+    }
+
+    /** */
+    public static class DrCacheArguments implements DrAbstractSubCommand.Arguments<VisorDrCacheTaskArgs> {
+        private final String regex;
+        private final boolean config;
```
akuznetsov-gridgain: Missing `/** */` for fields.

modules/core/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrCacheCommand.java
```diff
@@ -0,0 +1,301 @@
...
+        SenderGroup senderGroup = SenderGroup.ALL;
+        String senderGroupName = null;
+        Action action = null;
+
+        String nextArg;
+
+        //noinspection LabeledStatement
+        args_loop: while ((nextArg = argIter.peekNextArg()) != null) {
+            switch (nextArg.toLowerCase(Locale.ENGLISH)) {
+                case "--config":
```
akuznetsov-gridgain: May be it make sense to declare static constants for all strings like "--config" in order to have them in one place.

modules/core/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrAbstractSubCommand.java
ibessonov: fixed

modules/core/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrCacheCommand.java
ibessonov: fixed

PR #520, opened 2019-09-13
GG-23858 Added passwords-key-store and passwords-key-store-password to generated default.properties file.

akuznetsov-gridgain:


modules/web-console/web-agent/README.txt
```diff
@@ -32,6 +32,7 @@ Configuration file:
     server-trust-store
     server-trust-store-password
     passwords-key-store
+    passwords-key-store-password
```
smoldachev: We don't support the passwords-key-store-password in config, only user prompt or command line argument.

akuznetsov-gridgain: Fixed.

PR #525, opened 2019-09-15
GG-23972 .NET: Deprecate Ignition.LoadAssemblies on public API, add tests for assembly loading, cleanup

ptupitsyn:
* Mark `Ignition.LoadAssemblies` as Obsolete
* Add proper tests for assembly loading
* Reduce code duplication for .NET Core by reusing same `IgniteRunner` class


modules/platforms/dotnet/Apache.Ignite/IgniteRunner.cs
```diff
@@ -90,21 +89,25 @@ internal static void Main(string[] args)
...
                     {
-                        var ignite = Ignition.Start(Configurator.GetConfiguration(allArgs));
-
-                        // Wait until stopped.
-                        var evt = new ManualResetEventSlim(false);
-                        ignite.Stopped += (s, a) => evt.Set();
-                        evt.Wait();
+                        // Load assemblies before instantiating IgniteConfiguration,
+                        // it can reference types from those assemblies.
+                        allArgs = allArgs.LoadAssemblies().ToArray();
```
ashapkin: Should we handle this for a services as well?

PR #542, opened 2019-09-18
GG-24074: Fix issues with invalid headers names and connection over ssl.

smoldachev:


modules/core/src/main/java/org/apache/ignite/internal/processors/gmc/ManagementConfiguration.java
```diff
@@ -35,14 +35,17 @@
     /** */
     private static final long serialVersionUID = 0L;
 
-    /** Default server URI. */
+    /** Gmc server uri. */
+    private static final String GMC_SERVER_URI = "https://gmc.gridgain.com";
```
akuznetsov-gridgain: Port does not required?

PR #546, opened 2019-09-19
GG-24073: Add errors from spring-boot to console.

smoldachev:


modules/web-console/web-console-server/src/main/resources/application.yml
```diff
@@ -14,6 +14,9 @@
 # limitations under the License.
 #
 
+# Use logger settings same as for Ignite
```
akuznetsov-gridgain:
```suggestion
# Use logger settings for Spring same as for Ignite.
```

PR #547, opened 2019-09-19
GG-24046 .NET: Fix JVM thread leak - call DetachCurrentThread on CLR thread exit

ptupitsyn:
* Use thread local storage destructor callback as a way to know when thread is about to exit
* Tested on Windows (.NET Classic, .NET Core), Linux (.NET Core), macOS (.NET Core)
* Additionally: do not call AttachCurrentThread in callbacks - we already have JniEnv pointer, use it


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Unmanaged/UnmanagedThread.cs
```diff
@@ -0,0 +1,214 @@
...
+            }
+
+            throw new InvalidOperationException("Unsupported OS: " + Environment.OSVersion);
+        }
+
+        /// <summary>
+        /// Removes thread exit callback that has been set with <see cref="SetThreadExitCallback"/>.
+        /// </summary>
+        /// <param name="callbackId">Callback id returned from <see cref="SetThreadExitCallback"/>.</param>
+        public static void RemoveThreadExitCallback(int callbackId)
```
isapego: Actually, it not only removes callback, but also calls it in case of Windows: https://docs.microsoft.com/en-us/windows/win32/api/winnt/nc-winnt-pfls_callback_function . Need a test for it.

ptupitsyn: Good point. However:

1) This method is only called from tests - we don't need it during actual Ignite usage since we don't support JVM destruction. 

2) It is fine to call `DetachCurrentThread` multiple times, so it is fine anyway.

PR #548, opened 2019-09-19
GG-24123: Add support of the spring application random port in Agent tests.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -268,7 +270,7 @@ private void connect() {
         nodeConfigurationSrvc = new NodeConfigurationService(ctx, mgr);
         actionSrvc = new ActionService(ctx, mgr);
 
-        execSrvc = Executors.newSingleThreadExecutor();
+        execSrvc = new ThreadPoolExecutor(1, 1, 0L, TimeUnit.MILLISECONDS, new LinkedBlockingQueue<>());
```
akuznetsov-gridgain: Why we can't use one of may `Executors.xxx` methods?

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -364,7 +366,8 @@ private void writeToMetaStorage(ManagementConfiguration cfg) {
                 if (disconnected.compareAndSet(false, true)) {
                     log.error("Lost websocket connection with server: " + curSrvUri);
 
-                    execSrvc.submit(Agent.this::connect0);
+                    if (execSrvc.getActiveCount() == 0)
```
akuznetsov-gridgain: Please, give small comment for this condition.

smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
smoldachev: Because Executors return FinalizableDelegatedExecutorService and we can't get ThreadPoolExecutor.

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -364,9 +364,18 @@ private void writeToMetaStorage(ManagementConfiguration cfg) {
...
                 }
             }
         }
     }
+
+    /**
+     * Send the connect task to executor service.
+     */
+    private void submitConnectTask() {
+        // Submit a reconnection task only if we are no longer trying to reconnect
```
akuznetsov-gridgain: Lets move this comment to javadocs of this method.

modules/gmc-agent/src/test/java/org/gridgain/agent/AgentSelfTest.java
```diff
@@ -112,16 +86,18 @@ public void shouldSendInitialStates() throws Exception {
...
+        IgniteClusterEx cluster = ignite.cluster();
         cluster.active(true);
 
         startGrid(1);
 
         assertWithPoll(
             () -> {
                 TopologySnapshot top = interceptor.getPayload(buildClusterTopologyDest(cluster.id()), TopologySnapshot.class);
-                return top.getNodes().size() == 2;
+                return top != null && top.getNodes().size() == 2;
```
akuznetsov-gridgain: how we can get `top == null` ?

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -96,7 +96,7 @@
     private NodeConfigurationService nodeConfigurationSrvc;
 
     /** Execute service. */
-    private ExecutorService execSrvc;
+    private ThreadPoolExecutor execSrvc;
```
akuznetsov-gridgain: Why we can not use `ExecutorService ` ?

modules/gmc-agent/src/test/java/org/gridgain/agent/AgentSelfTest.java
smoldachev: Data race, when we try to get topology, but agent don't connect to new backend with new port.

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
smoldachev: Because ExecutorService doesn't have getActiveCount method.

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -364,9 +364,17 @@ private void writeToMetaStorage(ManagementConfiguration cfg) {
...
 
-                    execSrvc.submit(Agent.this::connect0);
+                    submitConnectTask();
                 }
             }
         }
     }
+
+    /**
+     * Submit a reconnection task only if we are no longer trying to reconnect.
```
akuznetsov-gridgain:
```suggestion
     * Submit a reconnection task only if there no active connect in progress.
```

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
akuznetsov-gridgain: Lets rename `execSrvc` to `connectPool` ?

smoldachev: Done.

modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
smoldachev: Done.

PR #549, opened 2019-09-20
GG-24116: Change expiration entry cache logic.

smoldachev:


modules/web-console/web-console-server/src/main/java/org/apache/ignite/console/db/CacheHolder.java
```diff
@@ -102,6 +112,6 @@ public V getAndPut(K key, V val) throws TransactionException {
      * @return Underlying cache
      */
     public IgniteCache<K, V> cache() {
-        return cache;
+        return cache.withExpiryPolicy(expiryPlc);
```
akuznetsov-gridgain:
```suggestion
        return expiryPlc  == null ? cache : cache.withExpiryPolicy(expiryPlc);
```

PR #557, opened 2019-09-23
GG-23480 Filtered docker network interface addresses.

vsisko:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2417,14 +2428,16 @@ public static boolean isLocalAddress(InetAddress addr) {
 
             if (itfs != null) {
                 for (NetworkInterface itf : asIterable(itfs)) {
-                    byte[] hwAddr = itf.getHardwareAddress();
+                    if (!IGNORED_INTERFACES.contains(itf.getName())) {
```
nva: Use continue

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2161,9 +2168,11 @@ public static synchronized boolean isLocalHostChanged() throws IOException {
                 List<InetAddress> locAddrs = new ArrayList<>();
 
                 for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
-                    for (InetAddress addr : asIterable(itf.getInetAddresses())) {
-                        if (!addr.isLinkLocalAddress())
-                            locAddrs.add(addr);
+                    if (!IGNORED_INTERFACES.contains(itf.getName())) {
```
nva: use continue

nva (changes requested):
Why you filter this address? What if node will be run in docker and it will has only one interface?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2246,8 +2255,10 @@ private static synchronized InetAddress resetLocalHost() throws IOException {
         else {
             List<NetworkInterface> itfs = new ArrayList<>();
 
-            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces()))
-                itfs.add(itf);
+            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                if (!IGNORED_INTERFACES.contains(itf.getName()))
```
nva: Why we need to do this, if we use interface whose name starts with 'e' as localhost first?

nva (changes requested):

vsisko:
@nva Node in docker has network interface with different system specific name.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2017,10 +2020,14 @@ public static void writeIntCollection(DataOutput out, Collection<Integer> col) t
         try {
             InetAddress inetAddr = InetAddress.getByName(addr);
 
-            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces()))
-                for (InetAddress itfAddr : asIterable(itf.getInetAddresses()))
-                    if (itfAddr.equals(inetAddr))
-                        return itf.getDisplayName();
+            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                if (!IGNORED_INTERFACES.contains(itf.getName())) {
```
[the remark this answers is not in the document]
vsisko: It is really not used. I think we should hide it anywhere.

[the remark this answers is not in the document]
nva: Why we should change something without any reason?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
akuznetsov-gridgain: I think change in IngniteUtils should be also reviewed by core team.

akuznetsov-gridgain: @alamar  Could you review this PR?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2017,10 +2020,15 @@ public static void writeIntCollection(DataOutput out, Collection<Integer> col) t
         try {
             InetAddress inetAddr = InetAddress.getByName(addr);
 
-            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces()))
-                for (InetAddress itfAddr : asIterable(itf.getInetAddresses()))
+            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
```
alamar: can we also add || !itf.isUp()?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2161,6 +2169,9 @@ public static synchronized boolean isLocalHostChanged() throws IOException {
                 List<InetAddress> locAddrs = new ArrayList<>();
 
                 for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                    if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
```
alamar: can we also add || !itf.isUp()?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2246,8 +2257,12 @@ private static synchronized InetAddress resetLocalHost() throws IOException {
         else {
             List<NetworkInterface> itfs = new ArrayList<>();
 
-            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces()))
+            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
```
alamar: can we also add || !itf.isUp()?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2352,15 +2367,16 @@ public static boolean sameMacs(ClusterNode loc, ClusterNode rmt) {
 
             if (itfs != null) {
                 for (NetworkInterface itf : asIterable(itfs)) {
-                    if (!itf.isLoopback()) {
-                        Enumeration<InetAddress> addrs = itf.getInetAddresses();
+                    if (IGNORED_INTERFACE_NAMES.contains(itf.getName()) || itf.isLoopback())
```
alamar: can we also add || !itf.isUp()?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2417,6 +2433,9 @@ public static boolean isLocalAddress(InetAddress addr) {
 
             if (itfs != null) {
                 for (NetworkInterface itf : asIterable(itfs)) {
+                    if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
```
alamar: can we also add || !itf.isUp()?

alamar (changes requested):

alamar (commented):
I don't think we should outright ignore it, just move to the bottom. Please let me consider it.

nva (changes requested):
From thread https://gridgain.slack.com/archives/C2WT0TQGH/p1569504691205800 need to exclude vboxnet0.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2017,10 +2020,15 @@ public static void writeIntCollection(DataOutput out, Collection<Integer> col) t
         try {
             InetAddress inetAddr = InetAddress.getByName(addr);
 
-            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces()))
-                for (InetAddress itfAddr : asIterable(itf.getInetAddresses()))
+            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
+                    continue;
```
nva: Why we want to hide name for such interface? I don't see any of usage of this method
@vsisko It is really not used. I think we should hide it anywhere.
@nva Why we should change something without any reason?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2161,6 +2169,9 @@ public static synchronized boolean isLocalHostChanged() throws IOException {
                 List<InetAddress> locAddrs = new ArrayList<>();
 
                 for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                    if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
+                        continue;
```
nva: Why we change something here (this method was used in TcpCommunicationSpi, TcpDiscoverySpi)?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2246,8 +2257,12 @@ private static synchronized InetAddress resetLocalHost() throws IOException {
         else {
             List<NetworkInterface> itfs = new ArrayList<>();
 
-            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces()))
+            for (NetworkInterface itf : asIterable(NetworkInterface.getNetworkInterfaces())) {
+                if (IGNORED_INTERFACE_NAMES.contains(itf.getName()))
+                    continue;
```
nva: Why we change something here Visor, Web Console doesn't use this method?

nva (changes requested):

akuznetsov-gridgain (changes requested):
Please fix this issue together with Andrey Novikov

PR #563, opened 2019-09-25
GG-24289 Modify Web Console pom.xml files to support new version format: YYYY.MM.pp

nva:


modules/web-console/frontend/app/services/Version.service.ts
```diff
@@ -39,9 +38,10 @@ interface ParsedVersion {
 type VersionInfo = {label: string, ignite: string}
 
 export default class IgniteVersion {
-    webConsole = buildInfo.webConsoleVersion;
-    supportedVersions: VersionInfo[]
-    currentSbj: BehaviorSubject<VersionInfo>
+    // @ts-ignore
+    webConsole = WEB_CONSOLE_VERSION;
```
Klaster1:
```suggestion
    webConsole: string = WEB_CONSOLE_VERSION;
```

PR #571, opened 2019-09-29
GG-24447: [Web Console] Modify frontend/backend Dockerfiles to support deployment to OpenShift cluster

imurchenko:


modules/web-console/docker/console/frontend/Dockerfile
```diff
@@ -21,4 +21,10 @@ COPY ./ignite-web-console-*/frontend /data/www
 COPY ./frontend/nginx/nginx.conf /etc/nginx/nginx.conf
 COPY ./frontend/nginx/web-console.conf /etc/nginx/web-console.conf
 
-EXPOSE 80
+RUN mkdir -p /var/cache/nginx && \
```
nva: Why not https://hub.docker.com/r/twalter/openshift-nginx/dockerfile ?

imurchenko: Let's follow OpenShift's documentation and examples for Dockerfiles https://docs.openshift.com/container-platform/3.11/creating_images/guidelines.html

PR #586, opened 2019-10-03
GG-24340

sanpwc:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
```diff
@@ -0,0 +1,163 @@
...
+import static org.apache.ignite.testframework.GridTestUtils.getFieldValue;
+import static org.apache.ignite.testframework.GridTestUtils.setFieldValue;
+
+/**
+ * Tests verifying that {@link IgniteTxAdapter#logTxFinishErrorSafe}
+ * runs without errors.
+ */
+public class SafeLogTxFinishErrorTest extends GridCommonAbstractTest {
+    /** Logger for listen log messages. */
+    private final ListeningTestLogger log = new ListeningTestLogger(false, super.log);
```
sanpwc: It's better to use `GridCommonAbstractTest.log` instead of `super.log` cause `log` is static.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
```diff
@@ -0,0 +1,163 @@
...
+
+    /** Flag to remove the FailureHandler when creating configuration for node. */
+    private boolean removeFailureHandler;
```
sanpwc: Abbreviation should be used.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
```diff
@@ -0,0 +1,163 @@
...
+        IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);
+
+        cfg.setCacheConfiguration(
+            new CacheConfiguration<>(DEFAULT_CACHE_NAME)
+                .setAffinity(new RendezvousAffinityFunction(false, 32))
+                .setAtomicityMode(TRANSACTIONAL)
+        );
+
+        if (removeFailureHandler)
+            cfg.setFailureHandler(null);
```
sanpwc: Seems that removing failureHandler doesn't actually effect test. In both cases (with and without failureHandler) activeTx.logTxFinishErrorSafe() output is the same. Why we need removeFailureHandler?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @throws Exception If any error occurs.
+     */
+    @Test
+    public void testSafeLogTxFinishErrorWithoutFailureHandler()throws Exception {
+        removeFailureHandler = true;
+
+        checkSafeLogTxFinishError();
+    }
+
```
sanpwc: Seems that we should add more tests in order to check catch clause of logTxFinishErrorSafe()
```
catch (Throwable e0) {
            e.addSuppressed(e0);

            U.error(log, String.format(fmt, commit, CU.txString(this)), e);
        }
```

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
```diff
@@ -0,0 +1,163 @@
...
+            GridNearTxLocal activeTx = (GridNearTxLocal)activeTxs.iterator().next();
+
+            AtomicBoolean containsFailedCompletingTxInLog = new AtomicBoolean();
+
+            /*
+             * When GridNearTxLocal#toString() is called, return string
+             * contains substring "duration", which is calculated as
+             * current time minus start time of the transaction.
+             * For test, we need so that when call GridNearTxLocal#toString(),
```
sanpwc: I don't understand this %)

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
```diff
@@ -0,0 +1,163 @@
...
+             * we get the same string with the same substring "duration".
+             */
+            AtomicLong curTimeMillis = new AtomicLong();
+
+            log.registerListener(logStr -> {
+                /*
+                 * Restore the current time to correctly call 
+                 * GridNearTxLocal#toString().
+                 */
+                setFieldValue(IgniteUtils.class,"curTimeMillis", curTimeMillis.get());
```
sanpwc: Tests are flaky because of gap between setFieldValue() and activeTx.toString(). In order to verify that it's possible to add Thread.sleep(10) after setFieldValue().

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/SafeLogTxFinishErrorTest.java
tkalkirill: According to the test plan https://ggsystems.atlassian.net/browse/GG-24352

PR #589, opened 2019-10-03
GG-24347 Extend test coverage [GG-23478] [GG-21272] Fix information about inclusive and exclusive boundaries in HistogramMetric json

denis-chudov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridTransactionsSystemUserTimeMetricsTest.java
```diff
@@ -354,10 +367,38 @@ private ClientTxTestResult measureClientTransaction(boolean systemDelay, boolean
...
+
+        assertEquals("Must be " + txCnt + " transaction(s), actually were: " + cnt + ". Histogram: " + histogramJson, txCnt, cnt);
+    }
+
+    /**
+     * Creates histogram regex to check transaction histograms.
+     * @return Regex.
+     */
+    private static String histogramRegex() {
+        String regex = toJson(new HistogramMetric(METRIC_TIME_BUCKETS))
```
sk0x50: Looks like we can just return the result of `toJson`. I don't a reason for creating temp variable.

PR #591, opened 2019-10-04
GG-24425: DR: Sender workers doesn't stop on deactivation.

AMashenkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/DiscoveryDataClusterState.java
```diff
@@ -48,6 +48,9 @@
     /** Flag indicating if the cluster in read-only mode. */
     private final boolean readOnly;
 
+    /** Last cluster activation time. */
+    private final long activationTime;
```
korlov42: Looks like this is not activation time, but time when state changed for the last time. May be we should name it stateChangeTime or something like this.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/DiscoveryDataClusterState.java
```diff
@@ -129,7 +135,7 @@ static DiscoveryDataClusterState createTransitionState(
             baselineTopology,
             transitionReqId,
             transitionTopVer,
-            transitionNodes
+            activationTime, transitionNodes
```
korlov42: transitionNodes should be on the next line.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/DiscoveryDataClusterState.java
```diff
@@ -148,11 +155,12 @@ private DiscoveryDataClusterState(
         @Nullable BaselineTopology baselineTopology,
         @Nullable UUID transitionReqId,
         @Nullable AffinityTopologyVersion transitionTopVer,
-        @Nullable Set<UUID> transitionNodes
+        long activationTime, @Nullable Set<UUID> transitionNodes
```
korlov42: transitionNodes should be on the next line.

modules/core/src/test/java/org/apache/ignite/internal/processors/datastreamer/MultipleDataStreamersOnClientTest.java
```diff
@@ -1,253 +1,253 @@
-/*
```
korlov42: could you please highlight what had changed here?

AMashenkov: Contents have differences only in line separators

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/DiscoveryDataClusterState.java
korlov42: but it’s better to correct the condition on GridClusterStateProcessor.java:642. We should get time from message only if msg.activate() is true and state.active() is false, if I understood the idea correctly

PR #596, opened 2019-10-07
GG-24594 Added generation of cluster activation code.

vsisko:


modules/web-console/frontend/app/services/configuration-checker.js
```diff
@@ -0,0 +1,30 @@
+/*
```
Klaster1: I'd move the file from "services" to "domain" dir. This module is not an AngularJS service.

modules/web-console/frontend/app/services/configuration-checker.spec.js
```diff
@@ -0,0 +1,53 @@
...
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+import {isPersistenceEnabled} from './configuration-checker.js'
+import {assert} from 'chai';
+
+suite.only('Test of configuration checker', () => {
```
Klaster1:
```suggestion
suite('Test of configuration checker', () => {
```

modules/web-console/frontend/app/services/configuration-checker.js
vsisko: Moved.

PR #600, opened 2019-10-07
GG-23330 Calculation inconsistency of  auto-adjust baseline default enabled flag on different nodes

ibessonov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/IgnitePdsCacheWalDisabledOnRebalancingTest.java
```diff
@@ -92,16 +90,12 @@
 
     /** {@inheritDoc} */
     @Override protected void beforeTestsStarted() throws Exception {
```
ibessonov: There are a lot of methods left that only invoke their "super" counterpart. Can you remove them?

PR #612, opened 2019-10-09
GG-24779 Common style variable name.

vsisko:


vsisko:
I and @akuznetsov-gridgain noticed that the variable has wrong name and can be missed in search of translation object.

Klaster1:
> I and @akuznetsov-gridgain noticed that the variable has wrong name and can be missed in search of translation object.

It can't be "missed" if you have type checks on. The service can be injected under any reasonable name developer wants.

PR #614, opened 2019-10-10
GG-19109 SQL: Uncorrelated subquery should run only once. 

korlov42:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/SubqueryCacheDisableTest.java
```diff
@@ -0,0 +1,158 @@
...
+    private static final int OUTER_SIZE = 10;
+
+    /** Inner table size. */
+    private static final int INNER_SIZE = 50;
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTestsStarted() throws Exception {
+        super.beforeTestsStarted();
+
+        System.setProperty(IgniteSystemProperties.IGNITE_H2_OPTIMIZE_REUSE_RESULTS, "false");
```
ygerzhedovich: There are our annotation WithSystemProperty, try to use it instead of it

PR #615, opened 2019-10-10
gg-24513

zstan:


modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryFieldImpl.java
```diff
@@ -285,10 +286,19 @@ public int fieldOrder(BinaryObjectExImpl obj) {
             BinaryType expType = ctx.metadata(typeId);
             BinaryType actualType = obj.type();
 
+            String actualTypeName = null;
+
+            try {
+                actualTypeName = actualType.typeName();
+            }
+            catch (Exception e) {
```
zstan: may be catch here more exact exception ? BinaryObjectException ?

modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryFieldExtractionSelfTest.java
```diff
@@ -193,6 +193,47 @@ public void testChangeTypeIdOfBinaryFieldCaseNotFoundExpectedTypeId() throws Exc
...
+        int beforeTypeId = timeValBinObj.typeId();
+
+        String fieldType = binaryContext(marsh).metadata(timeValBinObj.typeId()).fieldTypeName(timeBinField.name());
+
+        Field startField = U.findField(timeValBinObj.getClass(), "start");
+        int start = (int)startField.get(timeValBinObj);
+
+        Field arrField = U.findField(timeValBinObj.getClass(), "arr");
+        byte[] arr = (byte[])arrField.get(timeValBinObj);
+        arr[start + 4] = 42;
```
zstan: public static final int TYPE_ID_POS = 4; looks like public, if u mention it here.

modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryFieldExtractionSelfTest.java
```diff
@@ -193,6 +193,47 @@ public void testChangeTypeIdOfBinaryFieldCaseNotFoundExpectedTypeId() throws Exc
         assertThrows(log, () -> timeBinField.value(timeValBinObj), BinaryObjectException.class, expMsg);
     }
 
+    /**
+     * Checking the exception and its text when changing the typeId by spoil
```
zstan: why a u using "the" here ? )

zstan: what the main purpose of such test ? to check correct err message ?

modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryFieldExtractionSelfTest.java
zstan: looks like more accurate would be code like : arr[start + 4] = beforeTypeId + 1; for example.

modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryFieldExtractionSelfTest.java
tkalkirill: Check that the correct exception and text.

PR #617, opened 2019-10-11
GG-24827 [WC] Can't download Agent if WC was started from arbitrary directory.

nva:


modules/web-console/web-console-server/src/test/java/org/apache/ignite/console/testsuites/WebConsoleTestSuite.java
```diff
@@ -34,16 +38,21 @@
...
+    ActivitiesServiceTest.class,
+    AdminServiceTest.class,
     NotificationServiceTest.class,
-    IsolatedDiscoverySpiSelfTest.class,
-    IsolatedCacheFullApiSelfTest.class
+    AgentDownloadControllerTest.class,
+    PasswordEncoderTest.class,
+    AgentsServiceSelfTest.class,
+    BrowsersServiceTest.class,
+//    TransitionServiceSelfTest.class
```
akuznetsov-gridgain: Why this test was commented?

nva: Reverted

PR #621, opened 2019-10-11
GG-24824 Removed unreleased version.

vsisko:


modules/web-console/e2e/testcafe/fixtures/configuration/basic.js
```diff
@@ -34,6 +34,13 @@ fixture('Basic configuration')
     })
     .after(dropTestDB);
 
+test('Check selector picker version list', async(t) => {
+    const page = new PageConfigurationBasic();
+
+    await t.hover(page.versionPicker._selector)
+        .expect(page.versionPicker.menuItems.count).eql(8, 'Should contains 8 supported versions.');
```
Klaster1: This assertion looks very strange to me. Why 8? Maybe check something else, like a what version are displayed.

vsisko: Now we have 8 supported versions and do not plain to add some additional.

Klaster1: @akuznetsov-gridgain I believe we should check for what version are displayed. Do you agree?

vsisko: Added check of supported versions.

modules/web-console/e2e/testcafe/fixtures/configuration/basic.js
```diff
@@ -34,6 +34,17 @@ fixture('Basic configuration')
     })
     .after(dropTestDB);
 
+test('Check selector picker version list', async(t) => {
+    const knownSupportedVersions = ['Ignite 2.7', 'Ignite 2.6', 'Ignite 2.5', 'Ignite 2.4', 'Ignite 2.3', 'Ignite 2.1', 'Ignite 2.0', 'Ignite 1.x'];
+    const page = new PageConfigurationBasic();
+
+    await t.hover(page.versionPicker._selector)
+        .expect(page.versionPicker.menuItems.count).eql(knownSupportedVersions.length, 'Should contains 8 supported versions.');
```
Klaster1: This assertion is redundant.

vsisko: Version picker should not contain unsupported versions. Changed error message.

PR #626, opened 2019-10-14
GG-23314 "Cancel query" button breaks layout

Klaster1:
https://ggsystems.atlassian.net/browse/GG-23314

### DIFF
--- a/modules/web-console/frontend/public/stylesheets/style.scss
+++ b/modules/web-console/frontend/public/stylesheets/style.scss
@@ -269,7 +269,9 @@ body {
 
     min-height: 100vh;
     display: grid;
-    grid-template-columns: auto 1fr;
+    // Zero min width prevents from content blowing out column width
+    // https://stackoverflow.com/a/43312314/333777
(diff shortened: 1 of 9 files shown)


akuznetsov-gridgain (approved):

modules/web-console/frontend/public/stylesheets/style.scss
```diff
@@ -269,7 +269,9 @@ body {
 
     min-height: 100vh;
     display: grid;
-    grid-template-columns: auto 1fr;
+    // Zero min width prevents from content blowing out column width
+    // https://stackoverflow.com/a/43312314/333777
+    grid-template-columns: auto minmax(0, 1fr);
```
vikvorona: If long text is on the left side, shouldn't it be `minmax(0, auto) 1fr`?

Klaster1: @vikvorona left column if for navigation, `1fr` in right column causes the issue.

vikvorona (approved):

PR #629, opened 2019-10-15
GG-24938: Fix NPE when baseline topology is null.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/agent/Agent.java
```diff
@@ -354,6 +355,11 @@ private void writeToMetaStorage(ManagementConfiguration cfg) {
             });
         }
 
+        /** {@inheritDoc} */
+        @Override public void handleException(StompSession ses, StompCommand cmd, StompHeaders headers, byte[] payload, Throwable e) {
+            log.error("Failed to send message", e);
```
nva: I think we need log more info here and I don't think that this should be error

smoldachev: We don't need nothing from arguments, because we have stack-trace and always can to understand what happened. I've change  the log level on warn.

smoldachev: Also change warning message.

PR #632, opened 2019-10-15
GG-24882 First letter in first word of error cause converted to upper case.

vsisko:


modules/web-console/frontend/app/services/ErrorParser.service.ts
```diff
@@ -70,7 +70,8 @@ export default class {
                     const msgStart = err.message.indexOf(': ', causeIdx + CAUSE_STR.length) + 2;
                     const causeEndLine = err.message.indexOf('\n', msgStart);
                     const msgEnd = err.message.indexOf('[', msgStart);
-                    const cause = err.message.substring(msgStart, msgEnd >= 0 && msgEnd < causeEndLine ? msgEnd : causeEndLine);
+                    const cause = err.message.charAt(msgStart).toUpperCase() +
```
akuznetsov-gridgain: Lets use lodash here: `_.upperFirst()`

Klaster1: I agree with @akuznetsov-gridgain

modules/web-console/frontend/app/services/ErrorParser.service.ts
```diff
@@ -70,8 +70,7 @@ export default class {
                     const msgStart = err.message.indexOf(': ', causeIdx + CAUSE_STR.length) + 2;
                     const causeEndLine = err.message.indexOf('\n', msgStart);
                     const msgEnd = err.message.indexOf('[', msgStart);
-                    const cause = err.message.charAt(msgStart).toUpperCase() +
-                        err.message.substring(msgStart + 1, msgEnd >= 0 && msgEnd < causeEndLine ? msgEnd : causeEndLine);
+                    const cause = _.upperFirst(err.message.substring(msgStart, msgEnd >= 0 && msgEnd < causeEndLine ? msgEnd : causeEndLine));
```
Klaster1: Please either import `_` or `upperFirst`. Do not use global lodash.

modules/web-console/frontend/app/services/ErrorParser.spec.ts
```diff
@@ -20,13 +20,31 @@ import JavaTypes from './JavaTypes.service';
 const parser = new ErrorParser(new JavaTypes());
 
 import { assert } from 'chai';
+import foreach from 'lodash/forEach'
```
Klaster1: Why not use `Array#forEach` instead? I have nothing against Lodash, but this looks like an overkill to me.

PR #644, opened 2019-10-18
GG-20879 .NET: specify expiry policy when creating cache using thin client

ashapkin:


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cache/CacheClient.cs
```diff
@@ -41,6 +43,22 @@ namespace Apache.Ignite.Core.Impl.Client.Cache
...
+        /// <summary>
+        /// Additional flag values for cache operations.
+        /// </summary>
+        private enum ClientCacheRequestFlag : byte
+        {
+            /// <summary>
+            /// No flags
+            /// </summary>
+            None = 0,
+            // 1 << 0 is reserved for Java thin client.
```
ptupitsyn: Let's put `WithKeepBinary` here anyway, for consistency.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cache/CacheClient.cs
```diff
@@ -540,6 +563,14 @@ CacheConfiguration ICacheInternal.GetConfiguration()
             return new CacheClient<TK1, TV1>(_ignite, _name, true);
         }
 
+        /** <inheritDoc /> */
+        public ICacheClient<TK, TV> WithExpiryPolicy(IExpiryPolicy plc)
```
ptupitsyn: Please update `WithKeepBinary` method to pass along `_expiryPolicy`, and add tests to make sure that `WithKeepBinary().WithExpiryPolicy()` works, and vice versa.

ashapkin: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformConfigurationUtils.java
```diff
@@ -289,7 +289,7 @@ public static CacheConfiguration readCacheConfiguration(BinaryRawReaderEx in, Cl
      * @param in Reader.
      * @return Expiry policy factory.
      */
-    private static Factory<? extends ExpiryPolicy> readExpiryPolicyFactory(BinaryRawReader in) {
+    public static Factory<? extends ExpiryPolicy> readExpiryPolicyFactory(BinaryRawReader in) {
```
ptupitsyn: This change and the one and below seem to be unnecessary (no usages outside of this class)

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheRequest.java
```diff
@@ -69,6 +82,15 @@ protected boolean isKeepBinary() {
         return (flags & FLAG_KEEP_BINARY) == FLAG_KEEP_BINARY;
     }
 
+    /**
+     *  Gets a value indicating whether expiry policy is set in this request.
+     *
+     * @return expiry policy flag value.
+     */
+    protected boolean withExpiryPolicy() {
```
ptupitsyn: This method is not used outside of this class. I would inline it in the constructor, and use `expiryPolicy != null` check in `rawCache` to simplify.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheRequest.java
```diff
@@ -80,7 +102,10 @@ protected IgniteCache rawCache(ClientConnectionContext ctx) {
 
         String cacheName = cacheDesc.cacheName();
 
-        return ctx.kernalContext().grid().cache(cacheName);
+        IgniteCache<Object, Object> cache = ctx.kernalContext().grid().cache(cacheName);
+        if (withExpiryPolicy())
+            cache = cache.withExpiryPolicy(expiryPolicy);
+        return cache;
```
ptupitsyn: Blank line is required after the end of the `if` block

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/CacheTest.cs
```diff
@@ -916,6 +917,151 @@ public void TestCacheNames()
...
+            var expiryPolicy = new ExpiryPolicy(TimeSpan.FromMilliseconds(100), null, null);
+            var cache = GetClientCache<int>();
+            var cacheWithExpiryPolicy = cache.WithExpiryPolicy(expiryPolicy);
+
+            cache.Put(val, val);
+            cacheWithExpiryPolicy.Put(val + 1, val);
+
+            Thread.Sleep(200);
+            
+            Assert.IsTrue(cache.ContainsKey(val));
```
ptupitsyn: Let's add cross-checks as well:
```
Assert.IsTrue(cacheWithExpiryPolicy.ContainsKey(val));
Assert.IsFalse(cache.ContainsKey(val + 1);
```

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cache/CacheTest.cs
```diff
@@ -916,6 +917,151 @@ public void TestCacheNames()
...
+            cacheWithExpiryPolicy.Put(val, val);
+
+            // Initially added value is the same.
+            Assert.IsTrue(cacheWithExpiryPolicy.ContainsKey(val));
+
+            // Wait for an expiration.
+            Thread.Sleep(200);
+
+            // Expiry policies should be applied, no cache item exists.
+            Assert.IsFalse(cacheWithExpiryPolicy.ContainsKey(val));
```
ptupitsyn: Here and below - let's add a check for original cache too (without expiry policy).

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cache/CacheClient.cs
```diff
@@ -41,6 +43,28 @@ namespace Apache.Ignite.Core.Impl.Client.Cache
...
+        /// <summary>
+        /// Additional flag values for cache operations.
+        /// </summary>
+        private enum ClientCacheRequestFlag : byte
+        {
+            /// <summary>
+            /// No flags
+            /// </summary>
+            None = 0,
+            /// <summary>
```
ptupitsyn: Blank line missing before XMLDoc

modules/platforms/dotnet/Apache.Ignite.Core/Client/Cache/ICacheClient.cs
```diff
@@ -422,5 +423,16 @@ public interface ICacheClient<TK, TV>
         /// <typeparam name="TV1">Value type in binary mode.</typeparam>
         /// <returns>Cache instance with binary mode enabled.</returns>
         ICacheClient<TK1, TV1> WithKeepBinary<TK1, TV1>();
+
+        /// <summary>
+        /// Returns cache with the specified expired policy set. This policy will be used for each operation
+        /// invoked on the returned cache.
+        /// <para />
+        /// Expiry durations for each operation are calculated only once and then used as constants. Please
```
ptupitsyn: This is not correct, we actually call `IExpiryPolicy` members on every operation (in contrast with thick mode)

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformConfigurationUtils.java
ashapkin: I've made this method public in order to reuse it in ClientCacheConfigurationSerializer.java
when we need to parse a new client config section:
```
     case EXPIRY_POLICY: 
                 .....
```
So, I think it's ok

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheRequest.java
ashapkin: Thanks, converted to private. 
I think the explicit method declaration is a kind of self-documented approach, I would prefer to keep it as is. Thoughts?

ptupitsyn: No objections

PR #647, opened 2019-10-21
GG-25048 Resend events from agent to frontend.

nva:


modules/gmc-agent/src/test/java/org/gridgain/service/event/EventsServiceTest.java
```diff
@@ -0,0 +1,71 @@
...
+
+import static org.apache.ignite.events.EventType.EVT_NODE_LEFT;
+import static org.gridgain.agent.StompDestinationsUtils.buildEventsDest;
+import static org.mockito.Mockito.timeout;
+import static org.mockito.Mockito.verify;
+
+/**
+ * Events service test.
+ */
+public class EventsServiceTest extends AbstractServiceTest {
```
smoldachev: Let's add integration tests with real Ignite nodes.

modules/gmc-agent/src/main/java/org/gridgain/agent/StompDestinationsUtils.java
```diff
@@ -46,6 +46,9 @@
     /** Metrics destination. */
     private static final String METRICS_DEST = "/app/agent/metrics/%s/add";
 
+    /** Events destination. */
+    private static final String EVENTS_DEST = "/app/agent/events/%s/add";
```
smoldachev: Let's use cluster prefix: CLUSTER_PREFIX_DEST + "/events/%s/add"

modules/gmc-agent/src/test/java/org/gridgain/service/event/EventsServiceTest.java
nva: Can you describe simple test plan?

akuznetsov-gridgain: @nva 
Simple plan:
1. start first node.
2. start second node -> `node joined` event should be generated.
3. stop second node -> `node leaved` event should be generated.

nva: Done

PR #652, opened 2019-10-22
GG-25177: Add support of sql query parameters.

smoldachev:


modules/gmc-agent/src/main/java/org/gridgain/utils/QueryUtils.java
```diff
@@ -156,6 +155,9 @@ public static SqlFieldsQuery prepareQuery(QueryArgument arg) {
         if (!F.isEmpty(arg.getCacheName()))
             qry.setSchema(arg.getCacheName());
 
+        if (!F.isEmpty(arg.getParameters()))
+            qry.setArgs(arg.getParameters());
```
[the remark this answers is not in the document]
akuznetsov-gridgain: Agree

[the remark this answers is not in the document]
smoldachev: I've covered the existing https://apacheignite-sql.readme.io/docs/data-types exclude binary and geometry.

Also add test for queries with IN parameter and find interesting case :)

modules/gmc-agent/src/test/java/org/gridgain/action/controller/QueryActionsControllerWithParametersTest.java
```diff
@@ -0,0 +1,294 @@
...
+            ");";
+    }
+
+    /**
+     * @param obj Object.
+     */
+    private DocumentContext parse(Object obj) {
+        try {
+            return JsonPath.parse(mapper.writeValueAsString(obj));
+        } catch (JsonProcessingException e) {
```
akuznetsov-gridgain: `catch` should be on next line.

modules/gmc-agent/src/test/java/org/gridgain/action/controller/QueryActionsControllerWithParametersTest.java
```diff
@@ -0,0 +1,294 @@
...
+            .setAction("QueryActions.executeSqlQuery")
+            .setId(UUID.randomUUID())
+            .setArgument(
+                new QueryArgument()
+                    .setQueryId("qry")
+                    .setQueryText(getAllTypeCreateQueryTable() + insertAllTypesQry_1 + insertAllTypesQry_2)
+                    .setPageSize(10)
+            );
+
+        executeAction(initReq, (r) -> r.getStatus() == COMPLETED);
```
akuznetsov-gridgain: how about to assert returned result also?

smoldachev: I think that we should not check the creation and insertion of results, because if this did not succeed, then we will not get the correct results with the following statements.

modules/gmc-agent/src/test/java/org/gridgain/action/controller/QueryActionsControllerWithParametersTest.java
smoldachev: Ooops, fixed.

PR #653, opened 2019-10-22
GG-24825 Fixed header z-index.

vsisko:


modules/web-console/e2e/testcafe/fixtures/queries/notebook.js
```diff
@@ -92,3 +93,38 @@ test('Sending a request', async(t) => {
...
+            )
+    );
+
+    await t.useRole(user).navigateTo(resolveUrl('/queries/notebooks'));
+    await notebooks.createNotebook('Foo');
+    await t.click(notebooks.getNotebookByName('Foo'));
+
+    await t.click(addQueryButton);
+    await t.expect(paragraph2.cacheList.visible).ok();
+    await t.pressKey('home');
```
Klaster1: I'd replace this with `scrollToPageTop` helper function, similar to `scrollToPageBottom` that we already have. This will convey the intention more clear.

modules/web-console/e2e/testcafe/fixtures/queries/notebook.js
```diff
@@ -92,3 +93,38 @@ test('Sending a request', async(t) => {
...
+    await t.click(paragraph.metadataButton);
+    await t.expect(cacheMetadata.panel.visible).ok('Metadata tree should be visible');
+    await t.eval(() => window.scrollBy(0, 600));
```
Klaster1: This might potentially be unstable due to different layout at different resolutions. I suggest to explicitly set window size for this test.

modules/web-console/e2e/testcafe/fixtures/queries/notebook.js
```diff
@@ -92,3 +93,38 @@ test('Sending a request', async(t) => {
...
+
+    await t.hover(helpMenu.trigger);
```
Klaster1: What happens if there's a dialog open at the same time as metadata panel? Didn't you reproduce that too?

Klaster1 (changes requested):

akuznetsov-gridgain (changes requested):
Please fix this issue together with Ilya Borisov.

PR #659, opened 2019-10-24
GG-22138 OOM Fix. Added close() method for BPlusTree.ForwardCursor.

rkondakov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2Cursor.java
```diff
@@ -82,4 +84,9 @@ public H2Cursor(GridCursor<H2Row> cursor) {
     @Override public boolean previous() {
         throw DbException.getUnsupportedException("previous");
     }
+
+    /** {@inheritDoc} */
+    @Override public void close() throws Exception {
+        cursor.close();
```
mshonichev: shouldn't we add `cur=null` here ?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/IgniteCacheQueryLargeRecordsOomTest.java
```diff
@@ -0,0 +1,134 @@
...
+                @Override public Object call() {
+                    for (int j = 0; j < 10; j++) {
+                        if (Thread.currentThread().isInterrupted())
+                            return null;
+
+                        if (log.isInfoEnabled())
+                            log.info("Iteration " + j);
+
+                        FieldsQueryCursor<List<?>> qry =
+                            cache.query(new SqlFieldsQuery("select * from Person limit 10").setLazy(true));
```
mshonichev: let's try to use different OFFSET here at for different threads or iterations.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2Cursor.java
```diff
@@ -46,12 +49,7 @@ public H2Cursor(GridCursor<H2Row> cursor) {
 
     /** {@inheritDoc} */
     @Override public Row get() {
-        try {
-            return cursor.get();
-        }
-        catch (IgniteCheckedException e) {
-            throw DbException.convert(e);
-        }
+        return cur;
```
mshonichev: We could catch a situation where `.get()` is called before first `.next()`. previously, I guess `cursor.get()` would deal with this, now we silently return `null` while method is not `@Nullable`.

rkondakov: It looks like there is no  special contract for this case for `Cursor` interface. If underlying cursor in the previous version returned 'null', the 'H2Cursor' would also return 'null' in this case. Also other H2's 'Cursor' implementations simply return 'null' values  in such cases.

rkondakov:
Merged

PR #660, opened 2019-10-25
GG-24928 Removed Web Console from 8.8-master.

akuznetsov-gridgain:


pom.xml
```diff
@@ -131,9 +130,8 @@
         </profile>
 
         <profile>
-            <id>all-other</id> <!-- used to update project versions and check all modules compilation -->
-            <modules> <!-- sorted alphabetically -->
-                <module>modules/web-console</module>
+            <id>all-other</id> <!-- Used to update project versions and check all modules compilation. -->
```
vveider: How about to remove this profile completely if no modules left there?

akuznetsov-gridgain: @vveider From my opinion, this is out of this issue scope.

vveider: But empty profile is a bug, leftover.
@nva, WDYT?

akuznetsov-gridgain: @vveider I removed this profile and its usages (DEVNOTES.txt, update-versions.sh, and build.sh in ggprivate repository).

PR #661, opened 2019-10-25
GG-25323: Move agent to appropriate package.

smoldachev:


modules/gmc-agent/src/main/java/org/apache/ignite/agent/service/ActionService.java
```diff
@@ -80,7 +80,7 @@ public void onActionRequest(Request req) {
                 new Response()
                     .setId(req.getId())
                     .setStatus(FAILED)
-                    .setError(new ResponseError(PARSE_ERROR_CODE, ex.getMessage(), ex.getStackTrace()))
+                    .setError(new ResponseError(ResponseError.PARSE_ERROR_CODE, ex.getMessage(), ex.getStackTrace()))
```
akuznetsov-gridgain: static import

smoldachev: Fix.

PR #670, opened 2019-10-29
GG-25379 Fixed NegativeArraySizeException and covered with tests.

akuznetsov-gridgain:


modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentConfiguration.java
```diff
@@ -500,86 +500,85 @@ public void load(URL cfgUrl) throws IOException {
...
             serverUri(val);
 
-        val = props.getProperty("node-uri");
+        items = split(props.getProperty("node-uri"));
 
-        // Intentionally wrapped by ArrayList, for further manipulations.
-        if (val != null)
-            nodeURIs(new ArrayList<>(Arrays.asList(val.split(","))));
+        if (!F.isEmpty(items))
+            nodeURIs(new ArrayList<>(items)); // Intentionally wrapped by ArrayList, for further manipulations.
```
nva: All calls of `split` is wrapped by `new ArrayList<>`, move this into `split` function

akuznetsov-gridgain: Fixed.

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentUtils.java
```diff
@@ -323,6 +328,20 @@ public static void send(Session ses, WebSocketResponse evt, long timeout, TimeUn
         }
     }
 
+    /**
+     * @param s String to split.
+     * @return List of not empty items.
+     */
+    public static List<String> split(String s) {
+        if (F.isEmpty(s))
+            return new ArrayList<>();
```
nva:
```suggestion
            return Collections.emptyList();
```

modules/web-console/web-agent/src/main/java/org/apache/ignite/console/agent/AgentUtils.java
```diff
@@ -323,6 +328,20 @@ public static void send(Session ses, WebSocketResponse evt, long timeout, TimeUn
...
+     * @return List of not empty items.
+     */
+    public static List<String> split(String s) {
+        if (F.isEmpty(s))
+            return Collections.emptyList();
+
+        return Arrays.stream(s.trim().split(","))
+                .map(String::trim)
+                .filter(item -> !F.isEmpty(item))
+                .collect(toList());
```
MaxLosevskoy: I recommend to use `toSet()` collector to avoid duplicates which are very much possible with tokens

modules/web-console/web-agent/src/test/java/org/apache/ignite/console/agent/rest/RestExecutorSelfTest.java
```diff
@@ -504,6 +507,8 @@ public void testLongRunningQueries() throws Throwable {
      * @param qry Query text.
      */
     private boolean executeQuery(RestExecutor exec, String qry) throws Exception {
+        System.out.println(System.currentTimeMillis() + " " + qry + " start");
```
MaxLosevskoy: Is that a debug code?

akuznetsov-gridgain: Yes, I'm in progress of debug this test on TC. It is passing locally and failing on TC

PR #671, opened 2019-10-30
GG-25325 Disable management command in control.sh

nva:


modules/core/src/main/java/org/apache/ignite/internal/commandline/management/ManagementCommands.java
```diff
@@ -77,33 +78,38 @@ private void printHelp(Logger log) {
...
-            throw e;
+                print(log, cfg);
+            }
+            catch (Throwable e) {
+                log.severe("Failed to execute management command='" + args.command().text() + "'");
+                log.severe(CommandLogger.errorMessage(e));
+
+                throw e;
+            }
+        } else {
```
akuznetsov-gridgain: `else` should be on new line.

PR #673, opened 2019-10-30
GG-25424 Non-comparable keys for eviction policy cause failure handle…

ingvard:
… and node shutdown


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/EvictionPolicyFailureHandlerTest.java
```diff
@@ -0,0 +1,189 @@
...
+import org.apache.ignite.transactions.Transaction;
+import org.junit.Test;
+
+import static org.apache.ignite.transactions.TransactionConcurrency.PESSIMISTIC;
+import static org.apache.ignite.transactions.TransactionIsolation.REPEATABLE_READ;
+
+/**
+ * Sorted eviction policy tests.
+ */
+public class EvictionPolicyFailureHandlerTest extends GridCommonAbstractTest {
```
sanpwc: Seems that given test isn't added to a suite.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/EvictionPolicyFailureHandlerTest.java
```diff
@@ -0,0 +1,223 @@
...
+        }
+
+        for (int i = 0; i < 1000; i++) {
+            double d = (double)i;
+
+            if (affinity.isPrimary(node1.localNode(), d))
+                cache.put(d, 1);
+        }
+
+        assertFalse(nodeFailure.get());
```
sanpwc: Could you please add extra check in order to validate that all "double-typed" entries were evicted from on-heap?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/EvictionPolicyFailureHandlerTest.java
ingvard: It's true, I added it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/eviction/EvictionPolicyFailureHandlerTest.java
ingvard: Yes, I added it.

PR #680, opened 2019-10-31
GG-23925 .NET Thin Client: introduce ClusterGroup API

ashapkin:


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroupProjection.cs
```diff
@@ -0,0 +1,126 @@
...
+    /// <summary>
+    /// Projection builder that is used for remote nodes filtering.
+    /// </summary>
+    internal sealed class ClientClusterGroupProjection
+    {
+        /** */
+        private const int Attribute = 1;
+
+        /** */
+        private const int ServerNode = 2;
```
ashapkin: Right now it seems to me that this approach should be simplified.
We could take care only about attributes pairs since most of the filters just rely on a specific attribute name.
Consider forServers() and "org.apache.ignite.cache.client".  This will dramaticaly simplify the filterer. 

In contrast, we can not use this approach to say for "forOlderst()" filter.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/PlatformContextImpl.java
```diff
@@ -197,24 +197,11 @@ public PlatformContextImpl(GridKernalContext ctx, PlatformCallbackGateway gate,
...
-            w.writeCollection(node.hostNames());
-            w.writeLong(node.order());
-            w.writeBoolean(node.isLocal());
-            w.writeBoolean(node.isDaemon());
-            w.writeBoolean(node.isClient());
-            w.writeObjectDetached(node.consistentId());
-            PlatformUtils.writeNodeVersion(w, node.version());
-
-            writeClusterMetrics(w, node.metrics());
+            PlatformUtils.writeNodeInfo(w, node);
```
ptupitsyn: Let's not reuse thick client serialization logic in thin client.
This is error-prone: in thick client this does not go over the wire, but in thin it does, so we have to deal with versions and compatibility.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupGetNodeIdsRequest.java
```diff
@@ -0,0 +1,63 @@
...
+
+import java.util.UUID;
+
+/**
+ * Cluster group get node identifiers request.
+ */
+public class ClientClusterGroupGetNodeIdsRequest extends ClientRequest {
+    /** Topology version. */
+    private final long topVer;
+    /** Client cluster group projection. */
```
ptupitsyn: Please add blank lines between class member declarations

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupProjection.java
```diff
@@ -0,0 +1,110 @@
...
+        return new ClientClusterGroupProjection(mappings);
+    }
+
+    /**
+     * Apply projection.
+     *
+     * @param clusterGrp Cluster group before projection.
+     * @return Cluster group after projection.
+     */
+    public ClusterGroup Apply(ClusterGroup clusterGrp){
```
ptupitsyn: `Apply` -> `apply`

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupProjection.java
```diff
@@ -0,0 +1,110 @@
...
+            switch (code){
+                case ATTRIBUTES:{
+                    HashMap<String, String> attrs = new HashMap<>();
+                    int attrsCnt = reader.readInt();
+                    for(int j = 0; j < attrsCnt; j++){
+                        String attrName = reader.readString();
+                        String attrVal = reader.readString();
+                        attrs.put(attrName, attrVal);
+                    }
+                    mappings.put(ATTRIBUTES, attrs);
```
ptupitsyn: If there is multiple `ATTRIBUTES` entries, we'll silently overwrite the previous one in the map.
We should either support this scenario properly or throw an error.

I lean towards supporting this - just replace Map with List<Pair>. Also avoids unnecessary HashMap overhead - we don't need hashing, just iteration.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupProjection.java
```diff
@@ -0,0 +1,110 @@
...
+                    for (Map.Entry<String, String> attr : attrs.entrySet())
+                        clusterGrp = clusterGrp.forAttribute(attr.getKey(), attr.getValue());
+                    break;
+                }
+                case SERVER_NODE:{
+                    clusterGrp = clusterGrp.forServers();
+                    break;
+                }
+                default:
+                    throw new IllegalArgumentException("Unknown code: " + code);
```
ptupitsyn: This is not possible due to the same check above, right?

modules/platforms/dotnet/Apache.Ignite.Core/Client/IClientClusterGroup.cs
```diff
@@ -0,0 +1,76 @@
...
+ */
+
+namespace Apache.Ignite.Core.Client
+{
+    using System;
+    using System.Collections.Generic;
+    using System.Diagnostics.CodeAnalysis;
+    using Apache.Ignite.Core.Cluster;
+
+    public interface IClientClusterGroup
```
ptupitsyn: Missing XMLDoc

modules/platforms/dotnet/Apache.Ignite.Core/Client/IClientClusterGroup.cs
```diff
@@ -0,0 +1,76 @@
...
+        /// Creates projection for nodes containing given name and value
+        /// specified in user attributes.
+        /// </summary>
+        /// <param name="name">Name of the attribute.</param>
+        /// <param name="val">Optional attribute value to match.</param>
+        /// <returns>Grid projection for nodes containing specified attribute.</returns>
+        IClientClusterGroup ForAttribute(string name, string val);
+
+        /// <summary>
+        /// Creates grid projection for nodes supporting .Net, i.e. for nodes started with Apache.Ignite.exe.
```
ptupitsyn: .Net -> .NET

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupGetNodeIdsRequest.java
```diff
@@ -0,0 +1,63 @@
...
+    private ClientClusterGroupProjection prj;
+
+    /**
+     * Constructor.
+     *
+     * @param reader Reader.
+     */
+    public ClientClusterGroupGetNodeIdsRequest(BinaryRawReader reader) {
+        super(reader);
+        topVer = reader.readLong();
```
ptupitsyn: Let's simplify the protocol and remove topVer. `GetNodes` will simply request new node ids every time. Stateless ftw.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroup.cs
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        /** <inheritDoc /> */
+        public IClientClusterGroup ForServers()
+        {
+            return new ClientClusterGroup(_ignite, _marsh, _projection.ForServerNodes());
+        }
+
+        /** <inheritDoc /> */
+        public IClientClusterGroup ForPredicate(Func<IClusterNode, bool> p)
```
ptupitsyn: Let's remove this method - LINQ is there to filter by predicate locally.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroup.cs
```diff
@@ -0,0 +1,311 @@
...
+            return GetNodes().FirstOrDefault();
+        }
+
+        /// <summary>
+        /// Refresh projection nodes.
+        /// </summary>
+        /// <returns>Nodes.</returns>
+        private IList<IClusterNode> RefreshNodes()
+        {
+            long oldTopVer = Interlocked.Read(ref _topVer);
```
ptupitsyn: Let's simplify things and get rid of topVer handling. Just request new ids, that's it.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroup.cs
```diff
@@ -0,0 +1,311 @@
...
+        }
+
+        /// <summary>
+        /// Update topology.
+        /// </summary>
+        /// <param name="remoteTopVer">Remote topology version.</param>
+        /// <param name="nodeIds">Node ids.</param>
+        internal void UpdateTopology(long remoteTopVer, List<Guid> nodeIds)
+        {
+            lock (this)
```
ptupitsyn: Don't `lock (this)` on objects that are exposed on public API - someone may lock on this object too.
`lock (this)` is an antipattern in general; we use it carefully in some places to reduce allocations, but here we can't do this.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroup.cs
```diff
@@ -0,0 +1,311 @@
...
+            Func<IBinaryRawReader, T> readFunc)
+        {
+            return _ignite.Socket.DoOutInOp(opId, stream => WriteRequest(writeAction, stream),
+                stream => ReadRequest(readFunc, stream), HandleError<T>);
+        }
+
+        /// <summary>
+        /// Writes the request.
+        /// </summary>
+        protected void WriteRequest(Action<IBinaryRawWriter> writeAction, IBinaryStream stream)
```
ptupitsyn: Can be private

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroup.cs
```diff
@@ -0,0 +1,311 @@
...
+            {
+                UpdateTopology(topology.Item1, topology.Item2);
+                UpdateNodesInfo(topology.Item2);
+            }
+
+            // No topology changes.
+            Debug.Assert(_nodeIds != null, "At least one topology update should have occurred.");
+
+            // Local lookup with a native predicate is a trade off between complexity and consistency.
+            var nodes = _nodeIds.Select(id => _ignite.GetNode(id)).Cast<IClusterNode>();
```
ptupitsyn: Avoid LINQ (performance, allocations)

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/Cluster/ClientClusterGroup.cs
ptupitsyn: Or maybe I'm wrong, this provides an easy way to have filtered `IClientClusterGroup` instance, let's keep it.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupGetNodeIdsRequest.java
```diff
@@ -0,0 +1,63 @@
...
+    @Override public ClientResponse process(ClientConnectionContext ctx) {
+
+        IgniteClusterEx cluster = ctx.kernalContext().grid().cluster();
+        long curTopVer = cluster.topologyVersion();
+
+        if (curTopVer <= topVer)
+            return new ClientBooleanResponse(requestId(), false);
+
+        ClusterGroup clusterGrp = prj.Apply(cluster);
+        UUID[] nodeIds = clusterGrp.nodes().stream().map(ClusterNode::id).toArray(UUID[]::new);
```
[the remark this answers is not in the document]
ashapkin: That's an interesting comment. 
Is there any internal Ignite part where we could stay with streams/LINQ except for the unit testing?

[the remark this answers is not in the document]
ptupitsyn: Yes, we use LINQ sparingly for things like one-time initialization and initial configuration; there are other exceptions. But in general, we strive to keep allocations to a minimum.

`ToArray`/`ToList` are the most wasteful methods, because they don't know the collection size upfront, and cause multiple array reallocations. So these are the first candidates to replace with loops.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupGetNodeIdsResponse.java
```diff
@@ -0,0 +1,60 @@
...
+        super.encode(ctx, writer);
+
+        writer.writeBoolean(true);
+        writer.writeLong(topVer);
+
+        // At this moment topology version might have advanced, and due to this race
+        // we return outdated top ver to the callee. But this race is benign, the only
+        // possible side effect is that the user will re-request nodes and we will return
+        // the same set of nodes but with more recent topology version.
+        writer.writeUuidArray(nodeIds);
```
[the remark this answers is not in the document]
ptupitsyn: `doWriteUuid` writes an extra byte before each 8-byte value - that's what I'd like to avoid. Those extra bytes increase the response size for no reason.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupProjection.java
```diff
@@ -0,0 +1,95 @@
...
+    /**
+     * Reads projection from a stream.
+     * @param reader Reader.
+     * @return Projection.
+     */
+    public static ClientClusterGroupProjection read(BinaryRawReader reader) {
+        if (!reader.readBoolean())
+            return new ClientClusterGroupProjection(null);
+
+        Function<ClusterGroup, ClusterGroup> factory = clusterGrp ->
```
ptupitsyn: This function closes over `reader`, postponing actual stream reading until `apply` us called. This assumes that no one reads from that reader between `read` and `apply` calls, which works for now, but is fragile and will be hard to debug in case it fails in future.

Throughout our code base, `read` methods are supposed to read their share of data right away. Please refactor accordingly.

I think we can use functional composition here, something like:
```
            Function<ClusterGroup, ClusterGroup> factory = c -> c;

            int cnt = reader.readInt();
            for (int i = 0; i < cnt; i++) {
                short code = reader.readShort();
                final Function<ClusterGroup, ClusterGroup> prev = factory;
                switch (code) {
                    case ATTRIBUTE: {
                        String attName = reader.readString();
                        String attrVal = reader.readString();
                        factory = c -> prev.apply(c).forAttribute(attrName, attrVal);
                        break;
                    }
              ...
              return new ClientClusterGroupProjection(factory);
```

So the code is very similar, but reading is done right away.

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Client/Cluster/ClientClusterGroupTests.cs
```diff
@@ -0,0 +1,312 @@
...
+        /// <summary>
+        /// Test cluster group throws exception if predicate is null.
+        /// </summary>
+        [Test]
+        public void TestClusterGroupForPredicateThrowsExceptionIfItNull()
+        {
+            TestDelegate action = () => Client.GetCluster().ForPredicate(null);
+
+            var ex = Assert.Throws<ArgumentNullException>(action);
+            Assert.AreEqual("Value cannot be null.\r\nParameter name: p", ex.Message);
```
ptupitsyn: Does not work on Unix-like systems, please use `Environment.NewLine`

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Apache.Ignite.Core.Tests.csproj
```diff
@@ -153,6 +153,7 @@
     <Compile Include="Client\Cache\TestKey.cs" />
     <Compile Include="Client\Cache\TestKeyWithAffinity.cs" />
     <Compile Include="Client\ClientTestBase.cs" />
+    <Compile Include="Client\Cluster\ClientClusterGroupTests.cs" />
```
ptupitsyn: Don't forget to add to `DotNetCore` project (I've done it this time).

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupGetNodeIdsResponse.java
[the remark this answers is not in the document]
ashapkin: Rewritten to two writeLong

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cluster/ClientClusterGroupProjection.java
ashapkin: Refactored in .NET way with internal classes.

PR #685, opened 2019-11-05
GG-25386: Implemented Schema test plans for C++, ODBC and .NET

isapego:


modules/platforms/cpp/core-test/src/cache_query_schema_test.cpp
```diff
@@ -0,0 +1,311 @@
...
+
+using ignite::impl::binary::BinaryUtils;
+
+/**
+ * Ensure that cursor is empy fails.
+ *
+ * @param cur Cursor.
+ */
+template<typename Cursor>
+void CheckCursorEmpty(Cursor& cur)
```
ptupitsyn: In .NET code base we are starting to adapt more descriptive test names, maybe we can do that for C++ too.
In this case it can be something like `ChechEmptyCursorGetNextThrowsException`

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Query/CacheDmlQueriesTestSchema.cs
```diff
@@ -0,0 +1,260 @@
...
+    using Apache.Ignite.Core.Common;
+    using NUnit.Framework;
+
+    /// <summary>
+    /// Tests Data Manipulation Language queries related to schema.
+    /// </summary>
+    public class CacheDmlQueriesTestSchema
+    {
+        /// Table name
+        private const string TBL_NAME = "T1";
```
ptupitsyn: Our naming convention for constants is `UpperCamel`

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Query/CacheDmlQueriesTestSchema.cs
```diff
@@ -0,0 +1,259 @@
...
+            Sql("DROP TABLE " + SCHEMA_NAME_4 + '.' + TBL_NAME);
+        }
+
+        /// <summary>
+        /// Verify tables.
+        /// </summary>
+        private void VerifyTables()
+        {
+            Sql("SELECT SCHEMA_NAME, KEY_ALIAS FROM SYS.TABLES ORDER BY SCHEMA_NAME", res => {
+                res.Equals(new List<List<object>> {
```
ptupitsyn: Here and below - `Equals` does not do what you expect:
* It is a reference comparison, will be always `false` here
* It returns `bool`, which is just ignored, so the test does not do anything

Instead, we should say `Assert.AreEqual(new List..., res)` (first arg is `expected`, second `actual`).

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Query/CacheDmlQueriesTestSchema.cs
```diff
@@ -0,0 +1,259 @@
...
+            if (validator != null)
+                validator.Invoke(res);
+        }
+ 
+        /// <summary>
+        /// Generate one-row result set.
+        /// </summary>
+        private static List<List<object>> OneRowList(params object[] vals)
+        {
+            return new List<List<object>>{ new List<object>{ vals } };
```
ptupitsyn: Inner list here will contain the entire `vals` as a single element.
I think it should be `return new List<List<object>>{ vals.ToList() };`

modules/platforms/cpp/core-test/config/cache-query-schema-default.xml
```diff
@@ -0,0 +1,59 @@
...
+    <bean abstract="true" id="grid.cfg" class="org.apache.ignite.configuration.IgniteConfiguration">
+        <property name="localHost" value="127.0.0.1"/>
+        <property name="connectorConfiguration"><null/></property>
+
+        <property name="sqlSchemas">
+            <list>
+                <value>SCHEMA_1</value>
+                <value>SCHEMA_2</value>
+                <value>"ScHeMa3"</value>
+                <value>SCHEMA_4</value>
```
mshonichev: you've missed the point. SCHEMA_4 should be set as cache property, not as sqlSchemas. Is it possible in C++/.NET?

modules/platforms/cpp/odbc-test/config/queries-schema-default.xml
```diff
@@ -0,0 +1,67 @@
...
+                <property name="portRange" value="10"/>
+            </bean>
+        </property>
+
+        <property name="sqlSchemas">
+            <list>
+                <value>SCHEMA_1</value>
+                <value>SCHEMA_2</value>
+                <value>"ScHeMa3"</value>
+                <value>SCHEMA_4</value>
```
mshonichev: same comment here, there should be CacheConfiguration with schema property set to SCHEMA_4 instead of simply propagating SCHEMA_4 to sqlSchemas

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Query/CacheDmlQueriesTestSchema.cs
```diff
@@ -0,0 +1,254 @@
...
+        public void FixtureSetUp()
+        {
+            var cfg = new IgniteConfiguration(TestUtils.GetTestConfiguration())
+            {
+                SqlSchemas = new List<string>
+                {
+                    SchemaName1,
+                    SchemaName2,
+                    QSchemaName4,
+                    SchemaName4
```
mshonichev: same as above.

modules/platforms/cpp/core-test/config/cache-query-schema-default.xml
isapego: No, this is not possible

modules/platforms/cpp/core-test/src/cache_query_schema_test.cpp
```diff
@@ -0,0 +1,311 @@
...
+    int i;
+
+    MixedPred() : i(0)
+    {
+        // No-op.
+    }
+
+    bool operator()()
+    {
+        return (++i & 1) == 0;
```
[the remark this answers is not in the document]
isapego: Hey, I've just took Java tests and implemented them in C++/.NET, so...

modules/platforms/cpp/core-test/src/cache_query_schema_test.cpp
```diff
@@ -0,0 +1,311 @@
...
+    Sql("DELETE FROM " + SCHEMA_NAME_2 + '.' + TABLE_NAME);
+    Sql("DELETE FROM " + Q_SCHEMA_NAME_3 + '.' + TABLE_NAME);
+    Sql("DELETE FROM " + SCHEMA_NAME_4 + '.' + TABLE_NAME);
+
+    Sql("CREATE INDEX t1_idx_1 ON " + SCHEMA_NAME_1 + '.' + TABLE_NAME + "(s1_val)");
+    Sql("CREATE INDEX t1_idx_1 ON " + SCHEMA_NAME_2 + '.' + TABLE_NAME + "(s2_val)");
+    Sql("CREATE INDEX t1_idx_1 ON " + Q_SCHEMA_NAME_3 + '.' + TABLE_NAME + "(s3_val)");
+    Sql("CREATE INDEX t1_idx_1 ON " + SCHEMA_NAME_4 + '.' + TABLE_NAME + "(s4_val)");
+
+    Sql("SELECT * FROM " + SCHEMA_NAME_1 + '.' + TABLE_NAME);
```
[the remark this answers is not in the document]
isapego: Well, again, I've just took Java tests and implemented them in C++. I've thought that this is going to be enough since they were checked and accepted.

modules/platforms/cpp/odbc-test/config/queries-schema-default.xml
isapego: Can not be done in C++/.NET

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/Query/CacheDmlQueriesTestSchema.cs
isapego: -- // --

PR #688, opened 2019-11-05
GG-24762

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
```diff
@@ -485,6 +486,9 @@ private boolean finish(boolean commit,
             req.writeVersion(tx.writeVersion() != null ? tx.writeVersion() : tx.xidVersion());
 
             try {
+                if (isNull(cctx.discovery().node(n.id())))
+                    throw new ClusterTopologyCheckedException("Unable to send message (node left topology): " + n);
```
zstan: i hope no need to throw exception here, "fut.onNodeLeft" would be enough.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/local/GridCacheLocalRollbackSelfTest.java
```diff
@@ -0,0 +1,385 @@
...
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
zstan: too more licenses.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/local/GridCacheLocalRollbackSelfTest.java
```diff
@@ -0,0 +1,385 @@
...
+
+/**
+ * Class for testing rollback transaction with local cache.
+ */
+public class GridCacheLocalRollbackSelfTest extends GridCommonAbstractTest {
+    /** Logger for listen log messages. */
+    private static ListeningTestLogger listeningLog;
+
+    /** Number of nodes. */
+    private static final int NODES = 4;
```
zstan: static final - need to be in first enumeration priority.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/local/GridCacheLocalRollbackSelfTest.java
```diff
@@ -0,0 +1,385 @@
...
+        super.beforeTestsStarted();
+
+        listeningLog = new ListeningTestLogger(false, GridAbstractTest.log);
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTest() throws Exception {
+        super.beforeTest();
+
+        setFieldValue(GridDhtTxFinishFuture.class, "log", null);
```
zstan: plz add some comments here, why u need all this.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/local/GridCacheLocalRollbackSelfTest.java
```diff
@@ -0,0 +1,385 @@
...
+            grid(nodes),
+            cacheName,
+            range(txCnt / 2, txCnt).mapToObj(keys::get).collect(toList())
+        );
+
+        int stoppedNodeId = 2;
+
+        stopGrid(stoppedNodeId);
+
+        LogListener logLsnr = new LogListener();
```
zstan: do u really can`t use default LogListener with matches(...).andMatches syntax ?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
tkalkirill: I think it's worth logging the problem.

tkalkirill: You're right, corrected.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
```diff
@@ -485,6 +486,12 @@ private boolean finish(boolean commit,
             req.writeVersion(tx.writeVersion() != null ? tx.writeVersion() : tx.xidVersion());
 
             try {
+                if (isNull(cctx.discovery().node(n.id()))){
+                    log.error("Unable to send message (node left topology): " + n);
+
+                    fut.onNodeLeft();
```
sanpwc: Seems that it's not equivalent to the logic that was used earlier:
1. Originally as part of `GridCacheIoManager.send()`  we've called `onSend()` that contains following code:
```
        if (!onSend(msg, node.id()))
            return;
```
and 
```
        if (msg.error() != null && cctx.kernalContext().isStopping())
            return false;
```
so, originally we might successfully finish future even if there's no expected node in the map (if cluster is stopping), now  we will call fut.onNodeLeft();

Besides that, we don't call `cctx.deploy().prepare((GridCacheDeployable)msg);`. Not sure whether it's important or not.

sanpwc: Similar is about tracing `org.apache.ignite.internal.managers.communication.GridIoManager#send` - we won't trace message if there's no node in the map, originally we did.

Same is about checking communication listener. Originally we threw IgniteCheckedException and while catching it within 
```
 catch (IgniteCheckedException e) {
                if (!cctx.discovery().alive(node.id()) || !cctx.discovery().pingNode(node.id()))
                    throw new ClusterTopologyCheckedException("Node left grid while sending message to: " + node.id(), e);
```
we only threw ClusterTopologyCheckedException if node wasn't alive from spi point of view (or ping failure) and not discoCache one:
```
    @Nullable public ClusterNode getAlive(UUID nodeId) {
        assert nodeId != null;

        return getSpi().getNode(nodeId); // Go directly to SPI without checking disco cache.
    }
```
otherwise we've made another attempt of sending message.

sanpwc: I believe that there might be other similar issues.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
```diff
@@ -500,8 +507,11 @@ private boolean finish(boolean commit,
             }
             catch (IgniteCheckedException e) {
                 // Fail the whole thing.
-                if (e instanceof ClusterTopologyCheckedException)
+                if (e instanceof ClusterTopologyCheckedException) {
+                    log.error("Error sending message to remote node [node=" + n + ']', e);
```
sanpwc: We don't log (at least not in all cases) ClusterTopologyCheckedException. Why you've decided to all logging here?
See throwing IgniteCheckedException from within `org.apache.ignite.internal.managers.communication.GridIoManager#send` for an example of not logging ClusterTopologyCheckedException.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
```diff
@@ -485,6 +486,12 @@ private boolean finish(boolean commit,
             req.writeVersion(tx.writeVersion() != null ? tx.writeVersion() : tx.xidVersion());
 
             try {
+                if (isNull(cctx.discovery().node(n.id()))){
+                    log.error("Unable to send message (node left topology): " + n);
```
sanpwc: We don't log (at least not in all cases) ClusterTopologyCheckedException. Why you've decided to all logging here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
```diff
@@ -485,6 +486,12 @@ private boolean finish(boolean commit,
             req.writeVersion(tx.writeVersion() != null ? tx.writeVersion() : tx.xidVersion());
 
             try {
+                if (isNull(cctx.discovery().node(n.id()))){
```
sanpwc: Seems that it's not valid. If we will use some other implementation of CommunicationSPI instead of default one we might override node() method of IgniteSpiContext that won't use `ctx.discovery().node(nodeId)`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtTxFinishFuture.java
tkalkirill: I’ll clean up this mistake.

PR #692, opened 2019-11-06
GG-25039 Updates of REST tests.

vsisko:


MaxLosevskoy:
Ping

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorAbstractSelfTest.java
```diff
@@ -2955,8 +2980,12 @@ public VisorGatewayArgument forNodes(Collection<ClusterNode> nodes) {
          * @return This helper for chaining method calls.
          */
         public VisorGatewayArgument arguments(Object... vals) {
-            for (Object val : vals)
-                put("p" + idx++, String.valueOf(val));
+            if (vals != null) {
```
akuznetsov-gridgain: What if `vals` will be empty array? May be we should use `!F.isEmpty(vals)` ?

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorAbstractSelfTest.java
```diff
@@ -2955,8 +2980,12 @@ public VisorGatewayArgument forNodes(Collection<ClusterNode> nodes) {
...
+                for (Object val : vals)
+                    put("p" + idx++, String.valueOf(val));
+            }
+            else
+                put("p" + idx++, null);
```
MaxLosevskoy: Still NPE for
```
arguments("String", null);
```
We can go with the following:
```
for (Object val : vals)
                put("p" + idx++, val  == null ? null : String.valueOf(val));
```
or
```
for (Object val : vals)
                if (val == null){
                       throw new SomeMeaningfulException();
                }
                put("p" + idx++, String.valueOf(val));
```

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorAbstractSelfTest.java
MaxLosevskoy: You'll get a link to memory section, e.g. _[I@2ff4acd0_, a default object behavior.

It was always like this in the past, is that wrong?

vsisko: Added checks on methods of VisorGatewayArtument class to prevent errors in time of test creation.
Some methods of VisorGatewayArtument are renamed.

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorAbstractSelfTest.java
```diff
@@ -2949,43 +2976,48 @@ public VisorGatewayArgument forNodes(Collection<ClusterNode> nodes) {
...
-            for (Object val : vals)
-                put("p" + idx++, String.valueOf(val));
+        public VisorGatewayArgument addArguments(@Nullable Object... vals) {
+            if (idx == 3)
+                throw new IllegalStateException("Task argument class should be declared before adding of additional arguments. " +
+                    "Use VisorGatewayArgument.setTaskArgument first");
 
-            return this;
-        }
+            if (vals != null && F.isEmpty(vals))
```
akuznetsov-gridgain: Change to `if (F.isEmpty(vals)) ....` and move inside next if block.

vsisko: null value of vararg argument is not array with null value. It is null reference to array object.

PR #701, opened 2019-11-08
GG-25433 Test plan [GG-25433] Extend test coverage [GG-22696] [IGNITE…

ingvard:
…-10808] Discovery message queue may build up with TcpDiscoveryMetricsUpdateMessage


modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MetricsCompactionTest.java
```diff
@@ -0,0 +1,276 @@
...
+        assertTrue(GridTestUtils.waitForCondition(() -> records.size() == 2, 10_000));
+        assertTrue(records.stream().anyMatch(m -> m.passedLaps(srv.localNode().id()) == 1 && m.metrics().containsKey(TEST_UUID)));
+        assertTrue(records.stream().anyMatch(m -> m.passedLaps(srv.localNode().id()) == 0 && m.metrics().containsKey(TEST_UUID)));
+    }
+
+    /**
+     *
+     */
+    @Test
+    public void testQueueDoesNotContainMoreTwoMetricMessages() throws Exception {
```
sk0x50: I would prefer the following name - testQueueDoesNotContainMore*Than*TwoMetricMessages

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MetricsCompactionTest.java
```diff
@@ -0,0 +1,276 @@
...
+                    latchs.get(1).await();
+                }
+                catch (InterruptedException ignore) {
+                }
+            }
+        });
+
+        latchs.put(1, new CountDownLatch(1)); // Blocks srv discovery for sending
+
+        Thread.sleep(1_000);
```
sk0x50: Why do we need to sleep here for 1 sec?

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MetricsCompactionTest.java
```diff
@@ -0,0 +1,276 @@
...
+import org.apache.ignite.internal.IgniteEx;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.spi.discovery.tcp.messages.TcpDiscoveryAbstractMessage;
+import org.apache.ignite.spi.discovery.tcp.messages.TcpDiscoveryMetricsUpdateMessage;
+import org.apache.ignite.testframework.GridTestUtils;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/**
+ *
```
sk0x50: Please add javadoc here and below.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MetricsCompactionTest.java
ingvard: Thanks. Fixed.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MetricsCompactionTest.java
ingvard: Added

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/MetricsCompactionTest.java
ingvard: I added another type of dummy message, it should allow remove it.

PR #702, opened 2019-11-08
GG-24207 Add transactions support to thin clients

ashapkin:


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
```diff
@@ -219,4 +243,61 @@ public ClientAffinityTopologyVersion checkAffinityTopologyVersion() {
...
         }
     }
+
+    /**
+     * Next transaction id for this connection.
+     */
+    public int nextTxId() {
+        int txId = txIdSeq.incrementAndGet();
+
+        return txId == 0 ? txIdSeq.incrementAndGet() : txId;
```
ashapkin: We cant get 0 with incrementAndGet on a default AtomicInteger?

PR #721, opened 2019-11-15
GG-25821: Fix RetryableSender send method.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
```diff
@@ -119,28 +119,28 @@ boolean sendInternal(String dest, List<T> elements) throws InterruptedException
     /**
      * @param element Element to send.
      */
-    public void send(String dest, T element) {
+    public void send(String dest, Object element) {
         if (element != null)
             addToQueue(new IgniteBiTuple<>(dest, Collections.singletonList(element)));
```
nva: static import

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
```diff
@@ -119,28 +119,28 @@ boolean sendInternal(String dest, List<T> elements) throws InterruptedException
...
         if (elements != null)
             splitOnBatches(elements).stream().map(e -> new IgniteBiTuple<>(dest, e)).forEach(this::addToQueue);
     }
 
     /**
      * @param list List.
      */
-    private List<List<T>> splitOnBatches(List<T> list) {
-        List<T> batch = new ArrayList<>();
+    private List<List<Object>> splitOnBatches(List<?> list) {
```
nva:
```suggestion
    private List<List<Object>> splitOnBatches(List<Object> list) {
```

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
```diff
@@ -119,28 +119,28 @@ boolean sendInternal(String dest, List<T> elements) throws InterruptedException
...
+    public void send(String dest, Object element) {
         if (element != null)
             addToQueue(new IgniteBiTuple<>(dest, Collections.singletonList(element)));
     }
 
     /**
      * @param elements Elements to send.
      */
-    public void send(String dest, List<T> elements) {
+    public void sendList(String dest, List<?> elements) {
```
nva:
```suggestion
    public void sendList(String dest, List<Object> elements) {
```

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
```diff
@@ -119,28 +119,28 @@ boolean sendInternal(String dest, List<T> elements) throws InterruptedException
...
         if (elements != null)
```
nva: use F.isEmpty

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
nva: Call `sendList`

smoldachev: Fix.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
smoldachev: We cannot use Object here, because List<Span> is the invariance of List<Object> for fix, we must use List<? extends the object>, but it can be simplified to a List<?>.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
smoldachev: Fix.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ws/RetryableSender.java
smoldachev: Why we should try to split on batches message with single element?

PR #730, opened 2019-11-19
GG-23189: SQL: optimization for OR condition is not applied for prepared statement with params.

AMashenkov:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -715,6 +715,12 @@ public void testInWithEqualsIdxUsage() throws Exception {
 
         assertTrue(checkIdxUsage(res, idxName));
 
+        //Check OR -> IN optimization is applied.
+        res = qryProc.querySqlFields(new SqlFieldsQuery("explain select * from " + TEST_TBL_NAME +
+            " where (LANG = ?1 OR LANG = ?2) and ADDRESS = 5").setArgs(3, 4), true).getAll();
```
ygerzhedovich: Let's check additional case: OR clause for different columns, like a (LANG = ?1 OR NAME=?2) and ADDRESS = 5.

AMashenkov: Mmm.. What do you expect in this case?

PR #733, opened 2019-11-20
GG-25867 Fix log level for messages and thread dumps for ignored fail…

alievmirza:
…ures

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
@@ -20,6 +20,7 @@ import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.IgniteSystemProperties;
 import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.failure.AbstractFailureHandler;
 import org.apache.ignite.failure.FailureContext;
 import org.apache.ignite.failure.FailureHandler;
 import org.apache.ignite.failure.NoOpFailureHandler;
@@ -36,8 +37,15 @@ import org.apache.ignite.internal.util.typedef.internal.U;
  */
 public class FailureProcessor extends GridProcessorAdapter {
     /** Value of the system property that enables threads dumping on failure. */
-    private static final boolean IGNITE_DUMP_THREADS_ON_FAILURE =
-        IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_DUMP_THREADS_ON_FAILURE, true);
+    private final boolean igniteDumpThreadsOnFailure =
+        IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_DUMP_THREADS_ON_FAILURE, false);
+
+    /** Ignored failure log message. */
+    static final String IGNORED_FAILURE_LOG_MSG = "Possible failure suppressed accordingly to a configured handler ";
+
+    /** Failure log message. */
+    static final String FAILURE_LOG_MSG = "Critical system error detected. " +
+        "Will be handled accordingly to configured handler ";
 
     /** Ignite. */
     private final Ignite ignite;
@@ -124,8 +132,14 @@ public class FailureProcessor extends GridProcessorAdapter {
         if (this.failureCtx != null) // Node already terminating, no reason to process more errors.
             return false;
 
-        U.error(ignite.log(), "Critical system error detected. Will be handled accordingly to configured handler " +
-            "[hnd=" + hnd + ", failureCtx=" + failureCtx + ']', failureCtx.error());
+        if (failureTypeIgnored(failureCtx, hnd)) {
+            U.quietAndWarn(ignite.log(), IGNORED_FAILURE_LOG_MSG +
+                "[hnd=" + hnd + ", failureCtx=" + failureCtx + ']', failureCtx.error());
+        }
+        else {
+            U.error(ignite.log(), FAILURE_LOG_MSG +
+                "[hnd=" + hnd + ", failureCtx=" + failureCtx + ']', failureCtx.error());
+        }
 
         if (reserveBuf != null && X.hasCause(failureCtx.error(), OutOfMemoryError.class))
             reserveBuf = null;
@@ -137,8 +151,8 @@ public class FailureProcessor extends GridProcessorAdapter {
                 " WAL path: " + ctx.config().getDataStorageConfiguration().getWalPath() +
                 " WAL archive path: " + ctx.config().getDataStorageConfiguration().getWalArchivePath());
 
-        if (IGNITE_DUMP_THREADS_ON_FAILURE)
-            U.dumpThreads(log);
+        if (igniteDumpThreadsOnFailure)
+            U.dumpThreads(log, !failureTypeIgnored(failureCtx, hnd));
 
         DiagnosticProcessor diagnosticProcessor = ctx.diagnostic();
 
@@ -154,5 +168,14 @@ public class FailureProcessor extends GridProcessorAdapter {
         }
 
         return invalidated;
+    }
+
+    /**
+     * @param failureCtx Failure context.
+     * @param hnd Handler.
+     */
+    private static boolean failureTypeIgnored(FailureContext failureCtx, FailureHandler hnd) {
+        return hnd instanceof AbstractFailureHandler &&
+            ((AbstractFailureHandler)hnd).getIgnoredFailureTypes().contains(failureCtx.type());
     }
 }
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
```diff
@@ -39,6 +40,12 @@
     private static final boolean IGNITE_DUMP_THREADS_ON_FAILURE =
         IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_DUMP_THREADS_ON_FAILURE, true);
 
+    /** Ingonred failure log message. */
+    public static final String IGNORED_FAILURE_LOG_MESSAGE = "Possible failure suppressed accordingly to a configured handler ";
```
agura: It seems that package local is enough for both constants.

agura: Use `MSG` instead of `MESSAGE` for both constants accordingly to abbreviation rules.

agura: Add empty line between constants.

modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
```diff
@@ -124,8 +131,14 @@ public synchronized boolean process(FailureContext failureCtx, FailureHandler hn
         if (this.failureCtx != null) // Node already terminating, no reason to process more errors.
             return false;
 
-        U.error(ignite.log(), "Critical system error detected. Will be handled accordingly to configured handler " +
-            "[hnd=" + hnd + ", failureCtx=" + failureCtx + ']', failureCtx.error());
+        if (isFailureIgnored(failureCtx, hnd)) {
```
agura: We can ignore failure type not failure. Please rename method. Also accordingly to code style you should use GG-style getter (no `get` or `is` in method names).

modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
```diff
@@ -137,8 +150,12 @@ public synchronized boolean process(FailureContext failureCtx, FailureHandler hn
                 " WAL path: " + ctx.config().getDataStorageConfiguration().getWalPath() +
                 " WAL archive path: " + ctx.config().getDataStorageConfiguration().getWalArchivePath());
 
-        if (IGNITE_DUMP_THREADS_ON_FAILURE)
-            U.dumpThreads(log);
+        if (IGNITE_DUMP_THREADS_ON_FAILURE) {
+            if (isFailureIgnored(failureCtx, hnd))
+                U.dumpThreads(log, false);
```
agura: It could be rewritten in one line: `U.dumpThreads(log, !isFailureIgnored(failureCtx, hnd));`

modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
```diff
@@ -155,4 +172,13 @@ public synchronized boolean process(FailureContext failureCtx, FailureHandler hn
 
         return invalidated;
     }
+
+    /**
+     * @param failureCtx Failure context.
+     * @param hnd        Handler.
```
agura: Don't use indent here. Just one space.

modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
```diff
@@ -155,4 +172,13 @@ public synchronized boolean process(FailureContext failureCtx, FailureHandler hn
...
+     */
+    private boolean isFailureIgnored(FailureContext failureCtx, FailureHandler hnd) {
```
agura: Method could be static.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1416,20 +1416,32 @@ private static double roundedHeapSize(double heap, int precision) {
...
+
+    /**
+     * Performs thread dump and prints all available info to the given log
+     * with WARN or ERROR logging level depending on {@code isErrorLevel} parameter.
+     *
+     * @param log          Logger.
+     * @param isErrorLevel {@code true} if thread dump must be printed with ERROR logging level,
+     *                     {@code false} if thread dump must be printed with WARN logging level.
+     */
+    public static void dumpThreads(@Nullable IgniteLogger log, boolean isErrorLevel) {
```
agura: No `is` for vsriable name and use `err` instead of `error` accordingly to abbreviation rules.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1416,20 +1416,32 @@ private static double roundedHeapSize(double heap, int precision) {
...
      */
     public static void dumpThreads(@Nullable IgniteLogger log) {
+        dumpThreads(log, false);
+    }
+
+    /**
+     * Performs thread dump and prints all available info to the given log
+     * with WARN or ERROR logging level depending on {@code isErrorLevel} parameter.
+     *
+     * @param log          Logger.
```
agura: Don't use indents in javadocs.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1451,7 +1463,20 @@ public static void dumpThreads(@Nullable IgniteLogger log) {
 
         sb.a(NL);
 
-        warn(log, sb.toString());
+        logMessage(log, sb.toString(), isErrorLevel);
+    }
+
+    /**
+     * @param log          Logger.
```
agura: Son't use indets in javadocs.

modules/core/src/test/java/org/apache/ignite/failure/FailureHandlerLoggingTest.java
```diff
@@ -0,0 +1,85 @@
...
+     */
+    private final ListeningTestLogger log = new ListeningTestLogger(false, GridCommonAbstractTest.log);
+
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        super.afterTest();
+
+        log.clearListeners();
+
+        stopAllGrids();
```
agura: You can use the same Ignite instance for both test cases. Just start node before all tests and stop it after all tests. Starting of node is time consuming operation.

modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
alievmirza: seems that is not enough
`org.apache.ignite.internal.processors.failure.FailureProcessor`
`org.apache.ignite.failure.FailureHandlerLoggingTest`

modules/core/src/main/java/org/apache/ignite/internal/processors/failure/FailureProcessor.java
```diff
@@ -39,6 +40,13 @@
     private static final boolean IGNITE_DUMP_THREADS_ON_FAILURE =
         IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_DUMP_THREADS_ON_FAILURE, true);
 
+    /** Ignored failure log message. */
+    public static final String IGNORED_FAILURE_LOG_MSG = "Possible failure suppressed accordingly to a configured handler ";
```
agura: Once again. Make both constants  package-local and move your test class to the same package.

modules/core/src/test/java/org/apache/ignite/failure/FailureHandlerLoggingTest.java
```diff
@@ -0,0 +1,104 @@
...
+        super.afterTest();
+
+        TEST_LOG.clearListeners();
+    }
+
+    /**
+     * Tests log message for ingnored failure types.
+     */
+    @Test
+    public void testFailureHandlerLoggedIgnoredFailureTest() throws Exception {
```
agura: Now you have the same code for both methods. Only two differences: failure type and exepcted messages. Could be moved to one method.

modules/core/src/test/java/org/apache/ignite/failure/FailureHandlerLoggingTest.java
```diff
@@ -0,0 +1,104 @@
...
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static org.apache.ignite.internal.processors.failure.FailureProcessor.FAILURE_LOG_MSG;
+import static org.apache.ignite.internal.processors.failure.FailureProcessor.IGNORED_FAILURE_LOG_MSG;
+
+/**
+ * Tests log messages of failure handler.
+ */
+public class FailureHandlerLoggingTest extends GridCommonAbstractTest {
```
agura: Actually it is failure processor test. Please rename and move to the same package as `FailureProcessor` placed.

modules/core/src/test/java/org/apache/ignite/failure/FailureHandlerLoggingTest.java
alievmirza: According to test plan there should be two different tests https://ggsystems.atlassian.net/browse/GG-25912

PR #735, opened 2019-11-21
GG-25913: Print warring message if tracing feature is not enabled.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/ManagementConsoleProcessor.java
```diff
@@ -132,9 +134,13 @@ public ManagementConsoleProcessor(GridKernalContext ctx) {
     @Override public void onKernalStart(boolean active) {
         this.metaStorage = ctx.distributedMetastorage();
         this.evtsExporter = new EventsExporter(ctx);
-        this.spanExporter = new SpanExporter(ctx);
         this.metricExporter = new MetricsExporter(ctx);
 
+        if (isTracingEnabled())
+            this.spanExporter = new SpanExporter(ctx);
+        else
+            U.quietAndWarn(log, "The trace feature is disabled because not all nodes support this");
```
akuznetsov-gridgain:
```suggestion
            U.quietAndWarn(log, "Current Ignite configuration does not support tracing functionality " +
            "and management console agent will not collect traces " +
            "(consider adding ignite-opencensus module to classpath).");
```

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ManagementConsoleProcessor.java
```diff
@@ -216,6 +222,13 @@ public WebSocketManager webSocketManager() {
         return mgr;
     }
 
+    /**
+     * @return {@code true} if tracing is enable.
```
nva:
```suggestion
     * @return {@code True} if tracing is enable.
```

modules/management-console-agent/src/test/java/org/apache/ignite/agent/testsuites/AgentTestSuite.java
```diff
@@ -63,7 +64,8 @@
     EventsExporterTest.class,
     RetryableSenderTest.class,
     RequestDeserializerTest.class,
-    TopologySnapshotTest.class
+    TopologySnapshotTest.class,
```
akuznetsov-gridgain: Lets sort tests as we agreed - by the same order as in "imports".

PR #737, opened 2019-11-22
GG-25549 .NET: Add NuGet verification script

ptupitsyn:
Add a script to verify NuGet packages:
* Creates a test project with `dotnet new console`
* Installs all packages from specified folder with `dotnet add package`
* Runs a test program that uses basic Cache, SQL and LINQ APIs
* Checks exit codes to make sure there were no exceptions/errors

### DIFF
--- a/modules/platforms/dotnet/release/Program.cs
+++ b/modules/platforms/dotnet/release/Program.cs
@@ -1,0 +1,77 @@
+﻿using System;
+using System.Diagnostics;
+using System.Linq;
+using System.Threading.Tasks;
+using Apache.Ignite.Core;
+using Apache.Ignite.Core.Cache.Configuration;
+using Apache.Ignite.Core.Client;
+using Apache.Ignite.Core.Discovery.Tcp;
+using Apache.Ignite.Core.Discovery.Tcp.Static;
+using Apache.Ignite.Linq;
+using Apache.Ignite.NLog;
+
+namespace test_proj
+{
+    public static class Program
+    {
+        public static async Task Main()
+        {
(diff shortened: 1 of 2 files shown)


modules/platforms/dotnet/release/verify-nuget.ps1
```diff
@@ -0,0 +1,85 @@
...
+
+#>
+
+param (
+    [string]$packageDir="..\nupkg"
+)
+
+
+# Find NuGet packages (*.nupkg)
+$dir = Join-Path $PSScriptRoot $packageDir
```
isapego: Does it only work with relative paths?

modules/platforms/dotnet/release/Program.cs
```diff
@@ -0,0 +1,99 @@
...
+            }
+        }
+    }
+
+    public class Person
+    {
+        public Person(int age)
+        {
+            Age = age;
+            Name = $"Person-{age}";
```
isapego: Wow, didn't know .NET can do that.

ptupitsyn: Killer feature, right? 😁 
There is lots of impressive stuff in recent versions, but we are stuck with 4.0 in the main code base for now :(
This script uses modern tooling (.NET Core), so we can use newer language features.

modules/platforms/dotnet/release/verify-nuget.ps1
ptupitsyn: Good point, fixed

PR #744, opened 2019-11-26
GG-25993: Fix issue with trailing slashed in monitoring url.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
```diff
@@ -64,7 +64,7 @@ private AgentUtils() {
      * @param clusterId Cluster ID.
      */
     public static String monitoringUri(String srvUri, UUID clusterId) {
-        return srvUri + "/clusters/" + clusterId + "/monitoring-dashboard";
+        return srvUri.replaceAll("/+$", "") + "/clusters/" + clusterId + "/monitoring-dashboard";
```
akuznetsov-gridgain: Could we have utility function that will be the same signature as `java.nio.file.Paths#get(java.lang.String, java.lang.String...)` ?
For example: `Utility.buildUri(String first, Object... more)` ? And hide all logic for `\` manipulation in utility function?
So usage will be: `buildUri(srvUri, "clusters", clusterId,  "monitoring-dashboard");`

nva: Try something like this or search in spring utils.
```
URI uri = URI.create(srvUri);

try {
    uri = new URI("http".equalsIgnoreCase(uri.getScheme()) ? "ws" : "wss",
        uri.getUserInfo(),
        uri.getHost(),
        uri.getPort(),
        "/clusters/" + clusterId + "/monitoring-dashboard",
        null,
        null
    );
}
catch (URISyntaxException x) {
    throw new IllegalArgumentException(x.getMessage(), x);
}
```

smoldachev: Replace on StringUtils#trimTrailingCharacter

modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
```diff
@@ -64,7 +65,7 @@ private AgentUtils() {
      * @param clusterId Cluster ID.
      */
     public static String monitoringUri(String srvUri, UUID clusterId) {
-        return srvUri + "/clusters/" + clusterId + "/monitoring-dashboard";
+        return URI.create(srvUri + "/clusters/" + clusterId + "/monitoring-dashboard").normalize().toString();
```
akuznetsov-gridgain: Lets create utility method for this?

smoldachev: Utility method for utility method? :) I think this method will only be used to monitoring URI.

PR #745, opened 2019-11-26
GG-23862 Sorted configuration properties. Fixed showing of not configured object.

vsisko:


modules/web-console/e2e/testcafe/page-models/pageSignin.js
```diff
@@ -24,6 +25,7 @@ export const pageSignin = {
     selector: Selector('page-signin'),
     async login(email, password) {
         return await t
+            .expect(activeLoadingOverlay.exists).notOk()
```
Klaster1: How is this related to the task?

vsisko: In time of test development I notice that opening of Sign In page is longer than waiting time for input field.

PR #746, opened 2019-11-26
GG-25962  Fix issues related to client cache stop and SQL metadata retrieval

tledkov-gridgain:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1788,6 +1788,15 @@ public UpdateResult executeUpdateOnDataNode(
         return null;
     }
 
+    /** {@inheritDoc} */
+    @Override public void clearCacheInfo(String cacheName) {
+        GridCacheContextInfo cacheInfo = registeredCacheInfo(cacheName);
```
ygerzhedovich: may be need to add assertion here to check cacheInfo is not null

PR #749, opened 2019-11-27
GG-25416 Add more info to "known metrics" backend endpoints.

nva:


modules/management-console-agent/src/test/resources/ignite-cfg.xml
```diff
@@ -41,6 +41,14 @@
             </bean>
         </property>
 
+        <property name="cacheConfiguration">
```
akuznetsov-gridgain: Why not simply use `ignite.getOrCreateCache("test-cache")` ?

Klaster1:
@akuznetsov-gridgain what exactly you expect me to review?

modules/management-console-agent/src/main/java/org/apache/ignite/agent/StompDestinationsUtils.java
```diff
@@ -76,8 +76,8 @@ public static String buildSaveSpanDest(UUID clusterId) {
     /**
      * @return Metrics destination.
      */
-    public static String buildMetricsDest(UUID clusterId) {
-        return METRICS_DEST + clusterId + "/add";
+    public static String buildMetricsDest() {
+        return METRICS_DEST;
```
akuznetsov-gridgain: @Klaster1 It is OK from fronrend side that `clusterId` not used any more in endpoint name?

Klaster1: You mean all clusters will share the same metric definitions? That sounds wrong and totally not what we discussed before.

modules/management-console-agent/src/test/resources/ignite-cfg.xml
smoldachev: Because we use this configuration from CommandLineLauncher and we can't use getOrCreate :(

modules/management-console-agent/src/main/java/org/apache/ignite/agent/StompDestinationsUtils.java
smoldachev: @akuznetsov-gridgain Andrey remove cluster id from destination, because MetricResponse has cluterId in body.

nva: @akuznetsov-gridgain @Klaster1 This endpoint on backend where agent send metrics from nodes. This metrics package has binary format and already contains clusterId.

Klaster1: Then I have no complaints.

PR #754, opened 2019-12-02
GG-26127: Fix GMC agent sql metadata export.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/cache/CacheSqlMetadata.java
```diff
@@ -208,6 +209,30 @@ public CacheSqlMetadata setIndexes(List<CacheSqlIndexMetadata> indexes) {
         return this;
     }
 
+    /** {@inheritDoc} */
+    @Override public boolean equals(Object o) {
+        if (this == o) return true;
```
akuznetsov-gridgain: return should be on next line by code style.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/CacheChangesProcessor.java
```diff
@@ -126,10 +127,10 @@ private void sendCacheInfo() {
     private Collection<CacheSqlMetadata> getCacheSqlMetadata() {
         GridCacheProcessor cacheProc = ctx.cache();
 
-        List<CacheSqlMetadata> cachesMetadata = new ArrayList<>();
+        Collection<CacheSqlMetadata> cachesMetadata = new HashSet<>();
 
         for (Map.Entry<String, DynamicCacheDescriptor> item : cacheProc.cacheDescriptors().entrySet()) {
-            if (item.getValue().sql()) {
+            if (!item.getValue().schema().entities().isEmpty() && !isSystemCache(item.getKey())) {
```
akuznetsov-gridgain: I think that we can add "sql" flag to CacheSqlMetadata. This flag can be used on UI to group SQL table or add some visual sign that table was created via SQL, for example.

smoldachev: Add createBySql parameter.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/cache/CacheSqlMetadata.java
```diff
@@ -208,6 +209,32 @@ public CacheSqlMetadata setIndexes(List<CacheSqlIndexMetadata> indexes) {
...
+
+        CacheSqlMetadata that = (CacheSqlMetadata) o;
+
+        return cacheName.equals(that.cacheName) &&
+            schemaName.equals(that.schemaName) &&
+            tblName.equals(that.tblName) &&
+            typeName.equals(that.typeName) &&
+            Objects.equals(keyCls, that.keyCls) &&
+            Objects.equals(valCls, that.valCls) &&
+            Objects.equals(fields, that.fields) &&
```
akuznetsov-gridgain: `createdBySql` is not checked in `equals` and `hashCode` - is this expected?

modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/cache/CacheInfo.java
```diff
@@ -29,6 +29,9 @@
     /** Deployment id. */
     private IgniteUuid deploymentId;
 
+    /** Created by sql. */
+    private boolean createdBySql;
```
nva: Why this flag has different name in compare to ignite-core?

modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/cache/CacheSqlMetadata.java
smoldachev: `createdBySql` available only in `CacheInfo`, not in `CacheSqlMetadata`

modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/cache/CacheInfo.java
smoldachev: The core name will only confuses us, because initially I thought that sql flag mean availability of sql queries for this cache.

PR #756, opened 2019-12-03
GG-25908: Client node stopped with AssertionError in GridReduceQueryExecutor.

AMashenkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -400,11 +400,14 @@ else if (msg.page() == 0) // Count down only on each first page received.
             }
 
             if (attempt > 0 && retryTimeout > 0 && (U.currentTimeMillis() - startTime > retryTimeout)) {
+                //TODO: GG-23176: Retry logic looks too complicated.
```
mshonichev: GG-26176?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
```diff
@@ -466,11 +469,9 @@ else if (msg.page() == 0) // Count down only on each first page received.
                 partsMap = nodesParts.partitionsMap();
                 qryMap = nodesParts.queryPartitionsMap();
 
-                if (nodes == null)
+                if (F.isEmpty(nodes))
```
mshonichev: Seems that similar condition `(nodes == null)` , where `nodes` comes from `nodesForPartitions.nodes()` is used in the `update()` method here. Should we replace it with `F.isEmpty()` too (and add corresponding test)?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java
AMashenkov: Yes

PR #757, opened 2019-12-03
GG-25851 Concurrent removeAll() on the same cache leads to deadlock

alievmirza:


modules/core/src/test/java/org/apache/ignite/cache/RemoveAllDeadlockTest.java
```diff
@@ -0,0 +1,109 @@
...
+import java.util.Set;
+import java.util.TreeSet;
+import org.apache.ignite.IgniteCache;
+import org.apache.ignite.configuration.CacheConfiguration;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/**
+ *
```
glukos: Let's describe what's being tested here.

modules/core/src/test/java/org/apache/ignite/cache/RemoveAllDeadlockTest.java
```diff
@@ -0,0 +1,109 @@
...
+ */
+public class RemoveAllDeadlockTest extends GridCommonAbstractTest {
+    /** Threads number for reporducing deadlock. */
+    public static final int THREADS = 4;
+
+    /**
+     * {@inheritDoc}
```
glukos: /** {@inheritDoc} */ is usually placed in our code as one-liner

modules/core/src/test/java/org/apache/ignite/cache/RemoveAllDeadlockTest.java
```diff
@@ -0,0 +1,109 @@
...
+    /** {@inheritDoc} */
+    @Override protected void afterTest() throws Exception {
+        stopAllGrids();
+    }
+
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testTreeSetRemoveAll() throws Exception {
```
glukos: What's the point of running this test nightly?
It was never meant to be deadlocked.

PR #761, opened 2019-12-05
GG-26275: Add new constructor for MetricResponse

smoldachev:


modules/management-console-agent/src/test/java/org/apache/ignite/agent/processor/export/MetricsExporterTest.java
```diff
@@ -72,6 +72,61 @@ public void testResponseWithoutUserTag() {
...
+                if (m instanceof BooleanMetric)
+                    type = 0;
+                else if (m instanceof IntMetric)
+                    type = 1;
+                else if (m instanceof LongMetric)
+                    type = 2;
+                else if (m instanceof DoubleMetric)
+                    type = 3;
+                else
+                    throw new IllegalArgumentException("Unknown metric type.");
```
MaxLosevskoy: Is there a chance that new metric types will be introduced in future?
If yes, the test will be failing - is it better than storing types in a kind of enum?

Just asking :)

PR #767, opened 2019-12-09
GG-25278 Add default query timeout for SQL Queries

pavlukhin:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1590,6 +1589,7 @@ private void checkSecurity(Collection<Integer> cacheIds) {
             .setEnforceJoinOrder(fldsQry.isEnforceJoinOrder())
             .setLocal(fldsQry.isLocal())
             .setPageSize(fldsQry.getPageSize())
+            // t0d0 add a comment
```
AMashenkov: is it TODO?

PR #769, opened 2019-12-10
GG-26266

sanpwc:
Initial implementation of logic that repairs doubtful keys.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/tasks/RepairRequestTask.java
```diff
@@ -458,6 +479,9 @@ public RepairEntryProcessor(
 
                 GridCacheContext cctx = (GridCacheContext)entry.unwrap(GridCacheContext.class);
 
+                if (!cctx.affinity().affinityTopologyVersion().equals(startTopVer))
```
glukos: What if client topology event has happened?
Comparing cctx.exchange().lastAffinityChangedTopologyVersion(<currentTopologyVersion>) with startTopVer should help.

sanpwc: Fixed.

PR #770, opened 2019-12-10
GG-24254

korlov42:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -233,28 +253,30 @@ protected H2Tree(
...
-            initTree(initNew, inlineSize);
+                H2ExtrasLeafIO.getVersions(inlineSize, mvccEnabled)
+            );
         }
 
-        cols = unwrappedPk ? unwrappedColsInfo.cols() : wrappedColsInfo.cols();
-
         created = initNew;
+
+        initTree(created, inlineSize);
```
ygerzhedovich: I think more clearly will be keep the imvocation inside ``else`` block with ``true`` for first argument

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -472,7 +504,7 @@ private void upgradeMetaPage(boolean inlineObjSupported) throws IgniteCheckedExc
                 lastIdxUsed++;
 
                 if (c != 0)
-                    return c;
+                    return fixSort(c, col.sortType);
```
ygerzhedovich: it could be breaking change from compatibility perspective

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -125,34 +82,76 @@ protected H2TreeIndexBase(GridH2Table tbl, String name, IndexColumn[] cols, Inde
...
             return 0;
 
-        if (cfgInlineSize == -1) {
-            if (propSize == 0)
-                return 0;
+        if (cfgInlineSize != -1)
+            return Math.min(PageIO.MAX_PAYLOAD_SIZE, cfgInlineSize);
 
-            int size = 0;
+        int propSize = maxInlineSize == -1 ? IgniteSystemProperties.getInteger(IgniteSystemProperties.IGNITE_MAX_INDEX_PAYLOAD_SIZE,
```
AMashenkov: getInteger() is synchronous HashTable operation.
Let's make is static.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -125,34 +82,76 @@ protected H2TreeIndexBase(GridH2Table tbl, String name, IndexColumn[] cols, Inde
...
+     * @param tbl Table.
+     * @param cols Columns.
+     * @param factory Factory.
+     * @param inlineObjHashSupported Whether hash inlining is supported or not.
+     * @return List of {@link InlineIndexColumn} objects.
+     */
+    static List<InlineIndexColumn> getAvailableInlineColumns(boolean affinityKey, String cacheName,
+        String idxName, IgniteLogger log, boolean pk, Table tbl, IndexColumn[] cols,
+        InlineIndexColumnFactory factory, boolean inlineObjHashSupported) {
+        List<InlineIndexColumn> res = new ArrayList<>();
```
AMashenkov: Result array size is known here.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
ygerzhedovich: I suppose it could break a tests, which change the property. The method invoke so seldom, what shouldn't affect a performance

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/inlinecolumn/InlineIndexColumnFactory.java
```diff
@@ -0,0 +1,236 @@
...
+     * @param inlineObjHash Inline object hash.
+     */
+    public InlineIndexColumn createInlineHelper(Column col, boolean inlineObjHash) {
+        int type = col.getType().getValueType();
+
+        if (type == Value.JAVA_OBJECT && !inlineObjHash)
+            // fallback to old inlining logic
+            return new ObjectBytesInlineIndexColumn(col, mode.isBinaryUnsigned());
+
+        return helpers.computeIfAbsent(col.getName(), name -> createHelper(mode, col));
```
ygerzhedovich: Is there any place where creating inlineHelper could be concurrent? I haven't found such place.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
korlov42: ok, I'll move it

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
korlov42: not, it did not breaks compatibility, cause decision about which way to choose (left or right) is taken on the upper level (in BPlusTree). By the way I added compatibility test to check such behaviour

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: agree with Yuri. I would prefer not to use static constant as it brings some troubles to testing.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: ok, add initial capacity

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/inlinecolumn/InlineIndexColumnFactory.java
korlov42: no, it currently used in synchronous way. I'll change type of the map to HashMap so as not to be misleading

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/InlineIndexColumn.java
```diff
@@ -0,0 +1,100 @@
...
+     * @param pageAddr Page address.
+     * @param off Offset.
+     * @param maxSize Max size.
+     * @param v Value that should be compare.
+     * @param comp Comparator.
+     *
+     * @return -1, 0 or 1 if inlined value less, equal or greater
+     * than given respectively, or -2 if inlined part is not enough to compare.
+     */
+    public int compare(long pageAddr, int off, int maxSize, Value v, Comparator<Value> comp);
```
pavlukhin: Premature optimization concern: we introduce class hierarchy here and megamorphic virtual calls might be significantly slower than _switch-case_. Running some (micro) benchmarks might be a good idea.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/inlinecolumn/InlineIndexColumnFactory.java
```diff
@@ -0,0 +1,237 @@
...
+import org.apache.ignite.internal.util.lang.GridTuple;
+import org.h2.table.Column;
+import org.h2.value.CompareMode;
+import org.h2.value.Value;
+
+/**
+ * Factory to create inline column wrapper for given column.
+ */
+public class InlineIndexColumnFactory {
+    /** PageContext for use in IO's */
```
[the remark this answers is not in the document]
korlov42: Yes, it is context for current page that should be written along with original entry.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/inlinecolumn/ObjectHashInlineIndexColumn.java
```diff
@@ -0,0 +1,80 @@
...
+        super(col, Value.JAVA_OBJECT, (short)4);
+    }
+
+    /** {@inheritDoc} */
+    @Override protected int compare0(long pageAddr, int off, Value v, int type) {
+        if (type() != type)
+            return COMPARE_UNSUPPORTED;
+
+        int val1 = PageUtils.getInt(pageAddr, off + 1);
+        int val2 = v.getObject().hashCode();
```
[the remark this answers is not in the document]
korlov42: actually only BinaryObject is allowed here, but currently it is not true :(. I added the check in code and created a ticket https://ggsystems.atlassian.net/browse/GG-26697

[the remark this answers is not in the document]
korlov42: I was totally wrong. The object may (and should) belongs to the one of the following type: any core JVM type (like primitive type, Date, UUID, etc) or BinaryObject for any custom type. 

If it is not true, then we have problem not with inlining but with value comparing in general.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/InlineIndexColumn.java
korlov42: The benchmark shows that after refactoring there is no decrease in performance. You can see the result in the [Jira ticket](https://ggsystems.atlassian.net/browse/GG-24254).

PR #784, opened 2019-12-13
GG-20230 Fixing error when concurrently releasing preloaded history f…

alievmirza:
…or WAL


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -423,6 +424,9 @@
     /** Pointer to a memory recovery record that should be included into the next checkpoint record. */
     private volatile WALPointer memoryRecoveryRecordPtr;
 
+    /** Lock for history preloading. */
+    private ReentrantLock lockForHistoryPreloading = new ReentrantLock();
```
glukos: Lock is actually used for releasing only. Let's mention it in variable name.

PR #785, opened 2019-12-13
GG-26176: SQL: Simplify GridReduceQueryExecutor.query() method.

AMashenkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2JavaObjectSerializer.java
```diff
@@ -0,0 +1,54 @@
...
+
+    /** Marshaller. */
+    private final Marshaller marshaller;
+
+    /**
+     * Constructor.
+     *
+     * @param ctx Kernal context.
+     */
+    H2JavaObjectSerializer(@NotNull GridKernalContext ctx) {
```
ygerzhedovich: may be better to have Marshaller as parameter rather GridKernalContext?

PR #792, opened 2019-12-16
GG-26416: Add node order field in topology snapshot.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/topology/Node.java
```diff
@@ -141,6 +147,23 @@ public void setConsistentId(String consistentId) {
         this.consistentId = consistentId;
     }
 
+    /**
+     * @return Order.
```
akuznetsov-gridgain:
```suggestion
     * @return Node order. Will return {@code -1} for baseline node.
```

PR #797, opened 2019-12-17
GG-21236

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/QueryCursorImpl.java
```diff
@@ -142,14 +152,37 @@ public QueryCursorImpl(Iterable<T> iterExec, GridQueryCancel cancel, boolean isQ
...
+                    STATE_UPDATER.set(this, NO_DATA);
+            }
+            catch (Exception e) {
+                // Ignore exception on check iterator
+                // because Iterator.hasNext() may throw error on invalid / error query.
+                STATE_UPDATER.set(this, NO_DATA);
+            }
+
+            if (STATE_UPDATER.compareAndSet(this, NO_DATA, CLOSED)) {
+                closeIter();
```
ygerzhedovich: potentially Iterator can be closed twice in case we have two concurrent ``close`` invocations here, because earlier you  set ``NO_DATA`` without any doubts

PR #800, opened 2019-12-18
GG-25501 Latch ack message listen on future if it hasn't created yet

akalash:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/latch/ExchangeLatchManager.java
```diff
@@ -99,6 +101,13 @@
     @GridToStringExclude
     private final ConcurrentMap<AffinityTopologyVersion, ClusterNode> joinedNodes = new ConcurrentHashMap<>();
 
+    /** This map holds version on which ack message have been received before local latch was created. */
+    private final ConcurrentSkipListMap<AffinityTopologyVersion, GridFutureAdapter<Object>> awaitedLatchVersions =
```
glukos: ExchangeLatchManager can handle several latches for the same topologyVersion, see CompletableLatchUid.
We need to handle correctly all of them.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/latch/ExchangeLatchManager.java
```diff
@@ -362,6 +395,38 @@ public boolean canSkipJoiningNodes(AffinityTopologyVersion topVer) {
...
+     *
+     * @param from Node sent ack.
+     * @param message Ack message.
+     */
+    private void processAck(UUID from, LatchAckMessage message) {
+        lock.lock();
+
+        try {
+            if (!isLocalLatchCreated(message.topVer())) {
+                awaitedLatchVersions.computeIfAbsent(message.topVer(), (k) -> new GridFutureAdapter<>())
```
glukos: The same issue with several latches here. We should take LatchAckMessage#latchId into account.
We can write a test with second latch using PartitionsExchangeAware mechanism.

PR #802, opened 2019-12-18
GG-26007 Improvement of human-readable format of WAL records

a-polyakov:


modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
```diff
@@ -35,4 +41,86 @@
...
+        String keyStr;
+        if (key instanceof String) {
+            keyStr = (String)key;
+        }
+        else if (key instanceof BinaryObject) {
+            keyStr = key.toString();
+        }
+        else {
+            keyStr = (key != null) ? toStringRecursive(key.getClass(), key) : null;
+        }
```
denis-chudov: We don't use braces for one-line blocks of code.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
```diff
@@ -35,4 +41,86 @@
...
+        String valueStr;
+        if (value instanceof String) {
+            valueStr = (String)value;
+        }
+        else
+        if (value instanceof BinaryObject) {
+            valueStr = value.toString();
+        }
+        else {
+            valueStr = (value != null) ? toStringRecursive(value.getClass(), value) : null;
```
denis-chudov: We don't use braces for one-line blocks of code.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
```diff
@@ -35,4 +41,86 @@
...
+     *
+     * @param cls Declaration class of the object.
+     * @param obj Object to get a string presentation for.
+     * @return String presentation of the given object.
+     */
+    public static String toStringRecursive(Class cls, Object obj) {
+        String result = null;
+        if (cls != Object.class) {
+            result = S.toString(cls, obj, toStringRecursive(cls.getSuperclass(), obj));
+        }
```
denis-chudov: You can simplify this using conditional operator.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
```diff
@@ -35,4 +41,86 @@
...
+            try {
+                valueStr = Base64.getEncoder().encodeToString(entry.value().valueBytes(cacheObjValCtx));
+            }
+            catch (IgniteCheckedException e) {
+                cacheObjValCtx.kernalContext().log(UnwrapDataEntry.class)
+                    .error("Unable to convert value [" + entry.value() + "]", e);
+            }
+        }
+
+        return entry.getClass().getSimpleName() + "[k = " + keyStr + ", v = ["
```
denis-chudov: I think we should add flag to be able to hide sensitive data in WAL reader output. For example, we can use cryptographic hash to hide data and see some representation of it to know is it unique or not, with high probability.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -81,33 +148,70 @@ public static void main(String[] args) throws Exception {
...
+                c = c.getSuperclass();
+            }
+        }
+        return result;
+    }
 
-        if (stat != null)
-            System.out.println("Statistic collected:\n" + stat.toString());
+    /**
+     * Get a field Value in object by filed name
```
denis-chudov: typo (filed -> field)

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -81,33 +148,70 @@ public static void main(String[] args) throws Exception {
...
+     *
+     * @param o         Object.
+     * @param fieldName Field name.
+     * @return Filed value.
```
denis-chudov: typo

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -81,33 +148,70 @@ public static void main(String[] args) throws Exception {
...
+     * @throws IllegalAccessException
```
denis-chudov: please add some description (e.g. `@throws IllegalAccessException If failed.` )

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
```diff
@@ -0,0 +1,381 @@
...
+ *             <li>Put several entity</li>
+ *             <li>Stop node</li>
+ *             <li>Change byte in WAL</li>
+ *             <li>Read wal</li>
+ *             <li>Check one error when reading WAL</li>
+ *         </ul>
+ *     </li>
+ * </ol>
+ */
+public class IgniteWalConverterTest extends GridCommonAbstractTest {
```
denis-chudov: Please add some tests for incorrect command line input parameters. We should be sure that all parameters are checked properly, and user gets informative error message in case of mistake or typo.
Also, it's usually not necessary to do such extended description in javadoc of test class, you can describe each test case only in method javadoc.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
a-polyakov: No need? because if not set binary metadata object fields will not be displayed (parameter is implemented within different issue), only its binary representation

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
a-polyakov: Checks will be added to https://ggsystems.atlassian.net/browse/GG-26099

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
```diff
@@ -0,0 +1,381 @@
...
+     *    <li>Start node</li>
+     *    <li>Create cache with <a href="https://apacheignite.readme.io/docs/indexes#section-registering-indexed-types">Registering Indexed Types</a></li>
+     *    <li>Put several entity</li>
+     * </ul>
+     *
+     * @param list Returns entities that have been added.
+     * @return Node folder name.
+     * @throws Exception
+     */
+    private String createWal(List<Person> list) throws Exception {
```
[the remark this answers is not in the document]
a-polyakov: You have to duplicate a lot of code in the static method (startGrid, getConfiguration ...)

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
```diff
@@ -0,0 +1,381 @@
...
+                            if (index > 0) {
+                                index = result.indexOf("name=" + person.getName(), index + 4);
+                                find = index > 0;
+                            }
+                        }
+                    }
+                }
+            }
+            if (!find) {
+                countErrorRead++;
```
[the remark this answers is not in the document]
a-polyakov: System.setProperty(IgniteSystemProperties.IGNITE_PDS_SKIP_CRC, "true");
RecordV1Serializer.skipCrc = true;
Disables the crc check.
The test is made so that spoils value in the interior DataRecord and does not find a match to the added data (countErrorRead == 1)

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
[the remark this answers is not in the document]
denis-chudov: You can make an abstract class with common logic (such as getConfiguration) and child classes, one with `createWal` called in `beforeTestsStarted`, and one with tests that run grids for themselves. Anyway, it's important to reduce tests' running time, and please also keep in mind that most likely you will have to add more tests to implement the test plan.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
denis-chudov: Security policies don't allow us to receive WAL reader output even when there is only binary representation, because there is theoretical possibility to decode data. If we will have only hashes instead of data in any representation, there will be no need to deny us from receiving WAL reader output.

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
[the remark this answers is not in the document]
denis-chudov: Yes, I got it, but I didn't see any difference in output between correct and incorrect records. There should be a way to distinguish them.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
```diff
@@ -35,4 +41,86 @@
...
+            keyStr = (String)key;
+        }
+        else if (key instanceof BinaryObject) {
+            keyStr = key.toString();
+        }
+        else {
+            keyStr = (key != null) ? toStringRecursive(key.getClass(), key) : null;
+        }
+
+        if (keyStr == null || keyStr.isEmpty()) {
```
denis-chudov: Here and below (for valueStr) you could use `F.isEmpty()`

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -45,34 +51,95 @@
...
+     */
+    public static void convert(
+        final PrintStream out,
+        final int pageSize,
+        final File walDir,
+        final File walArchiveDir,
+        final File binaryMetadataFileStoreDir,
+        final File marshallerMappingFileStoreDir,
+        final boolean keepBinary
+
```
denis-chudov: Please add `skipCrc` to this method's parameters, and also add it to command line parameters, or get from system properties inside of `main` method.

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
```diff
@@ -0,0 +1,381 @@
...
+import org.apache.ignite.configuration.WALMode;
+import org.apache.ignite.internal.IgniteEx;
+import org.apache.ignite.internal.pagemem.wal.record.WALRecord;
+import org.apache.ignite.internal.processors.cache.persistence.wal.serializer.RecordV1Serializer;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/**
+ * Test for IgniteWalConverter
```
denis-chudov: You have added the new test class, but didn't add it into any test suite. Please do it.

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
a-polyakov: Create https://ggsystems.atlassian.net/browse/GG-26945

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
```diff
@@ -35,4 +41,86 @@
...
+
+        if (valueStr == null || valueStr.isEmpty()) {
+            try {
+                valueStr = Base64.getEncoder().encodeToString(entry.value().valueBytes(cacheObjValCtx));
+            }
+            catch (IgniteCheckedException e) {
+                cacheObjValCtx.kernalContext().log(UnwrapDataEntry.class)
+                    .error("Unable to convert value [" + entry.value() + "]", e);
+            }
+        }
```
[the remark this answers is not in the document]
a-polyakov: Differ, in one case for "key" in the second for " value"

modules/dev-utils/pom.xml
```diff
@@ -45,6 +45,34 @@
...
+            <version>${spring.version}</version>
+            <scope>test</scope>
+        </dependency>
+
+        <dependency>
+            <groupId>org.springframework</groupId>
+            <artifactId>spring-context</artifactId>
+            <version>${spring.version}</version>
+            <scope>test</scope>
+        </dependency>
```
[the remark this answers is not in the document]
a-polyakov: Yes, for test

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
a-polyakov: Added a stub, the implementation will enter https://ggsystems.atlassian.net/browse/GG-26099

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
a-polyakov: It is enough to configure in TC

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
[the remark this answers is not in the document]
denis-chudov: Yes, and there are exactly the same actions for Object key and Object value to convert them to String - and I suppose you can compactify the code by moving them to separate methods.

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
denis-chudov: It's much better to have this configuration in code. In TC we only have suites, configured in project settings - we are not going to configure each single test there.

a-polyakov: module "ignite-dev-utils" don't have a test, IgniteWalConverterTest the first test, by default all tests of the module will be run

modules/core/src/main/java/org/apache/ignite/internal/pagemem/wal/record/UnwrappedDataEntry.java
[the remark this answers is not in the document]
a-polyakov: extracted

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterTest.java
denis-chudov: Sorry, I missed that.
FYI, we are going to add suite for dev-utils in https://ggsystems.atlassian.net/browse/GG-27221

PR #810, opened 2019-12-20
GG-26728 .NET Thin Client: Fix WithExpiryPolicy crash on old servers

ptupitsyn:
Old servers do not support the extra flag, which causes exception and disconnect.
* Refactor client operation handling to be able to check ProtocolVersion of the exact connection that is used for the given operation
* Add `ProtocolVersion` check for cache operations with expiry policy.
* Fix `ProtocolVersion` check for `CacheConfiguration`


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientResponseContext.cs
```diff
@@ -0,0 +1,81 @@
...
+namespace Apache.Ignite.Core.Impl.Client
+{
+    using System.Diagnostics;
+    using Apache.Ignite.Core.Impl.Binary;
+    using Apache.Ignite.Core.Impl.Binary.IO;
+
+    /// <summary>
+    /// Response context.
+    /// </summary>
+    internal class ClientResponseContext
```
ashapkin: Code duplication for `ClientResponseContext `and `ClientRequestContext `could be reduced with a base class.

ptupitsyn: Good point, fixed

PR #816, opened 2019-12-20
GG-26773 Fixed features parsing for demo nodes after IGNITE-11142 changes in GridTopologyCommandHandler since v8.7.6.

akuznetsov-gridgain:


modules/web-console/web-console-common/src/main/java/org/apache/ignite/console/websocket/TopologySnapshot.java
```diff
@@ -280,12 +283,20 @@ public boolean isUltimate() {
...
         if (F.isEmpty(nodes))
-            return new byte[0];
+            return NO_FEATURES;
 
         return nodes.stream()
             .map(n -> attribute(n.getAttributes(), ATTR_IGNITE_FEATURES))
             .filter(Objects::nonNull)
-            .map(f -> BitSet.valueOf(Base64.getDecoder().decode(f.toString())))
+            .map(f -> {
+                // Demo  nodes will be stared inside Web Agent JVM.
```
akuznetsov-gridgain:
```suggestion
                // Demo nodes will be stared inside Web Agent JVM.
```

PR #823, opened 2019-12-24
GG-26800 Add cacheId to cache entity.

nva:


modules/management-console-agent/src/test/java/org/apache/ignite/agent/processor/CacheChangesProcessorTest.java
```diff
@@ -61,6 +63,45 @@ public void shouldSendInitialStates() throws Exception {
...
+        ignite.getOrCreateCache("test-cache");
+
+        assertWithPoll(() -> {
+            List<CacheInfo> cacheInfos = interceptor.getListPayload(buildClusterCachesInfoDest(cluster.id()), CacheInfo.class);
+
+            CacheInfo actual = F.find(cacheInfos, null,
+                (P1<CacheInfo>)i -> "test-cache".equals(i.getName()));
+
+            return actual != null &&
+                CU.cacheId("test-cache") == actual.getCacheId() && !actual.isCreatedBySql();
```
MaxLosevskoy: Let's also assert that they both are not null

nva: CU.cacheId("test-cache") from non empty string can't be zero.

PR #827, opened 2019-12-24
GG-24933 Optionally to do not write free list updates to WAL

ibessonov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -1636,6 +1864,197 @@ private void changed() {
...
+         * @param pageId Page id.
+         * @return {@code True} if page was found and succesfully removed, {@code false} if page not found.
+         */
+        public boolean removePage(long pageId) {
+            int stripeIdx = (int)pageId & (STRIPES_COUNT - 1);
+
+            synchronized (stripeLocks[stripeIdx]) {
+                GridLongList stripe = stripes[stripeIdx];
+
+                boolean rmvd = stripe != null && stripe.removeValue(0, pageId) >= 0;
```
ibessonov: Remove from the middle of the array will lead to System.arraycopy. What is the expected size of arrays in stripes? Is it a MAX_SIZE constant? Should we worry about it here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -1636,6 +1864,197 @@ private void changed() {
...
+        /**
+         * Poll next page from the list.
+         *
+         * @return pageId.
+         */
+        public long poll() {
+            if (size == 0)
+                return 0L;
+
+            for (int i = 0; i < STRIPES_COUNT; i++) {
```
ibessonov: False-negative results are fine, right? Is it ok if method returns 0L but there is a free page somewhere?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -1636,6 +1864,197 @@ private void changed() {
...
+         * Flush all stripes to one list and clear stripes.
+         */
+        @SuppressWarnings("NonAtomicOperationOnVolatileField")
+        public GridLongList flush() {
+            GridLongList res = null;
+
+            if (size == 0) {
+                boolean stripesChanged = false;
+
+                if (++emptyFlushCnt >= EMPTY_FLUSH_GC_THRESHOLD) {
```
ibessonov: emptyFlushCnt is not volatile, please add a comment why. I guess that it can be changed only from a checkpoint thread, but this might not be obvious later.

ibessonov (commented):
Looks good overall, I have only minor comments.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
zstan: Expected size is : MAX_SIZE / STRIPES_COUNT per stripe = 16, not so much i hope, may be once more tests with WAL=None mode give some clue here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
zstan: hope it`s ok here, i suppose i understand you, it`s a pay for sync free here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
zstan: yep, all calls under exclusive lock or in case of destroyCacheDataStore0 will be called without conc, is it enough or additional description needed ?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
ibessonov: I saw it later. If all arrays are small then everything is fine.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
ibessonov: It's enough I think

PR #831, opened 2019-12-25
GG-24031: Add running queries action.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/action/query/RunningQuery.java
```diff
@@ -0,0 +1,192 @@
...
+    /**
+     * @param duration Duration.
+     * @return {@code This} for chaining method calls.
+     */
+    public RunningQuery setDuration(long duration) {
+        this.duration = duration;
+
+        return this;
+    }
+
```
akuznetsov-gridgain:
```suggestion
```

modules/management-console-agent/src/main/java/org/apache/ignite/agent/dto/action/query/RunningQuery.java
```diff
@@ -0,0 +1,192 @@
...
+    /** Schema name. */
+    private String schemaName;
+
+    /** */
+    private long startTime;
+
+    /** */
+    private long duration;
+
```
MaxLosevskoy: Please mention what timeunit is supposed to be used here?
Is this in milliseconds? In seconds? Smth else?

MaxLosevskoy: And it can be even better to specify timeunits in Getter and Setter methods

PR #832, opened 2019-12-26
GG-26839 Add system flag to cache entity.

nva:


modules/management-console-agent/src/test/java/org/apache/ignite/agent/processor/CacheChangesProcessorTest.java
```diff
@@ -57,35 +59,14 @@ public void shouldSendInitialStates() throws Exception {
...
-        ignite.getOrCreateCache("test-cache");
-
         assertWithPoll(() -> {
-            List<CacheInfo> cachesInfo = interceptor.getListPayload(buildClusterCachesInfoDest(cluster.id()), CacheInfo.class);
+            List<CacheInfo> cacheInfos = interceptor.getListPayload(buildClusterCachesInfoDest(cluster.id()), CacheInfo.class);
 
-            return !cachesInfo.isEmpty() && cachesInfo.stream().noneMatch(i -> "ignite-sys-cache".equals(i.getName()));
+            CacheInfo actual = F.find(cacheInfos, null, (P1<CacheInfo>)i -> CU.isSystemCache(i.getName()));
+
+            return cacheInfos.size() == 1 && actual != null && actual.isSystemCache();
```
MaxLosevskoy: Do we really require `actual.isSystemCache()` condition? Can it be **false** if `CU.isSystemCache(i.getName())` is **true**?

nva: yes

PR #839, opened 2020-01-10
GG-26737 Updated dependencies

akuznetsov-gridgain:


modules/zookeeper/pom.xml
```diff
@@ -40,6 +40,12 @@
             <version>${project.version}</version>
         </dependency>
 
+        <dependency>
```
nva: Please revert this. This changes doesn't related with your ticket

modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/zk/TcpDiscoveryZookeeperIpFinder.java
```diff
@@ -42,7 +42,7 @@
 import org.apache.ignite.resources.LoggerResource;
 import org.apache.ignite.spi.IgniteSpiException;
 import org.apache.ignite.spi.discovery.tcp.ipfinder.TcpDiscoveryIpFinderAdapter;
-import org.codehaus.jackson.map.annotate.JsonRootName;
+import com.fasterxml.jackson.annotation.JsonRootName;
```
nva: Please revert this. This changes doesn't related with your ticket

parent/pom.xml
```diff
@@ -76,8 +76,7 @@
         <hamcrest.version>1.2</hamcrest.version>
         <httpclient.version>4.5.1</httpclient.version>
         <httpcore.version>4.4.3</httpcore.version>
-        <jackson.version>2.9.9</jackson.version>
-        <jackson1.version>1.9.13</jackson1.version>
```
nva: Please revert this. This changes doesn't related with your ticket

modules/zookeeper/pom.xml
akuznetsov-gridgain: Reverted

modules/zookeeper/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/zk/TcpDiscoveryZookeeperIpFinder.java
akuznetsov-gridgain: Reverted

parent/pom.xml
akuznetsov-gridgain: Reverted

PR #854, opened 2020-01-16
GG-26148 fix reserving checkpoints when segment is corrupted

alievmirza:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
```diff
@@ -458,7 +458,16 @@ private long absFileIdx(CheckpointEntry pointer) {
                 }
             }
             catch (IgniteCheckedException ex) {
-                U.error(log, "Failed to process checkpoint: " + (chpEntry != null ? chpEntry : "none"), ex);
+                U.warn(log, "Failed to process checkpoint: " + (chpEntry != null ? chpEntry : "none"));
```
glukos: Let's pass exception "ex" to U.warn - stack trace would be helpful to see.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1340,6 +1340,11 @@
      */
     public static final String IGNITE_PAGES_LIST_DISABLE_ONHEAP_CACHING = "IGNITE_PAGES_LIST_DISABLE_ONHEAP_CACHING";
 
+    /**
+     * Disable group state lazy store. Should be used for test purposes only.
```
glukos: Please add more descriptive comment on how does this property affect the system's behavior.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/crc/WalTestUtils.java
```diff
@@ -52,7 +57,7 @@
      * @throws IOException If IO exception.
      * @throws IgniteCheckedException If iterator failed.
      */
-    public static FileWALPointer corruptWalSegmentFile(
+    public static FileWALPointer corruptWalRecord(
```
glukos: maybe "corruptRandomWalRecord" is better name?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/crc/WalTestUtils.java
```diff
@@ -81,7 +86,7 @@ public static FileWALPointer corruptWalSegmentFile(
      * @param desc WAL segment descriptor.
      * @param pointer WAL pointer.
      */
-    public static void corruptWalSegmentFile(
+    public static void corruptWalRecord(
```
glukos: This method and the one above have the same javadocs.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/crc/WalTestUtils.java
```diff
@@ -98,6 +103,78 @@ public static void corruptWalSegmentFile(
...
+    }
+
+    /**
+     * Corrupts zip file. According to section 4.3.7 of zip file specification, this method just sets 0
+     * to the first 30 bytes of the first local file header to achieve corruption.
+     *
+     * @param desc File descriptor.
+     * @see <a href="https://pkware.cachefly.net/webdocs/casestudies/APPNOTE.TXT">Zip file specification </a>
+     */
+    public static void corruptCompressedSegment(FileDescriptor desc) throws IOException {
```
glukos: This method doesn't necessarily work with segments, it's applicable to any zip file.
Shouldn't we rename it?

PR #858, opened 2020-01-16
GG-26831

akalash:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -3276,31 +3296,28 @@ private void finalizeCheckpointOnRecovery(
...
-                        pageMem.checkpointWritePage(fullId, writePageBuf, pageStoreWriter, null);
-                    }
+                        pageMem.checkpointWritePage(curPageId.get2(), writePageBuf, pageStoreWriter, null);
 
-                    // Add number of handled pages.
-                    cpPagesCnt.addAndGet(pages0.size());
+                        curPageId = pages.poll();
+
+                        // Add number of handled pages.
+                        cpPagesCnt.incrementAndGet();
```
akalash: Maybe it's better to increment global cpPagesCnt only one time at the end of the thread execution but inside of the thread using the local counter?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -4270,9 +4277,11 @@ private Checkpoint markCheckpointBegin(CheckpointMetricsTracker tracker) throws
 
             IgniteFuture snapFut = null;
 
-            IgniteBiTuple<Collection<GridMultiCollectionWrapper<FullPageId>>, Integer> cpPagesTuple;
+            T3<Collection<T2<PageMemoryEx, GridMultiCollectionWrapper<FullPageId>>>, Integer, Boolean> cpPagesTriple;
```
akalash: Maybe it makes sense to replace this ugly T3 + T2 construction to some specific class with clear fields?

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,162 @@
...
+ * p1 = new Pair<1, [1, 3, 5, 7]> <br>
+ * p2 = new Pair<2, [2, 3]> <br>
+ * p3 = new Pair<3, [200, 100]> <br>
+ * and further sequence of {@code poll} calls may produce output like: <br>
+ * [3, 200], [3, 100], [1, 1], [1, 3], [1, 5], [1, 7], [2, 2], [2, 3]
+ *
+ * @param <K> The type of key in input pair collection.
+ * @param <V> The type of value array.
+ */
+public class GridConcurrentMultiPairQueue<K, V> {
```
akalash: What do you think about the following idea:
Instead of class like this which contains key(PageMemory) and value(Pages) create a class which would be returned iterator over all PageMemory which in turn would be returned iterator over pages.

Something like that
```
PageMemory memory = pageMemories.poll();
while(memory != null){
   FullPageId page = memory.poll();
   while(page != null){
      ...
      page = memory.poll();
   }
   memory = pageMemories.poll();
}
```

This method, at least, helps to avoid wrapping in T2 on every page during iteration.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1296,6 +1303,95 @@ public void testDoInParallelException() {
         }
     }
 
+    /**
+     *
+     */
+    @Test
+    public void testGridConcurrentMultiPairQueueCorrectness() throws Exception {
```
akalash: This test contains two different scenarios, can you separate it accordingly, please. Also, I think it makes sense to extract this test to their personal class like GridConcurrentMultiPairQueueTest

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -4826,68 +4778,60 @@ public void awaitPendingTasksFinished() throws IgniteCheckedException {
...
 
-        if (persistenceCfg.getCheckpointWriteOrder() == CheckpointWriteOrder.SEQUENTIAL) {
-            Comparator<FullPageId> cmp = new Comparator<FullPageId>() {
-                @Override public int compare(FullPageId o1, FullPageId o2) {
-                    int cmp = Long.compare(o1.groupId(), o2.groupId());
-                    if (cmp != 0)
-                        return cmp;
+            for (int i = 0; i < regPages.get2().collectionsSize(); i++) {
+                for (FullPageId page : regPages.get2().innerCollection(i)) {
+                    if (realPagesArrSize++ == totalPagesCnt)
```
ibessonov: This has to be "++realPagesArrSize", right? Very dangerous

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -4826,68 +4778,60 @@ public void awaitPendingTasksFinished() throws IgniteCheckedException {
...
+                    return cmp1;
 
-            int to = (int)((long)realPagesArrSize * (i + 1) / pagesSubLists);
+                return Long.compare(o1.effectivePageId(), o2.effectivePageId());
+            };
 
-            pagesSubListArr[i] = new GridReadOnlyArrayView(pagesArr, from, to);
+            for (T2<PageMemoryEx, FullPageId[]> pagesPerReg : cpPagesPerRegion) {
+                if (pagesPerReg.getValue().length >= parallelSortThreshold)
+                    parallelSortInIsolatedPool(pagesPerReg.get2(), cmp);
```
ibessonov: This method will create new pool for every region the exceeds its threshold, but we can reuse it.

You see, this logic was very simple before the refactoring. Single array was already sorted by groupId, so resorting was not super heavy. Do you have some benchmarks showing that this loop actually helps?

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,162 @@
...
+
+/**
+ * Concurrent queue that wraps collection of {@code Pair<K, V[]>}
+ * The only garantee {@link #poll} provided is sequentially emtify values per key array.
+ * i.e. input like: <br>
+ * p1 = new Pair<1, [1, 3, 5, 7]> <br>
+ * p2 = new Pair<2, [2, 3]> <br>
+ * p3 = new Pair<3, [200, 100]> <br>
+ * and further sequence of {@code poll} calls may produce output like: <br>
+ * [3, 200], [3, 100], [1, 1], [1, 3], [1, 5], [1, 7], [2, 2], [2, 3]
```
ibessonov: Is this example correct? I see order "3, 1, 2" here, this confuses me.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
ibessonov: Another way to approach this problem is to pass handler into the queue and it will be internally iterated, like "forEach((a, b) -> doSomeAction(a, b))".

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,162 @@
...
+        keysArr = (K[])new Object[pairCnt];
+
+        lenSeq = new long[pairCnt];
+
+        int keyPos = 0;
+
+        int size = -1;
+
+        for (Map.Entry<K, ? extends Collection<V>> p : items.entrySet()) {
+            if (p.getValue().size() == 0)
```
ibessonov: "isEmpty()" would look nicer I think

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,162 @@
...
+     *
+     * @return the head of this queue, or {@code null} if this queue is empty
+     */
+    public @Nullable T2<K, V> poll() {
+        long absPos = pos.getAndIncrement();
+
+        if (absPos >= maxPos)
+            return null;
+
+        int segment = cachedSegment.get();
```
ibessonov: My advice with "forEach" will allow you to throw away this ThreadLocal cachedSegment so it won't be this complicated.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,162 @@
...
+
+        if (absPos > lenSeq[segment]) {
+            segment = Arrays.binarySearch(lenSeq, absPos);
+
+            segment = segment < 0 ? -segment - 1 : segment;
```
ibessonov: I have not read your tests yet. Does this particular line has its own test case?

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,162 @@
...
+    public @Nullable T2<K, V> poll() {
+        long absPos = pos.getAndIncrement();
+
+        if (absPos >= maxPos)
+            return null;
+
+        int segment = cachedSegment.get();
+
+        if (absPos > lenSeq[segment]) {
+            segment = Arrays.binarySearch(lenSeq, absPos);
```
ibessonov: If you like binary searches this much then you can specify lower bound as "segment+1" ;)

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1296,6 +1303,95 @@ public void testDoInParallelException() {
...
+
+        final GridConcurrentMultiPairQueue<Integer, Integer> queue = new GridConcurrentMultiPairQueue<>(keyWithArr);
+
+        GridTestUtils.runMultiThreaded(() -> {
+            T2<Integer, Integer> pair = queue.poll();
+
+            while (pair != null) {
+                assertTrue(mapForCheck.containsKey(pair.get1()));
+
+                mapForCheck.get(pair.get1()).remove(pair.get2());
```
ibessonov: Please assert that the element was removed by this particular invocation and not by other thread.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1296,6 +1303,95 @@ public void testDoInParallelException() {
...
+
+                if (mapForCheck.get(pair.get1()).isEmpty())
```
ibessonov: You can fail with NPE here, collection might have been removed by another thread. Please be careful.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1296,6 +1303,95 @@ public void testDoInParallelException() {
...
+
+        final GridConcurrentMultiPairQueue<Integer, Integer> queue2 = new GridConcurrentMultiPairQueue<>(keyWithColl);
+
+        GridTestUtils.runMultiThreaded(() -> {
+            T2<Integer, Integer> pair = queue2.poll();
+
+            while (pair != null) {
+                assertTrue(mapForCheck2.containsKey(pair.get1()));
+
+                mapForCheck2.get(pair.get1()).remove(pair.get2());
```
ibessonov: Same comments on this loop.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1296,6 +1303,95 @@ public void testDoInParallelException() {
...
+            }
+        }, ThreadLocalRandom.current().nextInt(1, 20), "GridConcurrentMultiPairQueue arr test");
+
+        assertTrue(mapForCheck.isEmpty());
+
+        assertTrue(queue.isEmpty());
+
+        assertTrue(queue.initialSize() == arr1.length + arr2.length + arr3.length + arr4.length);
+
+        Map<Integer, Collection<Integer>> keyWithColl = new HashMap<>();
```
ibessonov: Why do you repeat basically the same test twice? Am I missing something, can you please write come comment here?

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
zstan: fixed using @ibessonov below approach.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
zstan: yep, done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
zstan: good catch, thanks ! :) fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
zstan: refactor it, plz check.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
zstan: yes, this is correct, the documemntation for class explains the same:
The only garantee {@link #poll} or {@link next} provided is sequentially emptify values per key array.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
zstan: i use similar but little differ approach, not lambda`s but exact class for params extraction.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
zstan: -segment - 1 from bsearch documentation shows that input key not exactly belongs to array but nested between to indexes.

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
zstan: cause GridConcurrentMultiPairQueue has two constructors for 2 different interfaces Map and Collection, it`s not easy to squash all GridConcurrentMultiPairQueue usage with one constructor, if you not agree plz highlight it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
zstan: no, i think a little bit more and hope i correct here !!

ibessonov: I insist you to think one more time. Anyway, what's the point of this check? It validated array overflow in previous code. Now we have smaller array and we write into it without such validation. These are just my thoughts.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -4910,9 +4851,10 @@ private static void parallelSortInIsolatedPool(
             }
         };
 
-        ForkJoinPool forkJoinPool = new ForkJoinPool(PARALLEL_SORT_THREADS + 1, factory, null, false);
+        ExecutorService execPool = pool == null ?
```
ibessonov: Hmmmmm..........
Can you please put back that ForkJoinPool? Otherwise the default one will be used.
It was the whole point of introducing additional pool here - avoid using default one.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,229 @@
...
+    private final AtomicInteger pos = new AtomicInteger();
+
+    /** Precalculated max position. */
+    private final int maxPos;
+
+    /** Keys array. */
+    private final K[] keysArr;
+
+    /** Thread local cached segment. */
+    private ThreadLocal<Integer> cachedSegment = ThreadLocal.withInitial(() -> 0);
```
ibessonov: No one calls ThreadLocal#remove on this one, this will lead to memory leak in checkpoint thread.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,229 @@
...
+    public @Nullable T2<K, V> poll() {
+        int absPos = pos.getAndIncrement();
+
+        if (absPos >= maxPos)
+            return null;
+
+        int segment = cachedSegment.get();
+
+        if (absPos > lenSeq[segment]) {
+            segment = Arrays.binarySearch(lenSeq, segment, lenSeq.length - 1, absPos);;
```
ibessonov: ";;"

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
ibessonov: I know that) What I meant is that it's easy to miss some "+1" or "-1" in these tricky expressions, I'm always a bit paranoid about such errors.
I rechecked this line with paper and pen, it's correct.

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
```diff
@@ -0,0 +1,229 @@
...
+
+        int relPos = segment == 0 ? (int)absPos : (int)(absPos - lenSeq[segment - 1] - 1);
+
+        K key = keysArr[segment];
+
+        return new T2<>(key, vals[segment][relPos]);
+    }
+
+    /** Return next initialized input class instance. */
+    public boolean next(Result<K, V> res) {
```
ibessonov: I see that this method copy-pasted. Can we reuse code of "poll"? Does someone uses that method? I believe that you only use it in tests.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -4896,9 +4836,10 @@ public void awaitPendingTasksFinished() throws IgniteCheckedException {
      * @param pagesArr Pages array.
      * @param cmp Cmp.
      */
-    private static void parallelSortInIsolatedPool(
+    private static ExecutorService parallelSortInIsolatedPool(
         FullPageId[] pagesArr,
-        Comparator<FullPageId> cmp
+        Comparator<FullPageId> cmp,
+        ExecutorService pool
```
ibessonov: Please use ForkJoinPool as parameter type. Not every executor service is suitable for the job.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
zstan: I don`t understand your suggestion here, you think this code should be replaced ?

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
zstan: no one calls "poll" now + GridConcurrentMultiPairQueue instance has life time smaller than checkpoint frequency, do we need some changes here ?

modules/core/src/main/java/org/apache/ignite/internal/util/GridConcurrentMultiPairQueue.java
zstan: I change signature, now they didn`t clone each other.

PR #865, opened 2020-01-21
Gg 26977

pavlukhin:


modules/h2/src/main/java/org/h2/table/Table.java
```diff
@@ -1342,4 +1342,15 @@ else if (b == ValueNull.INSTANCE)
 
         return (comp > 0) == bigger ? a : b;
     }
+
+    public Column[] getVisibleColums() {
+        List<Column> visibleCols = new ArrayList<>(columns.length);
```
ygerzhedovich: may be stream operations?

tledkov-gridgain: The source version of the h2 module is 1.7.
Streams not supported.

PR #867, opened 2020-01-21
GG-23582

ascherbakoff:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -1485,69 +1485,89 @@ private void prepareIndexRebuildFuture(int cacheId) {
...
-                            }
-                        });
+                rebuildFut.listen(fut -> {
+                    idxRebuildFuts.remove(cacheId, usrFut);
+
+                    Throwable err = fut.error();
+
+                    usrFut.onDone(err);
+
+                    if (isNull(err)) {
```
ascherbakoff: These methods are indented to use in predicates.
Using them in _if_ conditions looks weird.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -1996,9 +2017,13 @@ private boolean rebuildIsMeaningless(GridCacheContext cctx) {
      * Rebuilds indexes for provided caches from corresponding hash indexes.
      *
      * @param cctx Cache context.
+     * @param execSvc Thread pool for rebuilding indexes.
      * @return Future that will be completed when rebuilding is finished.
      */
-    public IgniteInternalFuture<?> rebuildIndexesFromHash(GridCacheContext cctx) {
+    public IgniteInternalFuture<?> rebuildIndexesFromHash(GridCacheContext cctx, ExecutorService execSvc) {
```
ascherbakoff: What the point in having an argument for executor and exposing executor at all ?
Looks like you can use single instance allowing threads timeout and doesn't allowing more threads than number of cores.
I see no need in dynamic executor management.
The logic should be placed to org.apache.ignite.internal.GridKernalContext

exec = new IgniteThreadPoolExecutor(
            "rebuild-idx-runner",
            ctx.igniteInstanceName(),
            0,
            parallelism,
            DFLT_THREAD_KEEP_ALIVE_TIME,
            new LinkedBlockingQueue<>(),
            GridIoPolicy.UNDEFINED,
            oomeHnd
        );

exec .allowCoreThreadTimeOut(true);

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -1485,69 +1485,89 @@ private void prepareIndexRebuildFuture(int cacheId) {
...
+            GridFutureAdapter<Void> usrFut = idxRebuildFuts.get(cacheId);
 
-                        compoundAllIdxsRebuilt.add(rebuildFut);
+            if (isNull(rebuildIdxExecSvc))
+                rebuildIdxExecSvc = qryProc.getIndexing().rebuildIndexExecutorService(0);
 
-                        rebuildFut.listen(new CI1<IgniteInternalFuture>() {
-                            @Override public void apply(IgniteInternalFuture fut) {
-                                idxRebuildFuts.remove(cacheId, usrFut);
+            IgniteInternalFuture<?> rebuildFut = qryProc.rebuildIndexesFromHash(cacheCtx, rebuildIdxExecSvc);
```
ascherbakoff: Why rebuildFut is necessary ? Looks like usrFut can be used instead.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -621,7 +639,7 @@ public static int operationTimeout(int qryTimeout, IgniteTxAdapter tx) {
         if (tx != null) {
             int remaining = (int)tx.remainingTime();
 
-            return remaining > 0 && qryTimeout > 0 ? Math.min(remaining, qryTimeout) : Math.max(remaining, qryTimeout);
+            return remaining > 0 && qryTimeout > 0 ? min(remaining, qryTimeout) : Math.max(remaining, qryTimeout);
```
ascherbakoff: Why Math.min is statically imported but Math.max is not ?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
tkalkirill: I think it doesn't look like a mistake, I leave it as it is.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
tkalkirill: Method [1] and [2] stores GridFutureAdapter. do you think it's better to store GridCompoundFuture in the [2]?

[1] - org.apache.ignite.internal.processors.query.h2.IgniteH2Indexing#rebuildIndexesFromHash returns GridCompoundFuture
[2] - org.apache.ignite.internal.processors.cache.persistence.GridCacheDatabaseSharedManager#idxRebuildFuts

modules/core/src/main/java/org/apache/ignite/internal/GridKernalContext.java
```diff
@@ -751,4 +751,11 @@
      * @return Local continuous tasks processor.
      */
     public DurableBackgroundTasksProcessor durableBackgroundTasksProcessor();
+
+    /**
+     * Return Thread pool for create/rebuild indexes.
+     *
+     * @return Thread pool for create/rebuild indexes.
+     */
+    ExecutorService buildIndexExecutorService();
```
ygerzhedovich: let's add `public` for common style.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -1928,6 +1931,23 @@ private void start0(GridStartContext startCtx, IgniteConfiguration cfg, TimeBag
...
+                int buildIdxThreadPoolSize = cfg.getBuildIndexThreadPoolSize();
+
+                validateThreadPoolSize(buildIdxThreadPoolSize, "build-idx");
+
+                buildIdxExecSvc = new IgniteThreadPoolExecutor(
+                    "build-idx-runner",
+                    cfg.getIgniteInstanceName(),
+                    0,
+                    buildIdxThreadPoolSize,
+                    DFLT_THREAD_KEEP_ALIVE_TIME,
```
AMashenkov: Default keepalive time is 60sec. 
Is there any reason to have idle threads alive?

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -1928,6 +1931,23 @@ private void start0(GridStartContext startCtx, IgniteConfiguration cfg, TimeBag
...
+                    new LinkedBlockingQueue<>(),
+                    GridIoPolicy.UNDEFINED,
+                    oomeHnd
+                );
+
+                buildIdxExecSvc.allowCoreThreadTimeOut(true);
```
AMashenkov: This make no sense as coreThreads is set to "0"

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
tkalkirill: I can't find the reason, but what's wrong with the default value?

PR #871, opened 2020-01-23
Checking of assumption that nested locks are taken from one place

alievmirza:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/atomic/GridDhtAtomicCache.java
```diff
@@ -2999,7 +2999,7 @@ else if (GridDhtCacheEntry.ReaderId.contains(readers, nearNode.id())) {
             while (true) {
                 GridDhtCacheEntry entry = entryExx(key, topVer);
 
-                entry.lockEntry();
+                entry.lockEntry(true);
```
alievmirza:
```suggestion
                entry.lockEntry(false);
```

PR #873, opened 2020-01-24
gg-26229 non-classpath binary objects support added

glukos:
Signed-off-by: Ivan Rakov <irakov@apache.org>

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/tasks/RepairRequestTask.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/tasks/RepairRequestTask.java
@@ -93,26 +93,27 @@ public class RepairRequestTask extends ComputeTaskAdapter<RepairRequest, Executi
         Map<UUID, Map<KeyCacheObject, Map<UUID, VersionedValue>>> targetNodesToData = new HashMap<>();
 
         for (Map.Entry<KeyCacheObject, Map<UUID, VersionedValue>> dataEntry: repairReq.data().entrySet()) {
-            KeyCacheObject keyCacheObject;
+            KeyCacheObject keyCacheObj;
+
+            GridCacheContext<Object, Object> ctx = ignite.cachex(repairReq.cacheName()).context();
 
             try {
-                keyCacheObject = unmarshalKey(dataEntry.getKey(), ignite.cachex(repairReq.cacheName()).context());
+                keyCacheObj = unmarshalKey(dataEntry.getKey(), ctx);
             }
-            catch (IgniteCheckedException e) {
+            catch (IgniteCheckedException ignored) {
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/tasks/RepairRequestTask.java
```diff
@@ -93,26 +93,27 @@
...
+            KeyCacheObject keyCacheObj;
+
+            GridCacheContext<Object, Object> ctx = ignite.cachex(repairReq.cacheName()).context();
 
             try {
-                keyCacheObject = unmarshalKey(dataEntry.getKey(), ignite.cachex(repairReq.cacheName()).context());
+                keyCacheObj = unmarshalKey(dataEntry.getKey(), ctx);
             }
-            catch (IgniteCheckedException e) {
+            catch (IgniteCheckedException ignored) {
```
sanpwc: Seems that it's not ignored cause we log an exception.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/util/ConsistencyCheckUtils.java
```diff
@@ -240,7 +240,7 @@ public static KeyCacheObject unmarshalKey(KeyCacheObject unmarshalKey,
                     cacheObjOpt.isPresent() ?
                         new PartitionReconciliationValueMeta(
                             cacheObjOpt.get().valueBytes(ctx),
-                            Optional.ofNullable(cacheObjOpt.get().value(ctx, false)).map(Object::toString).orElse(null),
+                            cacheObjOpt.map(o -> objectStringView(ctx, o)).orElse(null),
```
sanpwc: For simple keys that'll return almost useless string representation: "CacheObjectImpl [val=null, hasValBytes=true]"
Optional.ofNullable(cacheObjOpt.get().value(ctx, false)).map(Object::toString) will return "abc_100".
As a result, in case of simple keys, partition reconciliation report in --verbose mode won't contain human readable version of per-node-values. Seems that GridCommandHandlerPartitionReconciliationAbstractTest'tests will fail.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/tasks/RepairRequestTask.java
glukos: We didn't log an exception; fixed that and variable name.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/util/ConsistencyCheckUtils.java
glukos: Fixed

PR #874, opened 2020-01-26
GG-27216 Extension log in rebuild indexes to search problems.

a-polyakov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryEx.java
```diff
@@ -1001,10 +1002,12 @@ public boolean lockedLocallyByIdOrThread(GridCacheVersion lockVer, long threadId
      * Update index from within entry lock, passing key, value, and expiration time to provided closure.
      *
      * @param clo Closure to apply to key, value, and expiration time.
+     * @return Type descriptor (can be {@code null}).
      * @throws IgniteCheckedException If failed.
      * @throws GridCacheEntryRemovedException If entry was removed.
+     * @return Type descriptor.
      */
-    public void updateIndex(SchemaIndexCacheVisitorClosure clo)
+    public @Nullable GridQueryTypeDescriptor updateIndex(SchemaIndexCacheVisitorClosure clo)
```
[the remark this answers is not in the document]
slukyano: I guess other changed methods need to be nullable too.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -4464,24 +4465,34 @@ protected void removeValue() throws IgniteCheckedException {
...
 
             checkObsolete();
 
             CacheDataRow row = cctx.offheap().read(this);
 
-            if (row != null)
+            if (row != null) {
                 clo.apply(row);
+
+                return cctx.kernalContext().query().typeByValue(
```
slukyano: It is a bit worrying that we started doing this for every put. It looks like we at least need to pass `false` in the last parameter.

slukyano: Looking at the code more closely, I think it would be better to call this from processPartition method and not even change updateIndex and processKey. Moreover, I suggest to move all of new code under the new system property - in case this have unnoticed performance impact.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheVisitorImpl.java
```diff
@@ -155,28 +223,35 @@ public SchemaIndexCacheVisitorImpl(GridCacheContext cctx, SchemaIndexOperationCa
...
      * @param remainder Remainder.
+     * @return Index statistics.
      * @throws IgniteCheckedException If failed.
      */
-    private void processPartitions(List<GridDhtLocalPartition> parts, SchemaIndexCacheVisitorClosure clo,
+    private SchemaIndexCacheStat processPartitions(List<GridDhtLocalPartition> parts, SchemaIndexCacheVisitorClosure clo,
         int remainder)
         throws IgniteCheckedException {
+
+        SchemaIndexCacheStat tmp = new SchemaIndexCacheStat();
```
slukyano: I'd suggest to return null and not collect any stats if the new sys prop is false.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheStat.java
```diff
@@ -0,0 +1,35 @@
...
+
+package org.apache.ignite.internal.processors.query.schema;
+
+import java.util.HashSet;
+import java.util.Set;
+
+/**
+ * Class for accumulation of record types and number of indexed records in index tree.
+ */
+public class SchemaIndexCacheStat {
```
slukyano: Does this need to be public?

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheStat.java
```diff
@@ -0,0 +1,35 @@
...
+    /**
+     * Indexed types.
+     */
+    public Set<String> types = new HashSet<>();
```
slukyano: final?

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheStat.java
antonovsergey93: Changed to default

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -4752,6 +4752,8 @@ private void markCheckpointEnd(Checkpoint chp) throws IgniteCheckedException {
             isCancelled = true;
 
             synchronized (this) {
+                scheduledCp.reason = "shutdown";
```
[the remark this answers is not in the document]
antonovsergey93: done

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheVisitorImpl.java
antonovsergey93: done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -4480,19 +4483,23 @@ protected void removeValue() throws IgniteCheckedException {
...
-                    row.value(),
-                    true
-                );
+                if (stat != null) {
+                    QueryTypeDescriptorImpl type = cctx.kernalContext().query().typeByValue(
+                        cctx.cache().name(),
+                        cctx.cacheObjectContext(),
+                        row.key(),
+                        row.value(),
+                        true
```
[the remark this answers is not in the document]
antonovsergey93: No, we can't. IndexRebuildPartialClosure doesn't make that check with the`true` flag in the logic.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaIndexCacheStat.java
```diff
@@ -16,17 +16,18 @@
...
 /**
  * Class for accumulation of record types and number of indexed records in index tree.
  */
-class SchemaIndexCacheStat {
+public class SchemaIndexCacheStat {
     /**
      * Indexed types.
      */
-    public Set<String> types = new HashSet<>();
+    public Map<String, QueryTypeDescriptorImpl> types = new HashMap<>();
```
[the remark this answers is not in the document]
antonovsergey93: of course

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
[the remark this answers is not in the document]
slukyano: OK, nevermind then

PR #887, opened 2020-01-30
GG-21448

akalash:


modules/core/src/main/java/org/apache/ignite/events/EventType.java
```diff
@@ -795,6 +795,18 @@
...
+     * Built-in event type: Cluster state change initiated.
+     * <p>
+     * Fired when cluster deactivation process started.
+     * <p>
+     * NOTE: all types in range <b>from 1 to 1000 are reserved</b> for
+     * internal Ignite events and should not be used by user-defined events.
+     *
+     * @see ClusterStateChangeStartedEvent
+     */
+    public static final int EVT_CLUSTER_STATE_CHANGE_STARTED = 145;
```
akuznetsov-gridgain: It is OK, that `EVT_CLUSTER_STATE_CHANGE_STARTED` is `145` and not `144`?

PR #890, opened 2020-01-30
GG-25083 .NET: Add native Near Cache

ptupitsyn:
Add .NET Near Cache:

* When enabled, a local map (`ConcurrentDictionary<,>`) is created to store deserialized values on .NET side
* Implementation is based on Java Near Cache: callbacks for updates and evictions keep the local map up-to-date
* On server nodes, in addition to near cache entries, this local map stores all primary keys and their values


modules/core/src/main/java/org/apache/ignite/configuration/PlatformNearCacheConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+import org.apache.ignite.internal.util.typedef.internal.S;
+
+import java.io.Serializable;
+
+/**
+ * Platform near cache configuration.
+ * <p>
+ * Additional near caching mechanism on platform side (.NET).
+ */
+public class PlatformNearCacheConfiguration implements Serializable {
```
ashapkin: Do we really need this class?
From my understanding, it is reserved for other platforms, otherwise, we could just send a raw byte array between .NET nodes.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCache.cs
```diff
@@ -0,0 +1,261 @@
...
+            }
+
+            NearCacheEntry<TV> entry;
+            var key0 = (TK) (object) key;
+            
+            if (_map.TryGetValue(key0, out entry))
+            {
+                if (IsValid(entry))
+                {
+                    val = (TVal) (object) entry.Value;
```
ashapkin: You mentioned no-boxing approach, am I missing something?

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCache.cs
```diff
@@ -0,0 +1,261 @@
...
+
+            if (entryVer >= currentVer)
+            {
+                return true;
+            }
+
+            var part = entry.Partition;
+            var valid = _affinity.IsAssignmentValid(entryVer, part);
+
+            // Update version or mark as invalid (null).
```
ashapkin: Sry for possible question duplicate, can't we set a callback for this affinity changes directly from the Java side?

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCache.cs
```diff
@@ -0,0 +1,261 @@
...
+
+            var mode = _keepBinary ? BinaryMode.ForceBinary : BinaryMode.Deserialize;
+            var reader = marshaller.StartUnmarshal(stream, mode);
+
+            var key = reader.ReadObject<TK>();
+            var hasVal = stream.ReadBool();
+
+            if (hasVal)
+            {
+                var val = reader.ReadObject<TV>();
```
ashapkin: Is there are tests for keepBinary enabled mode?

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCacheManager.cs
```diff
@@ -0,0 +1,193 @@
...
+        }
+        
+        /// <summary>
+        /// Handles client disconnect.
+        /// </summary>
+        private void OnClientDisconnected(object sender, EventArgs e)
+        {
+            foreach (var cache in _nearCaches)
+            {
+                cache.Value.Clear();
```
ashapkin: FYI: Could hit performance, customers could use a thick client for health checking instead of a thin one

modules/platforms/dotnet/Apache.Ignite.Core/Cache/Configuration/CacheConfiguration.cs
```diff
@@ -448,6 +453,16 @@ internal void Write(BinaryWriter writer, ClientProtocolVersion srvVer)
             ExpiryPolicySerializer.WritePolicyFactory(writer, ExpiryPolicyFactory);
 
             writer.WriteCollectionRaw(KeyConfiguration);
+            
+            if (PlatformNearConfiguration != null)
```
ashapkin: For some reasons, this config seems to not having affect when creating cache with Ignite.CreateCache(cfg), but it works on Ignite.CreateCache(cfg, nearCfg)

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
```diff
@@ -949,6 +1243,45 @@ private Task<long> SizeAsync0(int? part, params CachePeekMode[] modes)
...
+            onlyNativeNear = false;
+            
+            bool hasPlatformNear;
+            var modes0 = IgniteUtils.EncodePeekModes(modes, out hasPlatformNear);
+
+            if (hasPlatformNear)
+            {
+                if (!loc)
+                {
+                    throw new InvalidOperationException(
```
ashapkin: It feels that Peek.All should work as "all available" instead of throwing an explicit exception.
Please, check tests, we could convert this into a warning temporary.

modules/core/src/main/java/org/apache/ignite/configuration/PlatformNearCacheConfiguration.java
isapego: Why not? To me it looks like a good idea to allow user to configure this from xml.

modules/core/src/main/java/org/apache/ignite/configuration/PlatformNearCacheConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+    /** */
+    private static final long serialVersionUID = 0L;
+
+    /** Key type name. */
+    private String keyTypeName;
```
isapego: By the way, is it possible to get this info from cache configuration in some cases?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtCacheEntry.java
```diff
@@ -165,6 +165,8 @@ public GridDhtCacheEntry(
 
     /** {@inheritDoc} */
     @Override public void onMarkedObsolete() {
+        super.onMarkedObsolete();
```
isapego: Was it originally a bug?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/PlatformContextImpl.java
```diff
@@ -80,10 +83,13 @@
  * Implementation of platform context.
  */
 @SuppressWarnings("TypeMayBeWeakened")
-public class PlatformContextImpl implements PlatformContext {
+public class PlatformContextImpl implements PlatformContext, PartitionsExchangeAware {
     /** Supported event types. */
     private static final Set<Integer> evtTyps;
 
+    /** Whether to use thread-local data to update platform near cache. */
+    private static final ThreadLocal<Boolean> nearUpdateUseThreadLocal = new ThreadLocal<>();
```
isapego: Do we use TLS here to avoid redundant copies on put from platform side? Looks a bit complicated.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/PlatformProcessor.java
```diff
@@ -47,6 +48,13 @@
      */
     public PlatformContext context();
 
+    /**
+     * Returns a value indicating whether {@link #context()} is available.
+     *
+     * @return value indicating whether {@link #context()} is available.
+     */
+    public boolean hasContext();
```
isapego: I don't think this is a good name for this method as it does not say anything for person reading the code. I actually had to go to implementation of `context()` method in `PlatformNoopProcessor` to understand it's meaning. Maybe `hasPlatformRuntime()` or `isPlatformNode()` would fit better here?

modules/core/src/main/java/org/apache/ignite/configuration/PlatformNearCacheConfiguration.java
ptupitsyn: Agree with @isapego . Another thing is that we can't (trivially) avoid this, because configuration needs to be sent across entire cluster, so this has to be a part of `CacheConfiguration`.

Relevant dev list thread: http://apache-ignite-developers.2346864.n4.nabble.com/NET-Near-Cache-new-flag-in-NearCacheConfiguration-java-td45961.html

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCache.cs
ptupitsyn: Yes, I mentioned no-boxing approach regarding `_map` when proper types are specified in `PlatformNearCacheConfiguration`.

This line is just a cast from `TV` to `TVal`, which can't be done any other way. Previously we used `TypeCaster` class to avoid boxing in such cases (see usages), and it was proved to be better on .NET 4. However, this is no longer the case on .NET Core:

```
Runtime=.NET Core 3.1  

|     Method |      Mean |     Error |    StdDev | Gen 0 | Gen 1 | Gen 2 | Allocated |
|----------- |----------:|----------:|----------:|------:|------:|------:|----------:|
| TypeCaster | 1.5402 ns | 0.0138 ns | 0.0122 ns |     - |     - |     - |         - |
| ObjectCast | 0.0284 ns | 0.0098 ns | 0.0087 ns |     - |     - |     - |         - |
```

(source: https://github.com/ptupitsyn/IgniteNetBenchmarks/blob/master/CasterBenchmark.cs)

As you can see, such a cast is optimized by JIT, there are zero allocations, and it is faster than `TypeCaster`.

(There is a ticket to get rid of TypeCaster: https://issues.apache.org/jira/browse/IGNITE-12699)

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCache.cs
ptupitsyn: Unfortunately, we can't. Because nothing is invoked on Java side in this case. What happens is:
* Affinity topology version changes (node enters or leaves, etc)
* `GridNearCacheEntry` becomes effectively invalid, because updates are no longer received (primary node changed for the key).
* Data becomes stale, but nothing is being done until user requests this entry
* When user calls cache.Get, `GridNearCacheEntry.valid()` is called, invalid entry is detected, etc

So we have to have the same `valid()` check on .NET side to avoid calling Java on every `Get`

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/Near/NearCacheManager.cs
ptupitsyn: I think that compared to the whole "thick client connect/disconnect" procedure, this call is very cheap.

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Cache/CacheImpl.cs
ptupitsyn: Agree, fixed. No exceptions now.

modules/platforms/dotnet/Apache.Ignite.Core/Cache/Configuration/CacheConfiguration.cs
ptupitsyn: @ashapkin can you please provide a full reproducer?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/PlatformContextImpl.java
ptupitsyn: Yes, when .NET performs `Put`, we already have a value on .NET side, no need to send it back through the callback. 

Before this TLS optimization we had a **50%** perf drop on `Put` when Platform Near is enabled. After TLS optimization the drop is about **7%**. I think this added complexity is worth it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtCacheEntry.java
ptupitsyn: Unnecessary change, reverted.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/PlatformContextImpl.java
isapego: Cool

modules/platforms/dotnet/Apache.Ignite.Core/Impl/Common/CopyOnWriteConcurrentDictionary.cs
```diff
@@ -89,6 +89,30 @@ public void Set(TKey key, TValue value)
...
+        /// </summary>
+        /// <param name="key">The key.</param>
+        /// <param name="val">Removed value, if any.</param>
+        public bool Remove(TKey key, out TValue val)
+        {
+            lock (this)
+            {
+                var dict0 = new Dictionary<TKey, TValue>(_dict);
+
+                if (!dict0.TryGetValue(key, out val))
```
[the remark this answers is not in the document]
ptupitsyn: Good catch, fixed

PR #895, opened 2020-02-03
GG-26790 Prevent BinaryObjectExImpl.toString failing with exception if a class of an object serialized with OptimizedMarshaller is not accessible

pavlukhin:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/ivan/BinaryToStringClassCastProblem.java
```diff
@@ -0,0 +1,112 @@
...
+        cache.put(1, new TestContainer(ext));
+
+        assertTrue(cache.withKeepBinary().get(1).toString()
+            .matches("org.apache.ignite.internal.processors.cache.ivan.BinaryToStringClassCastProblem\\$TestContainer " +
+                "\\[idHash=-?\\d+, hash=-?\\d+, x=Problematic\\(Class not found\\)]"));
+    }
+
+    /** */
+    private URLClassLoader extClassLoader(String srcFileName, String srcCode) throws IOException {
+        // t0d0 cleanup tmp files carefully
```
akuznetsov-gridgain: Such "todo" looks like a violation of code style for me.
We should use the following format: `// TODO GG-XXXXX cleanup tmp files carefully`.
Where GG-XXXXX - is a number of issue where this will be fixed.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/ivan/BinaryToStringClassNotFoundProblem.java
```diff
@@ -0,0 +1,112 @@
...
+        cache.put(1, new TestContainer(ext));
+
+        assertTrue(cache.withKeepBinary().get(1).toString()
+            .matches("org.apache.ignite.internal.processors.cache.ivan.BinaryToStringClassCastProblem\\$TestContainer " +
+                "\\[idHash=-?\\d+, hash=-?\\d+, x=Problematic\\(Class not found\\)]"));
+    }
+
+    /** */
+    private URLClassLoader extClassLoader(String srcFileName, String srcCode) throws IOException {
+        // t0d0 cleanup tmp files carefully
```
akuznetsov-gridgain: Such "todo" looks like a violation of code style for me.
We should use the following format: // TODO GG-XXXXX cleanup tmp files carefully.
Where GG-XXXXX - is a number of issue where this will be fixed.

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryObjectExImpl.java
```diff
@@ -240,6 +242,23 @@ private String toString(BinaryReaderHandles ctx, IdentityHashMap<BinaryObject, I
...
+            return field(ctx, name);
+        }
+        catch (Exception e) {
+            OptimizedMarshallerInaccessibleClassException e1 =
+                X.cause(e, OptimizedMarshallerInaccessibleClassException.class);
+
+            String msg = "Failed to create a string representation";
+
+            return e1 != null
+                ? "(" + msg + ": class not found " + e1.inaccessibleClass() + ")"
```
MaxLosevskoy: No `;` for this case?

MaxLosevskoy:
```java
? "(" + msg + ": class " + e1.inaccessibleClass() + " not found);"
```

pavlukhin: I suppose there is no `;` in both cases. There are 2 options for a message here:
* `(Failed to create a string representation: class not found ExternalTestClass1)`
* `(Failed to create a string representation)`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/ivan/BinaryToStringClassNotFoundProblem.java
pavlukhin: I cleaned up test code.

PR #897, opened 2020-02-04
GG-27348 Implemented support for Ent/Ult agent.

akuznetsov-gridgain:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/action/controller/SecurityActionsController.java
```diff
@@ -48,7 +48,7 @@
     public SecurityActionsController(GridKernalContext ctx) {
         this.ctx = ctx;
         this.log = ctx.log(SecurityActionsController.class);
-        this.registry = ((ManagementConsoleProcessor)ctx.managementConsole()).sessionRegistry();
+        this.registry = ((ManagementConsoleAgent)ctx.managementConsole()).sessionRegistry();
```
nva: Why we need this cast. Can we add this method to interface?

modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/action/DistributedActionProcessor.java
```diff
@@ -61,9 +61,9 @@
     public DistributedActionProcessor(GridKernalContext ctx) {
         super(ctx);
 
-        this.mgr = ((ManagementConsoleProcessor)ctx.managementConsole()).webSocketManager();
-        this.sesRegistry = ((ManagementConsoleProcessor)ctx.managementConsole()).sessionRegistry();
-        this.locNodeConsistentId = String.valueOf(ctx.grid().localNode().consistentId());
+        mgr = ((ManagementConsoleAgent)ctx.managementConsole()).webSocketManager();
```
nva: Why we need this cast. Can we add this method to interface?

modules/management-console-agent/src/main/java/org/apache/ignite/agent/action/controller/SecurityActionsController.java
akuznetsov-gridgain: No, this is implementation details so far.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/action/DistributedActionProcessor.java
akuznetsov-gridgain: No, this is implementation details so far.

PR #900, opened 2020-02-05
GG-27417: Fixed C++ documentation

isapego:


modules/platforms/cpp/cpp.dxg
```diff
@@ -1719,4 +1719,4 @@ DOT_CLEANUP            = YES
 ;EXCLUDE=core/include/ignite/impl core/os/linux/include/ignite/impl core/os/linux/src/impl core/os/win/include/ignite/impl core/os/win/src/impl core/src/impl binary/include/ignite/impl binary/src/impl common/include/ignite/common common/os thin-client/include/ignite/impl thin-client/src
 ;STRIP_FROM_PATH=core/include/ignite core/src binary/include/ignite binary/src thin-client/src
 ;OUTPUT_DIRECTORY=../../clients/target/cppdoc
-;PROJECT_LOGO=../../../assembly/docfiles/ignite_logo.png
+;PROJECT_LOGO=../dotnet/logo_gg_64x64.png
```
ptupitsyn: Let's create a copy of this file, or maybe use a larger one for the docs. Using a file from dotnet folder is error-prone, I would not expect outer folders to have dependencies on it.

isapego: Agree

PR #918, opened 2020-02-11
GG-26852 Allow custom authenticators to use SSL certificates

alamar:


modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/odbc/OdbcConnectionContext.java
```diff
@@ -117,7 +117,8 @@ public OdbcConnectionContext(GridKernalContext ctx, GridNioSession ses, GridSpin
     }
 
     /** {@inheritDoc} */
-    @Override public void initializeFromHandshake(ClientListenerProtocolVersion ver, BinaryReaderExImpl reader)
+    @Override public void initializeFromHandshake(GridNioSession ses,
```
isapego: If we pass session instance here then we don't need it a as class field anymore. Same for the JDBC.

alamar: Will do.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioSessionImpl.java
```diff
@@ -269,6 +273,22 @@ public void reset0() {
...
 
+    /** */
+    @Override public Certificate[] certificates() {
+        GridSslMeta meta = meta(SSL_META.ordinal());
+
+        if (meta != null) {
+            try {
+                return meta.sslEngine().getSession().getPeerCertificates();
+            }
+            catch (SSLPeerUnverifiedException e) {
```
ptupitsyn: Not sure, but we probably want to pass that exception along, not swallow it. Let the caller deal with it.

ptupitsyn (approved):

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioSessionImpl.java
alamar: This is expected exception - can be thrown in perfectly legitimate scenarios.

PR #940, opened 2020-02-20
GG-27202

ygerzhedovich:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -190,29 +182,44 @@ public QueryMemoryManager(GridKernalContext ctx) {
...
+            }
 
-        if (globalQuota0 > 0 && globalQuota0 < maxQueryMemory) {
-            U.warn(log, "Max query memory can't exceed SQL memory pool size. Will be reduced down to: " + globalQuota);
+            return null;
+        }
+
+        if (globalQuota0 > 0 && globalQuota0 < maxQryMemory) {
+            if (log.isDebugEnabled()) {
+                log.debug( "Max query memory can't exceed SQL memory pool size. Will be reduced down to: " +
```
ygerzhedovich: WDYT, may be better to have `info` level for the case?

ygerzhedovich: Also potentially will be useful  to print the same warning for `setQueryQuota` method.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -396,10 +421,15 @@ public GroupedExternalResult createGroupedExternalResult(Session ses, int size)
...
-        if (!ses.isOffloadedToDisk()) {
-            ses.setOffloadedToDisk(true);
+        IgniteTrace trace = ses.getIgniteTrace();
+
+        if (!trace.isOffloadStarted()) {
+            trace.offloadStarted(true);
 
             metrics.trackQueryOffloaded();
+
+            if (log.isDebugEnabled())
```
ygerzhedovich: I think will be better to log it with info level.

modules/h2/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteTrace.java
```diff
@@ -0,0 +1,157 @@
...
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+package org.apache.ignite.internal.processors.query.h2;
+
+/**
+ * Tracing object for SQL queries.
+ */
+public class IgniteTrace {
```
korlov42: why do we need one more holder for metrics? There is already QueryMemTracker. Since it is created on per query basis, we could just simply extend it with some query description. IMO it should be enough

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
rkondakov: I set info level with throttling and also added a warning on quota set up.

modules/h2/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteTrace.java
rkondakov: I wouldn't say I am absolutely agree with your point, but ok :)

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -130,10 +127,12 @@
     public QueryMemoryManager(GridKernalContext ctx) {
         this.log = ctx.log(QueryMemoryManager.class);
         this.ctx = ctx;
+
+        setGlobalQuota(ctx.config().getSqlGlobalMemoryQuota());
```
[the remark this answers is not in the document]
rkondakov: Because we need to throw exception if global quota is set more than heap size. See next line.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2QueryInfo.java
```diff
@@ -139,6 +139,19 @@ public void printLogMessage(IgniteLogger log, ConnectionManager connMgr, String
...
+    /**
+     * @return Query description.
+     */
+    public String buildShortQueryInfoString() {
+        return "[type=" + type +
+            ", distributedJoin=" + distributedJoin +
+            ", enforceJoinOrder=" + enforceJoinOrder +
+            ", lazy=" + lazy +
+            ", schema=" + schema +
+            ", sql='" + sql +
```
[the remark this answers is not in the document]
rkondakov: There is no data here. Only the query and it's params.

PR #941, opened 2020-02-20
GG-27697: Fix lazy query nightly suite for memory quota and offloading.

rkondakov:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/DiskSpillingAbstractTest.java
```diff
@@ -488,26 +490,32 @@ protected void checkQuery(Result res, String sql, int threadNum, int iterations)
...
-                    grid.cache(DEFAULT_CACHE_NAME)
-                        .query(new SqlFieldsQuery(sql)
-                            .setLazy(false))
-                        .getAll();
+                try {
+                    for (int i = 0; i < iterations; i++) {
+                        IgniteEx grid = fromClient() ? grid("client") : grid(0);
+                        grid.cache(DEFAULT_CACHE_NAME)
+                            .query(new SqlFieldsQuery(sql)
+                                .setLazy(true))
```
korlov42: maybe we should make laziness as param and run tests in two different configuration, WDYT?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/DiskSpillingDmlTest.java
```diff
@@ -302,7 +317,7 @@ private void testInsert(String insertDml) throws IOException {
      */
     private List<List<?>> runSql(String sql) {
         return grid(0).cache(DEFAULT_CACHE_NAME)
-            .query(new SqlFieldsQuery(sql))
+            .query(new SqlFieldsQuery(sql).setLazy(true))
```
korlov42: same question as above

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/DiskSpillingAbstractTest.java
rkondakov: It makes sense. We have a nightly suite `IgniteBinaryCacheQueryLazyTestSuite2` where all queries tests are checked with lazy=true settings. For now, I just removed this setter.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/oom/DiskSpillingDmlTest.java
rkondakov: See above.

PR #945, opened 2020-02-21
GG-26047 Added support of decimal inlining

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/BPlusMetaIO.java
```diff
@@ -305,12 +318,14 @@ private long flags(long pageAddr) {
      * @param pageAddr Page address.
      * @param unwrappedPk unwrapped primary key of this tree flag.
      * @param inlineObjSupported inline POJO by created tree flag.
+     * @param inlineDcSupported inline decimal by created tree flag.
      */
-    public void setFlags(long pageAddr, boolean unwrappedPk, boolean inlineObjSupported) {
+    public void setFlags(long pageAddr, boolean unwrappedPk, boolean inlineObjSupported, boolean inlineDcSupported) {
```
ygerzhedovich: abbreviation DC looks bad

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/BPlusMetaIO.java
```diff
@@ -328,10 +343,11 @@ public void setFlags(long pageAddr, boolean unwrappedPk, boolean inlineObjSuppor
     /**
      * @param pageAddr Page address.
      * @param inlineObjSupported Supports inline object flag.
+     * @param inlineDcSupported Supports decimal flag.
      * @param unwrappedPk Unwrap PK flag.
      * @param pageSize Page size.
      */
-    public static void upgradePageVersion(long pageAddr, boolean inlineObjSupported, boolean unwrappedPk, int pageSize) {
+    public static void upgradePageVersion(long pageAddr, boolean inlineObjSupported, boolean inlineDcSupported, boolean unwrappedPk, int pageSize) {
```
ygerzhedovich: is here any place or case when `inlineDcSupported` could be true? If asnwer is NO, seems we could remove the parameter at all.

modules/h2/src/main/java/org/h2/value/ValueDecimal.java
```diff
@@ -262,4 +263,15 @@ public static BigDecimal setScale(BigDecimal bd, int scale) {
         return bd.setScale(scale, RoundingMode.HALF_UP);
     }
 
+    /**
+     * Get cached float value or calculate using BigDecimal and cache it.
+     *
+     * @return float representation of BidDecimal
```
ygerzhedovich: BidDecimal -> BigDecimal

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -243,12 +244,27 @@ public H2Tree(
 
             boolean inlineObjSupported = inlineSize > 0 && inlineObjectSupported(metaInfo, inlineIdxs0);
 
-            inlineIdxs = inlineObjSupported ? inlineIdxs0 : inlineIdxs0.stream()
-                .filter(ih -> ih.type() != Value.JAVA_OBJECT)
-                .collect(Collectors.toList());
+            boolean inlineDcSupported = inlineSize > 0 && metaInfo.flagsSupported() && metaInfo.inlineDecimalSupported();
+
+            if (!inlineObjSupported || !inlineDcSupported) {
```
ygerzhedovich: hard to understand piece of code. Let's add some comments with explanation.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -243,12 +244,27 @@ public H2Tree(
...
+                while (it.hasNext()) {
+                    InlineIndexColumn ih = it.next();
+
+                    if (!inlineDcSupported && ih.type() == Value.DECIMAL)
+                        dc = true;
+
+                    if (dc || (!inlineObjSupported && ih.type() == Value.JAVA_OBJECT))
+                        it.remove();
+                }
+            }
```
ygerzhedovich: hard to understand piece of code. Let's add some comments with explanation.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -446,9 +462,10 @@ private MetaPageInfo getMetaInfo() throws IgniteCheckedException {
      * and created product version on root meta page).
      *
      * @param inlineObjSupported inline POJO by created tree flag.
+     * @param inlineDcSupported inline decimal by created tree flag.
      * @throws IgniteCheckedException On error.
      */
-    private void upgradeMetaPage(boolean inlineObjSupported) throws IgniteCheckedException {
+    private void upgradeMetaPage(boolean inlineObjSupported, boolean inlineDcSupported) throws IgniteCheckedException {
```
ygerzhedovich: let's remove redundant new `inlineDcSupported` parameter

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/database/inlinecolumn/InlineIndexColumnTest.java
```diff
@@ -488,6 +490,20 @@ public void testJavaObjectInlineHash() throws Exception {
         }
     }
 
+    /** */
+    @Test
+    public void testDecimal() throws Exception {
+        int maxSize = 1 + 4; // 1 byte header + 4 bytes value.
+
+        assertEquals(-2, putAndCompare(new DecimalFloat(123456.7f), new DecimalFloat(123456.7f), DecimalFloat.class, maxSize));
```
mshonichev: use `CANT_BE_COMPARE` instead of `-2`, please

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -243,12 +244,27 @@ public H2Tree(
...
+                    if (dc || (!inlineObjSupported && ih.type() == Value.JAVA_OBJECT))
+                        it.remove();
+                }
+            }
+
+            inlineIdxs = inlineIdxs0;
 
             if (!metaInfo.flagsSupported())
-                upgradeMetaPage(inlineObjSupported);
+                upgradeMetaPage(inlineObjSupported, false);
```
mshonichev: don't understand, why always `false` here ?

modules/yardstick/src/main/java/org/apache/ignite/yardstick/sql/IgniteInlineIndexBenchmark.java
```diff
@@ -110,16 +145,29 @@ private void init() {
...
-        return isJavaObj
-            ? new TestKey(ThreadLocalRandom.current().nextInt(range))
-            : ThreadLocalRandom.current().nextInt(range);
+        switch (testedType) {
+            case JAVA_OBJECT:
+                return new TestKey(ThreadLocalRandom.current().nextInt(range));
+            case PRIMITIVE:
+                return ThreadLocalRandom.current().nextInt(range);
+            case DECIMAL:
+                return BigDecimal.valueOf(ThreadLocalRandom.current().nextDouble(range));
```
korlov42: we should limit amount of different keys, not its upper bound. Otherwise the test will fail from time to time since memory limit is set for default data region

modules/h2/src/main/java/org/h2/value/ValueDecimal.java
```diff
@@ -262,4 +263,15 @@ public static BigDecimal setScale(BigDecimal bd, int scale) {
...
+     */
+    public float getFloatValue() {
+        if (valueFloat == null)
+            valueFloat = getBigDecimal().floatValue();
+        return valueFloat;
```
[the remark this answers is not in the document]
korlov42: please check style here and below

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/database/inlinecolumn/InlineIndexColumnTest.java
```diff
@@ -488,6 +490,20 @@ public void testJavaObjectInlineHash() throws Exception {
         }
     }
 
+    /** */
+    @Test
+    public void testDecimal() throws Exception {
```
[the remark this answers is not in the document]
RealZheka: I don't test custom values, only BigDecimal. PutAndCompare test function uses typeByClass(Class<?> cls) function that creates h2 Value based on provided class type. Float.class is already mapped to Value.FLOAT, so I need to create float class wrapper to create Value.DECIMAL. Same was done for String types previously.

Fixed other comments.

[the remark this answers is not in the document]
korlov42: But why not just use the BigDecimal class? 

There was the reason to create wrappers for string: we need to distinguish 3 types of the string, but jvm has only one type.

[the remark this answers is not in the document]
korlov42: by the way do you check values that different in digit beyond the size of float?

PR #947, opened 2020-02-21
GG-27582: SQL: EnforceJoinOrder flag not working for cached local queries.

mshonichev:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/H2StatementCache.java
```diff
@@ -102,4 +101,22 @@ void remove(String schemaName, String sql) {
...
+
+        return queryFlags(qryDesc.distributedJoins(), qryDesc.enforceJoinOrder());
+    }
+
+    /** */
+    public static byte queryFlags(
+        boolean distributedJoins,
+        boolean enforceJoinOrder
+        ) {
+        return (byte)(1 +
```
korlov42: why do lack of flags represented by 1, not by 0?

tledkov-gridgain: Great issue. It's an artifact of the old code. I'll change it.

PR #951, opened 2020-02-25
GG-27066 Forward-port volatile data region improvements from 8.7

vldpyatkov:
Always start and use volatile data region, since only a subset of nodes may be persistent.
Rename data region to volatileDsMemPlc.


modules/core/src/main/java/org/apache/ignite/internal/processors/datastructures/DataStructuresProcessor.java
```diff
@@ -362,8 +361,9 @@ public static boolean isDataStructureCache(String cacheName) {
      * @return {@code True} if group name is reserved to store data structures.
      */
     public static boolean isReservedGroup(@Nullable String grpName) {
-        return DEFAULT_DS_GROUP_NAME.equals(grpName) ||
-            DEFAULT_VOLATILE_DS_GROUP_NAME.equals(grpName);
+        return grpName != null &&
+            (DEFAULT_DS_GROUP_NAME.equals(grpName) ||
+            grpName.startsWith(DEFAULT_VOLATILE_DS_GROUP_NAME));
```
vldpyatkov: Need to replace DEFAULT_VOLATILE_DS_GROUP_NAME.equals(grpName) here

PR #956, opened 2020-02-26
Gg 27745

ingvard:


modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/PartitionReconciliationCancel.java
```diff
@@ -34,7 +33,7 @@
 public class PartitionReconciliationCancel implements Command<Void> {
     /** {@inheritDoc} */
     @Override public void printUsage(Logger log) {
-        String desc = "Does cancel of partition reconciliation command.";
+        String desc = "Cancels of partition reconciliation command.";
```
glukos: Cancels partition reconciliation command

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/AffectedEntryResult.java
```diff
@@ -37,71 +37,71 @@
 /**
  * Result with lists of broken and fixed, skipped entries.
  */
-public class PartitionReconciliationResult extends IgniteDataTransferObject {
+public class AffectedEntryResult extends IgniteDataTransferObject {
```
glukos: Maybe AffectedEntriesResult or ReconciliationAffectedEntries would be better?
Looks like this class encloses just one entry.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/VersionedEntry.java
```diff
@@ -20,89 +20,99 @@
...
+    @Override protected void readExternalData(byte protoVer,
+        ObjectInput in) throws IOException, ClassNotFoundException {
+        key = (VersionedKey)in.readObject();
+        val = (VersionedValue)in.readObject();
+    }
+
+    /**
+     * @return Key of this entry.
+     */
+    public KeyCacheObject getKey() {
```
glukos: In GG code style there's no "get" in getters, isn't it?
I don't see this as a must, but some DTO has "gets" and some hasn't. Looks misleading.

modules/core/src/main/java/org/apache/ignite/internal/visor/checker/VisorPartitionReconciliationCancelTask.java
```diff
@@ -33,7 +33,7 @@
     private static final long serialVersionUID = 0L;
 
     /**
-     *
+     * If the session is set to zero value, reconciliation interrupts.
```
glukos: reconciliation is interrupted

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ExecutionResult.java
```diff
@@ -53,41 +53,41 @@ public ExecutionResult() {
...
+    public ExecutionResult(String errorMsg) {
+        this.errorMsg = errorMsg;
     }
 
     /**
      *
      */
-    public String getErrorMessage() {
-        return errorMessage;
+    public String getErrorMsg() {
```
glukos: Do we have to use abbreviated names for getters?

PR #957, opened 2020-02-28
GG-23552 Add DR clearStore task to Web Console

nva:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -115,6 +115,9 @@
     /** */
     MANAGEMENT_CONSOLE(28),
 
+    /** Cluster has task to clear sender store. */
+    WC_DR_CLEAR_SENDER_STORE(29),
+
```
vsisko: We should check that gridgain-core library in classpath.

nva: Why and how we can check this in ignite-core module?
This ticket contains changes in ggprivate repository with tasks

akuznetsov-gridgain: We should check DR plugin. I think it can be done via classForName() ?

nva: We can't add checking of gridgain-core class in gridgain-ce code. In same time we has additional checks for that in our frontend/web-console code

PR #958, opened 2020-02-28
Gg 23528

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -222,13 +225,18 @@ public int execute(List<String> rawArgs) {
...
+        boolean printErrStackTrace = false;
+
         try {
             if (F.isEmpty(rawArgs) || (rawArgs.size() == 1 && CMD_HELP.equalsIgnoreCase(rawArgs.get(0)))) {
                 printHelp();
 
                 return EXIT_CODE_OK;
             }
 
+            printErrStackTrace = rawArgs.stream().anyMatch(CMD_PRINT_ERR_STACK_TRACE::equals);
```
vldpyatkov: May be more proper use F.exist(rawArgs, (arg) -> CMD_PRINT_ERR_STACK_TRACE.equals(CMD_PRINT_ERR_STACK_TRACE))

vldpyatkov: I think to use parser instead of go through console args manually.

modules/core/src/test/resources/org.apache.ignite.util/control.sh_help.output
```diff
@@ -4,7 +4,7 @@ User: <!any!>
...
 
-  control.(sh|bat) [--host HOST_OR_IP] [--port PORT] [--user USER] [--password PASSWORD] [--ping-interval PING_INTERVAL] [--ping-timeout PING_TIMEOUT] [--ssl-protocol SSL_PROTOCOL[, SSL_PROTOCOL_2, ..., SSL_PROTOCOL_N]] [--ssl-cipher-suites SSL_CIPHER_1[, SSL_CIPHER_2, ..., SSL_CIPHER_N]] [--ssl-key-algorithm SSL_KEY_ALGORITHM] [--keystore-type KEYSTORE_TYPE] [--keystore KEYSTORE_PATH] [--keystore-password KEYSTORE_PASSWORD] [--truststore-type TRUSTSTORE_TYPE] [--truststore TRUSTSTORE_PATH] [--truststore-password TRUSTSTORE_PASSWORD] [command] <command_parameters>
+  control.(sh|bat) [--host HOST_OR_IP] [--port PORT] [--user USER] [--password PASSWORD] [--ping-interval PING_INTERVAL] [--ping-timeout PING_TIMEOUT] [--print-err-stack-trace PRINT_ERR_STACK_TRACE] [--ssl-protocol SSL_PROTOCOL[, SSL_PROTOCOL_2, ..., SSL_PROTOCOL_N]] [--ssl-cipher-suites SSL_CIPHER_1[, SSL_CIPHER_2, ..., SSL_CIPHER_N]] [--ssl-key-algorithm SSL_KEY_ALGORITHM] [--keystore-type KEYSTORE_TYPE] [--keystore KEYSTORE_PATH] [--keystore-password KEYSTORE_PASSWORD] [--truststore-type TRUSTSTORE_TYPE] [--truststore TRUSTSTORE_PATH] [--truststore-password TRUSTSTORE_PASSWORD] [command] <command_parameters>
```
vldpyatkov: Not need parameter to set for print stack trace argument
[--print-err-stack-trace]

modules/core/src/test/resources/org.apache.ignite.util/control.sh_cache_help.output
```diff
@@ -7,7 +7,7 @@ Arguments: --cache help --yes
...
+  control.(sh|bat) [--host HOST_OR_IP] [--port PORT] [--user USER] [--password PASSWORD] [--ping-interval PING_INTERVAL] [--ping-timeout PING_TIMEOUT] [--print-err-stack-trace PRINT_ERR_STACK_TRACE] [--ssl-protocol SSL_PROTOCOL[, SSL_PROTOCOL_2, ..., SSL_PROTOCOL_N]] [--ssl-cipher-suites SSL_CIPHER_1[, SSL_CIPHER_2, ..., SSL_CIPHER_N]] [--ssl-key-algorithm SSL_KEY_ALGORITHM] [--keystore-type KEYSTORE_TYPE] [--keystore KEYSTORE_PATH] [--keystore-password KEYSTORE_PASSWORD] [--truststore-type TRUSTSTORE_TYPE] [--truststore TRUSTSTORE_PATH] [--truststore-password TRUSTSTORE_PASSWORD] --cache [subcommand] <subcommand_parameters>
```
vldpyatkov: Only --print-err-stack-trace would be enough for printing stack on log.

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
tkalkirill: Remade to F#exist. 
In the beginning I did through the parser, but then I realized that while parsing can also appear error, stack where you want to watch, so I made also the argument "--print-err-stack-trace" may be at the end and it not get.

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -319,34 +331,45 @@ public int execute(List<String> rawArgs) {
...
             return EXIT_CODE_UNEXPECTED_ERROR;
         }
         finally {
             LocalDateTime endTime = LocalDateTime.now();
 
             Duration diff = Duration.between(startTime, endTime);
 
+            if (printErrStackTrace && nonNull(err)) {
+                logger.info("Error stack trace:");
+                logger.info(X.getFullStackTrace(err));
```
vldpyatkov: Does this logger have an issue with printing several lines?
Like this manner:
logger.info("Error stack trace:\n" +
  X.getFullStackTrace(err));

modules/core/src/main/java/org/apache/ignite/internal/commandline/CommandHandler.java
```diff
@@ -222,13 +225,18 @@ public int execute(List<String> rawArgs) {
 
         String commandName = "";
```
vldpyatkov: Excess empty line.

tkalkirill: I think there should be an empty string here.

PR #962, opened 2020-03-02
GG-27538 Fix for KILL QUERY command hanging.

rkondakov:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryCancel.java
```diff
@@ -16,65 +16,52 @@
...
 
-            if (STATE_UPDATER.compareAndSet(this, tmp, CANCELLED)) {
-                if (tmp != null)
-                    tmp.doCancel();
+        canceled = true;
 
-                return;
-            }
-        }
+        for (QueryCancellable action : cancelActions)
```
ygerzhedovich: what is first `doCancel()` throw an Exception? Should we use try-catch block here to try execute next cancels?

PR #963, opened 2020-03-02
GG-27866 Implement waiting for backups.

alamar:


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1343,6 +1344,10 @@
      */
     public static final String IGNITE_ENABLE_EXTRA_INDEX_REBUILD_LOGGING = "IGNITE_ENABLE_EXTRA_INDEX_REBUILD_LOGGING";
 
+    /** */
```
sanpwc: Could you please clarify what "IGNITE_WAIT_FOR_BACKUPS_ON_SHUTDOWN" means?

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -289,15 +298,20 @@ public static IgniteState state(@Nullable String name) {
...
-     *      up to the actual job to exit from execution
+     * @param cancel If {@code true} then all jobs currently will be cancelled
+     *      by calling {@link ComputeJob#cancel()} method. Note that just like with
+     *      {@link Thread#interrupt()}, it is up to the actual job to exit from
+     *      execution. If {@code false}, then jobs currently running will not be
+     *      canceled. In either case, grid node will wait for completion of all
+     *      jobs running on it before stopping. If {@code null} then
+     *      not-{@link IgniteConfiguration#isGracefulShutdown()} is used instead.
+     * @param waitForBackups If {@code true} then node will wait until all of its data is backed up before shutting
+     *      down. If {@code null} then {@link IgniteConfiguration#isWaitForBackupsOnShutdown()} is used instead.
```
sanpwc: I don't like an idea of considering null as IgniteConfiguration#isWaitForBackupsOnShutdown(). Is it possible to use `IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_WAIT_FOR_BACKUPS_ON_SHUTDOWN)` instead of null within methods like `org.apache.ignite.Ignition#stop(boolean)`?

sanpwc: `IgniteConfiguration#isWaitForBackupsOnShutdown()` there's no such method.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2588,6 +2633,82 @@ private synchronized void stop0(boolean cancel) {
                     if (log != null && log.isDebugEnabled())
                         log.debug("Shutdown is in progress (ignoring): " + e.getMessage());
                 }
+            }
+
+            if (waitForBackups) {
```
sanpwc: Will be reworked as part of GG-27955

modules/core/src/test/java/org/apache/ignite/testframework/junits/multijvm/IgniteProcessProxy.java
```diff
@@ -215,7 +215,7 @@ public IgniteProcessProxy(
      * @return Initiated logger.
      * @throws Exception In case of an error.
```
sanpwc: Could you please also update javadoc?

modules/core/src/test/java/org/apache/ignite/testframework/junits/multijvm/IgniteProcessProxy.java
```diff
@@ -225,7 +225,7 @@ protected IgniteLogger logger(IgniteLogger log, Object ctgr) throws Exception {
      * @return Node runner class name.
      * @throws Exception In case of an error.
```
sanpwc: Could you please also update javadoc?

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
alamar: This is public API which is planned for phase II, I will try to disable these comments.

alamar: @sanpwc This is private API so I think it is tolerable.
If we require that IgniteSystemProperties.IGNITE_WAIT_FOR_BACKUPS_ON_SHUTDOWN is supplied explicitly by caller, this means that e.g. FailureHandler implementations have to compute these properties' values.
Ignition API will be clear without nulls when we add it in Phase II. WDYT?

PR #978, opened 2020-03-06
GG-27306

akalash:


modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationInverseConnectionEstablishingTest.java
```diff
@@ -0,0 +1,216 @@
...
+import static org.apache.ignite.cache.CacheAtomicityMode.ATOMIC;
+import static org.apache.ignite.cache.CacheAtomicityMode.TRANSACTIONAL;
+import static org.apache.ignite.cache.CacheWriteSynchronizationMode.FULL_SYNC;
+
+/**
+ *
+ */
+public class GridTcpCommunicationInverseConnectionEstablishingTest extends GridCommonAbstractTest {
+    /** */
+    private static final IgnitePredicate<ClusterNode> CLIENT_PRED = c -> c.isClient();
```
ibessonov: Can we inline this predicate? Its single usage looks strange.

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationInverseConnectionEstablishingTest.java
```diff
@@ -0,0 +1,216 @@
...
+        for (int i = 0; i < srvs; i++) {
+            ccfgs = cacheConfigurations1();
+
+            startGrid(i);
+        }
+
+        clientMode = true;
+        this.envType = envType;
+
+        startGrid(srvs);
```
ibessonov: You can call "startClientGrid" and check configuration instead of field. So the field could be removed.

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationInverseConnectionEstablishingTest.java
```diff
@@ -0,0 +1,216 @@
...
+            }
+
+            return super.createTcpClient(node, connIdx);
+        }
+
+        /**
+         * @param name Name.
+         */
+        private String createAttributeName(String name) {
+            return getClass().getSimpleName() + '.' + name;
```
ibessonov: Does it work? I mean, class name is not "TcpCommunicationSpi" anymore.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -418,6 +433,13 @@ public void resetMetrics() {
         });
     }
 
+    /** {@inheritDoc} */
+    @Override public IgniteInternalFuture<?> onReconnected(boolean clusterRestarted) throws IgniteCheckedException {
+        locNodeId = ctx.localNodeId();
```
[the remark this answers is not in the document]
sergey-chugunov-1985: I confirmed that we should update nodeId on reconnected event, as at the moment of DISCONNECTED notification nodeId is not updated yet.
Take a look at new test: org.apache.ignite.internal.processors.cache.distributed.IgniteCacheClientReconnectTest#testClientIdUpdateOnReconnect

PR #979, opened 2020-03-10
GG-27707 First version

ingvard:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -3664,6 +3667,11 @@ else if (rcvCnt < 0)
                     if (log.isDebugEnabled())
                         log.debug("Client creation failed [addr=" + addr + ", err=" + e + ']');
 
+                    if (hasCause(e, "Too many open files", SocketException.class)) {
+                        if (IgniteEx.class.isInstance(ignite))
+                            ((IgniteEx)ignite).context().failure().process(new FailureContext(CRITICAL_ERROR, e));
```
akalash: I urge not to do such error handling because first of all, it is the leak of abstraction, secondly, it can be easily broken by future updates. Instead of it, please, implement your custom critical exception which won't be caught in TcpCommunicationSpi#reserveClient.

PR #981, opened 2020-03-10
GG-27952 Fixed erroneously triggered partitions validation for a cach…

sk0x50:
…e group.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
@@ -3819,8 +3819,11 @@ public class GridDhtPartitionsExchangeFuture extends GridDhtTopologyFutureAdapte
 
                     // Do not validate read or write through caches or caches with disabled rebalance
                     // or ExpiryPolicy is set or validation is disabled.
-                    boolean eternalExpiryPolicy = grpCtx != null && (grpCtx.config().getExpiryPolicyFactory() == null
-                        || grpCtx.config().getExpiryPolicyFactory().create() instanceof EternalExpiryPolicy);
+                    boolean customExpiryPolicy = Optional.ofNullable(grpCtx)
+                        .map((v) -> v.caches())
+                        .orElseGet(() -> Collections.emptyList())
+                        .stream()
+                        .anyMatch(ctx -> ctx.expiry() != null && !(ctx.expiry() instanceof EternalExpiryPolicy));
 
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -3819,16 +3819,19 @@ private void validatePartitionsState() {
 
                     // Do not validate read or write through caches or caches with disabled rebalance
                     // or ExpiryPolicy is set or validation is disabled.
-                    boolean eternalExpiryPolicy = grpCtx != null && (grpCtx.config().getExpiryPolicyFactory() == null
-                        || grpCtx.config().getExpiryPolicyFactory().create() instanceof EternalExpiryPolicy);
+                    boolean customExpiryPolicy = Optional.ofNullable(grpCtx)
+                        .map((v) -> v.caches())
```
ingvard: Please use java 8 style -> CacheGroupContext::caches, Collections::emptyList

modules/core/src/test/java/org/apache/ignite/internal/DiagnosticLogForPartitionStatesTest.java
```diff
@@ -105,20 +166,33 @@ private void doTest(CacheConfiguration<Object, Object> cacheCfg, boolean msgExp)
 
         node1.cluster().active(true);
 
-        IgniteCache<Object, Object> cache = node1.getOrCreateCache(cacheCfg);
+        List<IgniteCache<Integer, Integer>> caches = new ArrayList<>(cfgs.size());
```
ingvard: Try to use streams everywhere:)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
sk0x50: To be honest, I don't see any advantages to use method reference instead of lambdas here. It is a matter of taste. Anyway, I will change it if you want.
By the way, please take a look at this comment from Brian Goetz: https://stackoverflow.com/a/24493905/8371000

modules/core/src/test/java/org/apache/ignite/internal/DiagnosticLogForPartitionStatesTest.java
sk0x50: Why do you think so? What is the reason to use streams everywhere? :)

PR #989, opened 2020-03-13
GG-27955

sanpwc:


modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2644,78 +2651,128 @@ private synchronized void stop0(boolean cancel, boolean waitForBackups) {
...
+
+                        int cacheSpecificAmountOfOwners = -1;
+
+                        if (fullMap != null) {
+                            nodesToExclude.retainAll(fullMap.keySet());
+
+                            int parts = grpCtx.topology().partitions();
+
+                            for (int part = 0; part < parts; part++) {
+                                Set<UUID> partIdealAssignment = grpCtx.affinity().idealAssignmentRaw().get(part).
```
glukos: Topology / affinity cache can be updated concurrently. Possible solution: we can use topology read lock and / or check topology version of the affinity cache.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2644,78 +2651,128 @@ private synchronized void stop0(boolean cancel, boolean waitForBackups) {
...
+                        U.error(log, "Unable to read " + GRACEFUL_SHUTDOWN_METASTORE_KEY +
+                            " value from metastore.", e);
 
-                            blocked = true;
+                        continue;
+                    }
 
-                            break;
-                        }
+                    for (CacheGroupContext grpCtx : grid.context().cache().cacheGroups()) {
```
glukos: Should this code run on client nodes?
How will it work in case group context are not initialized (which is possible for clients)?

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2644,78 +2651,128 @@ private synchronized void stop0(boolean cancel, boolean waitForBackups) {
...
+
+                            int parts = grpCtx.topology().partitions();
+
+                            for (int part = 0; part < parts; part++) {
+                                Set<UUID> partIdealAssignment = grpCtx.affinity().idealAssignmentRaw().get(part).
+                                    stream().map(ClusterNode::id).collect(Collectors.toSet());
+
+                                int cnt = 0;
+
+                                for (Map.Entry<UUID, GridDhtPartitionMap> entry : fullMap.entrySet()) {
```
glukos: We basically execute the same algorithm as in GridDhtPartitionsExchangeFuture#updateMaximumAmountOfNodesAvailableForSafeStop. Can we extract it to one common place instead of copypasting?

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2644,78 +2651,128 @@ private synchronized void stop0(boolean cancel, boolean waitForBackups) {
...
+                            safeToStop = false;
 
-                            blocked = true;
+                            break;
                         }
+
                     }
 
-                    if (blocked) {
+                    safeToStop = safeToStop && topVer == grid.context().discovery().topologyVersion();
```
glukos: What's the point for this check?
We can't protect from parition loss if safe shutdown is performed simultaneously with sudden node leave anyway.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
glukos: The same for inactive clusters. How will this code work if we'll try to shutdown a server node safely on inactive cluster?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -2402,6 +2407,61 @@ private String exchangeTimingsLogMessage(String header, List<String> timings) {
...
+
+                    if (cacheSpecificAmountOfOwners == -1)
+                        cacheSpecificAmountOfOwners = cnt;
+
+                    cacheSpecificAmountOfOwners = Math.min(cacheSpecificAmountOfOwners, cnt);
+                }
+            }
+
+            if (nodeSpecificAmountOfOwners == -1)
+                nodeSpecificAmountOfOwners = cacheSpecificAmountOfOwners;
```
glukos: What's the point for having two variables for storing intermediate minimum?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -2402,6 +2407,61 @@ private String exchangeTimingsLogMessage(String header, List<String> timings) {
...
+
+            nodeSpecificAmountOfOwners = Math.min(nodeSpecificAmountOfOwners, cacheSpecificAmountOfOwners);
+        }
+
+        cctx.kernalContext().metric().registry(SYS_METRICS).intMetric(
```
glukos: Will this metric be correctly updated on client nodes?

modules/core/src/main/java/org/apache/ignite/mxbean/ClusterMetricsMXBean.java
```diff
@@ -260,6 +260,10 @@
     @Override @MXBeanDescription("Current PME duration in milliseconds.")
     public long getCurrentPmeDuration();
 
+    /** {@inheritDoc} */
+    @Override @MXBeanDescription("Maximum amount of nodes available for safe stop - stop without data loss.")
```
glukos: I'd mention in metric description that caches with backups=0 are not reflected in the metric.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
```diff
@@ -252,6 +272,139 @@ public void testShutdownWithoutBackups() throws Exception {
...
+        grid(2).close();
+
+        expMetastoreContent.remove(node1Id);
+        expMetastoreContent.add(node2Id);
+
+        assertEquals(
+            expMetastoreContent,
+            grid(0).context().distributedMetastorage().read(GRACEFUL_SHUTDOWN_METASTORE_KEY));
+    }
+
```
glukos: I'd extend test coverage with the following tests:
- Customer-like rolling restart case (nodes are restarted one by one, data shouldn't be lost)
- Stress test with simultaneous safe shutdown (all nodes are safely stopped from multiple threads, data shoudn't be lost)
- In-memory / persistent clusters coverage
- Inactive cluster coverage (node stop on inactive cluster shouldn't be delayed)
- Client nodes coverage (metrics is calculated correctly, stop of client nodes shoudn't be delayed)

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
sanpwc: There's no sense in such check cause we compare `grid.cluster().topologyVersion()` fixed at the beginning of the method and `grid.context().discovery().topologyVersion()`: `safeToStop = safeToStop && topVer == grid.context().discovery().topologyVersion();`

glukos:
Let's also add value of new metric to "Local node metrics": org.apache.ignite.internal.IgniteKernal#ackNodeMetrics

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
sanpwc: 1. Fixed;
2. I believe it'll work fine cause `grid.context().cache().cacheGroups()` will be empty, however in order to exclude unnecessary checks I've added `grid.cluster().active()` to `            if (waitForBackups && !grid.context().clientNode() && grid.cluster().active()) {...}`

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
sanpwc: It's not exact the same, cause code inside `GridDhtPartitionsExchangeFuture#updateMaximumAmountOfNodesAvailableForSafeStop` doesn't contain metastore-stored-excluded-nodes logic. Seems that code modification that  will support such distinction will be a sort of overdesign.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
sanpwc: There's no sense in having two variables. Fixed.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
sanpwc: Reworked with 
`                        if (prevReadyTopVer == AffinityTopologyVersion.NONE.topologyVersion())
                            prevReadyTopVer = grpCtx.topology().readyTopologyVersion().topologyVersion();
                        else if (prevReadyTopVer != grpCtx.topology().readyTopologyVersion().topologyVersion()) {
                            // At the moment, there is an exchange.
                            safeToStop = false;

                            break;
                        }`

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
sanpwc: Reworked, using compassion readyTopologyVersion of each cache group.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
sanpwc: Reworked, using comparison readyTopologyVersion of each cache group.

modules/core/src/main/java/org/apache/ignite/mxbean/ClusterMetricsMXBean.java
sanpwc: Fixed.

sanpwc:
> Let's also add value of new metric to "Local node metrics": org.apache.ignite.internal.IgniteKernal#ackNodeMetrics
Added.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
sanpwc: >Customer-like rolling restart case (nodes are restarted one by one, data shouldn't be lost)

org.apache.ignite.internal.processors.cache.distributed.dht.GridCacheDhtPreloadWaitForBackupsTest#testRollingRestartEmulation

>Stress test with simultaneous safe shutdown (all nodes are safely stopped from multiple threads, data shoudn't be lost)

org.apache.ignite.internal.processors.cache.distributed.dht.GridCacheDhtPreloadWaitForBackupsTest#testSimultaneousSafeShutdown

>In-memory / persistent clusters coverage

Already implemented within GridCacheDhtPreloadWaitForBackupsTest and 
GridCacheDhtPreloadWaitForBackupsWithPersistenceTest classes respectively.

>Inactive cluster coverage (node stop on inactive cluster shouldn't be delayed)

org.apache.ignite.internal.processors.cache.distributed.dht.GridCacheDhtPreloadWaitForBackupsTest#testNodeOnInactiveClusterShouldStopImmediately

>Client nodes coverage (metrics is calculated correctly, stop of client nodes shoudn't be delayed)

org.apache.ignite.internal.processors.cache.distributed.dht.GridCacheDhtPreloadWaitForBackupsTest#testClientNodeShouldStopImmediately

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
```diff
@@ -252,6 +273,235 @@ public void testShutdownWithoutBackups() throws Exception {
...
+    /**
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testClientNodeShouldStopImmediately() throws Exception {
+        cacheMode = CacheMode.PARTITIONED;
+        atomicityMode = CacheAtomicityMode.ATOMIC;
+        synchronizationMode = CacheWriteSynchronizationMode.PRIMARY_SYNC;
+        backups = 1;
```
glukos: 1 backup and 1 node is suspicious as well.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
```diff
@@ -252,6 +273,235 @@ public void testShutdownWithoutBackups() throws Exception {
...
+
+        startGrids(2);
+
+        grid(0).cluster().active(true);
+
+        for (int i = 0; i < cacheSize(); i++)
+            grid(i % 2).cache("cache" + (1 + (i >> 3) % 3)).put(i, i);
+
+        for (int i = 0; i < 2; i++) {
+            grid(i).close();
```
glukos: In fact, stop of only one node will be delayed.
Can we try scenario with more nodes? E.g. 4 nodes and two cycles of rolling restart.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
```diff
@@ -252,6 +273,235 @@ public void testShutdownWithoutBackups() throws Exception {
...
+
+        GridTestUtils.runAsync(() -> grid(0).close());
+
+        GridTestUtils.runAsync(() -> grid(1).close());
+
+        Thread.sleep(STOP_TIMEOUT_LIMIT);
+
+        startGrid(2);
+
+        assertTrue(waitForCondition(() -> grid(2).cluster().nodes().size() == 1, 10_000));
```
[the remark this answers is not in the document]
glukos: Test shouldn't rely on a specific server node. We can check data from a client node, for example.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
sanpwc: 1 backup and 1 server node, means that it should not be possible to stop server node with graceful shutdown.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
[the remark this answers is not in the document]
sanpwc: It's perfectly fine for grid(1).close to be called first. It's actually an idea of simultaneous stop.

> Test shouldn't rely on a specific server node. 
There's no such code here, cause we start one more node (grid(2)) in order to check that data still available.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
```diff
@@ -223,6 +243,7 @@ public void testShutdownWithoutBackups() throws Exception {
         synchronizationMode = CacheWriteSynchronizationMode.PRIMARY_SYNC;
         rebalanceMode = CacheRebalanceMode.ASYNC;
         clientNodes = false;
+        backups = 3;
```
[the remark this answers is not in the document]
sanpwc: Yep.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridCacheDhtPreloadWaitForBackupsTest.java
sanpwc: Fixed.

PR #990, opened 2020-03-13
GG-27408  Use one timezone for all nodes in cluster

tledkov-gridgain:


modules/h2/src/main/java/org/h2/util/DateTimeUtils.java
```diff
@@ -1585,15 +1585,15 @@ public static long convertScale(long nanosOfDay, int scale) {
     /**
      * @return Server time zone is used to convert date.
      */
-    private static TimeZone getTimeZone() {
+    public static TimeZone getTimeZone() {
```
ygerzhedovich, line 1588: it's not 'donatianable' change, could we do it by another way?

modules/clients/pom.xml
```diff
@@ -126,6 +126,13 @@
             <scope>test</scope>
         </dependency>
 
+        <dependency>
```
ygerzhedovich, line 129: I don't see usage of new dependency at all.

tledkov-gridgain: Used by startRemoteGrid

PR #991, opened 2020-03-13
GG-28018 Flush intermediate partition reconciliation result instead of accumulating it on heap

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
```diff
@@ -202,49 +200,75 @@ private String conflicts(
...
         return res.toString();
     }
 
+    /**
+     * @param partId Partitions id.
+     * @param rows Collection of data rows.
+     * @param nodesIdsToConsistentIdsMap Mapping node ids to consistent ids.
+     * @param includeSensitive Include sensitive.
+     */
+    private String conflicts(
```
sanpwc: Unclear method name and javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
```diff
@@ -202,49 +200,75 @@ private String conflicts(
...
+
+        return res.toString();
+    }
+
+    /**
+     * @param row Data row.
+     * @param nodesIdsToConsistentIdsMap Mapping node ids to consistent ids.
+     * @param includeSensitive Include sensitive.
+     */
+    public static String conflicts(
```
sanpwc: Unclear method name and javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
sk0x50: Thanks! I will fix it.

PR #993, opened 2020-03-17
GG-27553

denis-chudov:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -6139,6 +6144,13 @@ private void processMetricsUpdateMessage(TcpDiscoveryMetricsUpdateMessage msg) {
                         Map<Integer, CacheMetrics> cacheMetrics = msg.hasCacheMetrics(nodeId) ?
                             msg.cacheMetrics().get(nodeId) : Collections.emptyMap();
 
+                        if (metricsSizeNotProcessed && cacheMetrics.size() >= METRICS_QNT_WARN) {
+                            log.warning("The Discovery message has metrics for " + cacheMetrics.size() +
+                                " caches.\nTo prevent Discovery blocking use" +
+                                " -DIGNITE_DISCOVERY_DISABLE_CACHE=true option.");
```
denis-chudov: It's the same code as in ServerImpl - you'd better to move this in protected method to TcpDiscoveryImpl

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryMetricsWarnLogTest.java
```diff
@@ -0,0 +1,103 @@
...
+import org.apache.ignite.testframework.ListeningTestLogger;
+import org.apache.ignite.testframework.LogListener;
+import org.apache.ignite.testframework.junits.WithSystemProperty;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_DISCOVERY_METRICS_QNT_WARN;
+
+/**
+ * Class for testing warning log message about too many cache metrics
```
denis-chudov: please add dot

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryMetricsWarnLogTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        awaitMetricsUpdate(1);
+
+        assertTrue(logLsnr.check());
+    }
+
+    /**
+     *
+     */
+    private void createAndFillCache(int cacheNum, Ignite ignite) {
```
denis-chudov: we use one-line default javadoc for methods, like this /** */

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryMetricsWarnLogTest.java
```diff
@@ -0,0 +1,103 @@
...
+    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        return super.getConfiguration(igniteInstanceName)
+            .setMetricsUpdateFrequency(500L)
+            .setGridLogger(testLog);
+    }
+
+    /**
+     * @throws Exception If any error occurs.
+     */
+    @Test
```
denis-chudov: I think we should either have complete javadoc, or just have default one if we don't need it. Method description with only "throws" doesn't look good.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryImpl.java
```diff
@@ -419,6 +447,32 @@ protected boolean checkAckTimeout(long ackTimeout) {
         return true;
     }
 
+    /** */
+    public void processMsg(TcpDiscoveryMetricsUpdateMessage msg, long tsNanos) {
```
denis-chudov: This is too common name for a method that does such a specific thing. Please rename it.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryImpl.java
```diff
@@ -419,6 +447,32 @@ protected boolean checkAckTimeout(long ackTimeout) {
...
+            Map<Integer, CacheMetrics> cacheMetrics = msg.hasCacheMetrics(nodeId) ?
+                msg.cacheMetrics().get(nodeId) : Collections.emptyMap();
+
+            if (endTimeMetricsSizeProcessWait <= System.currentTimeMillis()
+                && cacheMetrics.size() >= METRICS_QNT_WARN)
+            {
+                log.warning("The Discovery message has metrics for " + cacheMetrics.size() +
+                    " caches.\nTo prevent Discovery blocking use -DIGNITE_DISCOVERY_DISABLE_CACHE=true option.");
+
+                endTimeMetricsSizeProcessWait = System.currentTimeMillis() + LOG_WARN_MSG_TIMEOUT;
```
denis-chudov: Please use org.apache.ignite.internal.util.IgniteUtils#currentTimeMillis ( U._currentTimeMillis_() ) instead of System.currentTimeMillis(). Also you can do just one call and save current timestamp to variable.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ClientImpl.java
```diff
@@ -385,6 +385,34 @@
             feature);
     }
 
+    /** {@inheritDoc} */
+    @Override public void updateMetrics(UUID nodeId,
```
ibessonov: Why have you moved this method up in the file? It makes it harder to review, can you please put it back? Thank you!

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -364,6 +364,29 @@ private GridKernalContext gridKernalContext() {
         return IgniteFeatures.allNodesSupports(gridKernalContext(), F.view(upcast(ring.allNodes()), nodesPred), feature);
     }
 
+    /** {@inheritDoc} */
+    @Override public void updateMetrics(UUID nodeId,
```
ibessonov: Same thing here, let's make it easier to explore history in git.

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryMetricsWarnLogTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+    /**
+     * Test checks that the desired message occurs in logs.
+     *
+     * @throws Exception If any error occurs.
+     */
+    @Test
+    @WithSystemProperty(key = IGNITE_DISCOVERY_METRICS_QNT_WARN, value = "20")
+    public void testMetricsWarningLog() throws Exception {
+        System.setProperty(IGNITE_DISCOVERY_METRICS_QNT_WARN, "20");
```
ibessonov: Why do you duplicate setting of the system property? Please remove it

modules/core/src/test/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoveryMetricsWarnLogTest.java
```diff
@@ -0,0 +1,103 @@
...
+
+        awaitMetricsUpdate(1);
+
+        assertTrue(logLsnr.check());
+    }
+
+    /** */
+    private void createAndFillCache(int cacheNum, Ignite ignite) {
+        IgniteCache<Object, Object> cache = ignite.getOrCreateCache(
+                new CacheConfiguration<>(DEFAULT_CACHE_NAME + cacheNum).setStatisticsEnabled(true)
```
ibessonov: Please format it properly.

PR #998, opened 2020-03-17
Gg 27221

vldpyatkov:


modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+                    treeInfo.get().forEach((name, info) -> {
+                        info.innerPageIds.forEach(id -> {
+                            Class cls = name.equals(META_TREE_NAME)
+                                ? IndexStorageImpl.MetaStoreInnerIO.class
+                                : H2ExtrasInnerIO.class;
+
+                            pageIoIds.computeIfAbsent(cls, k -> new HashSet<>()).add(id);
+                        });
+
+                        pageIoIds.computeIfAbsent(BPlusMetaIO.class, k -> new HashSet<>()).add(info.rootPageId);
```
vldpyatkov: Is info.rootPageId differ from metaTreeRootId?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+        Map<Class, AtomicLong> ioStat = new HashMap<>();
+
+        TreeTraverseContext treeCtx = new TreeTraverseContext(treeName, store, ioStat, errors, null, null, null);
+
+        ByteBuffer buf = allocateBuffer(pageSize);
+
+        try {
+            long addr = bufferAddress(buf);
+
+            buf.rewind();
```
vldpyatkov: Why need rewind here?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+                if (classesInStat.contains(cls))
+                    // Already checked.
+                    return;
+
+                errors.add(compareError("pages", name, 0, cnt.get(), cls));
+            });
+        });
+
+        treeScans.forEach((name, tree) -> {
+            if (!treeIdxNames.contains(name))
```
vldpyatkov: Are you sure checking _key_PK correct?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+     *
+     * @param metaPageListId Page list meta id.
+     * @return Meta info.
+     */
+    private PageListsInfo getPageListsMetaInfo(long metaPageListId) {
+        Map<IgniteBiTuple<Long, Integer>, List<Long>> bucketsData = new HashMap<>();
+
+        Set<Long> allPages = new HashSet<>();
+
+        Map<Class, AtomicLong> pageListStat = new HashMap<>();
```
vldpyatkov: Is using atomic long here definitely needed?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+
+            if (tree.itemStorage.size() != scan.itemStorage.size())
+                errors.add(compareError("items", name, tree.itemStorage.size(), scan.itemStorage.size(), null));
+
+            Set<Class> classesInStat = new HashSet<>();
+
+            tree.ioStat.forEach((cls, cnt) -> {
+                classesInStat.add(cls);
+
+                AtomicLong scanCnt = scan.ioStat.get(cls);
```
vldpyatkov: Rewrite that without atomic:
long scanCnt = scan.ioStat.get(cls) == null ? 0 : scan.ioStat.get(cls).get();

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+     * Traverse single index tree from root to leafs.
+     *
+     * @param store File page store.
+     * @param rootPageId Root page id.
+     * @param treeName Tree name.
+     * @param itemStorage Items storage.
+     * @return Tree traversal info.
+     */
+    private TreeTraversalInfo traverseTree(FilePageStore store, long rootPageId, String treeName, ItemStorage itemStorage) {
+        Map<Class, AtomicLong> ioStat = new HashMap<>();
```
vldpyatkov: May atomic not need here.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+
+                ioProcessor = getIOProcessor(io);
+
+                pageContent = ioProcessor.getContent(io, addr, pageId, nodeCtx);
+            }
+            finally {
+                freeBuffer(buf);
+            }
+
+            return ioProcessor.getNode(pageContent, pageId, nodeCtx);
```
vldpyatkov: It looks too strange:
pageContent = ioProcessor.getContent(io, addr, pageId, nodeCtx);
ioProcessor.getNode(pageContent, pageId, nodeCtx);
You are getting some object from other and returning it to itself.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+
+        AtomicReference<Map<String, TreeTraversalInfo>> treeInfo = new AtomicReference<>();
+
+        AtomicReference<Map<String, TreeTraversalInfo>> horizontalScans = new AtomicReference<>();
+
+        AtomicReference<PageListsInfo> pageListsInfo = new AtomicReference<>();
+
+        List<Throwable> errors;
+
+        ProgressPrinter progressPrinter = new ProgressPrinter(System.out, "Reading pages sequentially", pagesNum);
```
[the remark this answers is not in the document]
denis-chudov: Here and below, using System.out is correct for progress printer - is should print **only** to System.out, not in output file.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
denis-chudov: Yes, metaTreeRootId is root for meta tree, and info.rootPageId is root of each index tree.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
denis-chudov: Yes, because tree names for _key_PK indexes are different as they include cache id and type id.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
denis-chudov: Maybe it does, but it is designed in such way to free the buffer before recursive call, i.e. it's splitting one operation to two steps, the buffer can be deallocated after first one.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+        }
+    }
+
+    /** */
+    private FilePageStore getIdxStore() throws IgniteCheckedException {
+        try {
+            return (FilePageStore)storeFactory.createPageStore(FLAG_IDX, idxFile, allocationTracker);
+        }
+        catch (IllegalArgumentException e) {
+            return null;
```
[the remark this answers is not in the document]
denis-chudov: It goes deeper - this was done to ignore IllegalArgumentException on snapshot pages. I am thinking about adding another constructor for transform operation, then this method will be removed.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+     * @param metaTreeRoot Meta tree root page id.
+     * @return Index trees info.
+     */
+    private Map<String, TreeTraversalInfo> traverseAllTrees(
+        String traverseProcCaption,
+        long metaTreeRoot,
+        Supplier<ItemStorage> itemStorageFactory,
+        TraverseProc traverseProc
+    ) {
+        Map<String, TreeTraversalInfo> treeInfos = new HashMap<>();
```
vldpyatkov: Better to use LinkedHashMap here, because I want to see MetaTree at head of view.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+                    horizontalScans.set(traverseAllTrees("Scan index trees horizontally", metaTreeRootId, itemStorageFactory, this::horizontalTreeScan));
+                }
+                else if (io instanceof PagesListMetaIO)
+                    pageListsInfo.set(getPageListsMetaInfo(pageId));
+                else {
+                    if (indexes == null) {
+                        ofNullable(pageIoIds.get(io.getClass())).ifPresent((pageIds) -> {
+                            if (!pageIds.contains(pageId)) {
+                                boolean foundInList =
+                                    (pageListsInfo.get() != null && pageListsInfo.get().allPages.contains(pageId));
```
vldpyatkov: Not need to mention as orphan if page found in reuse list.
I have got the warning when destroyed sql cache in test.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
vldpyatkov: Why does not free buffer after last one?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
```diff
@@ -0,0 +1,2130 @@
...
+    private interface PageIOProcessor {
+        /**
+         * Gets deserialized content.
+         * @param io Page IO.
+         * @param addr Page address.
+         * @param pageId Page id.
+         * @param nodeCtx Tree traversal context.
+         * @return Page content.
+         */
+        PageContent getContent(PageIO io, long addr, long pag
```
vldpyatkov: You can return CacheAwareLink here.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteIndexReader.java
denis-chudov: In some cases we need only PageContent without creating TreeNode (see #horizontalTreeScan)

PR #1000, opened 2020-03-18
GG-26599 Security permissions for setting query quota size

korlov42:


modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinTcpIo.java
```diff
@@ -338,14 +339,19 @@ private HandshakeResult handshake(ClientListenerProtocolVersion ver) throws IOEx
 
             String err = reader.readString();
 
+            int status = reader.readInt();
```
AMashenkov: It is possible to get EOFException here, isn't ir?

korlov42: only when you will try to connect to server with version < 1.1.0

PR #1002, opened 2020-03-18
GG-28148

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -1082,24 +1054,36 @@ public IgniteConfiguration setBuildIndexThreadPoolSize(int poolSize) {
...
      * will be switched off.
      *
      * @return SQL query history size.
+     *
+     * @deprecated Use {@link SqlInitialConfiguration#setSqlQueryHistorySize(int)} instead.
      */
+    @Deprecated
     public int getSqlQueryHistorySize() {
-        return sqlQryHistSize;
+        if (sqlInitCfg == null)
```
ygerzhedovich: Why you check on null here? you already has initial value at line 686

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/GridSqlQueryParser.java
```diff
@@ -1922,16 +1918,16 @@ else if (o instanceof GridSqlDelete)
...
-     * Parse query.
-     *
-     * @param prepared Prepared statement.
-     * @param useOptimizedSubqry Whether to user optimized subquery.
-     * @return Parsed query.
-     */
-    public static GridSqlQuery parseQuery(Prepared prepared, boolean useOptimizedSubqry, IgniteLogger log) {
-        return (GridSqlQuery)new GridSqlQueryParser(useOptimizedSubqry, log).parse(prepared);
-    }
+//    /**
```
ygerzhedovich: commented code

PR #1005, opened 2020-03-20
GG-27638: IdleVerify and ValidateIndexes use eventual payload check.

vmalin:


vmalin (commented):
I have left some comments

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/IdleVerifyUtility.java
```diff
@@ -109,6 +112,70 @@ public static boolean isCheckpointNow(@Nullable IgniteCacheDatabaseSharedManager
...
     }
 
+    /**
+     * Gather updateCounters info.
+     *
+     * @param ign Ignite instance.
+     * @param grpIds Group Id`s.
+     * @return Counters info.
+     */
+    public static T2<Set<Integer>, Map<Integer, Set<Map.Entry<Integer, Long>>>> updCountersSnapshot(
```
vmalin: Consider using small inner class instead of T2. This will allow you to use descriptive methods names. Or extend documentation for return type

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/IdleVerifyUtility.java
```diff
@@ -109,6 +112,70 @@ public static boolean isCheckpointNow(@Nullable IgniteCacheDatabaseSharedManager
...
+        Map<Integer, Set<Map.Entry<Integer, Long>>> partsWithCountersPerGrp = new HashMap<>();
+
+        for (Integer grpId : grpIds) {
+            CacheGroupContext grpCtx = ign.context().cache().cacheGroup(grpId);
+
+            GridDhtPartitionTopology top = grpCtx.topology();
+
+            partsWithCountersPerGrp.putIfAbsent(grpId, new HashSet<>());
+
+            Set<Map.Entry<Integer, Long>> partsWithCounters = partsWithCountersPerGrp.get(grpId);
```
vmalin: partsWithCountersPerGrp.computeIfAbsent(grpId, () -> new HashSet<>()); may be better here. It allows to avoid getting element that was put in line above.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/VerifyBackupPartitionsTaskV2.java
```diff
@@ -629,6 +620,8 @@ private boolean isCacheMatchFilter(DynamicCacheDescriptor desc) {
 
             completionCntr.incrementAndGet();
 
+            idleCheck.run();
```
vmalin: If I understand correctly executing method body is not needed if idleCheck fails. Consider moving it to the start of the method(not forgetting about completionCntr logic)

modules/indexing/src/main/java/org/apache/ignite/internal/visor/verify/ValidateIndexesClosure.java
```diff
@@ -283,11 +289,66 @@ private VisorValidateIndexesJobResult call0() {
...
 
+        List<Future<Map<PartitionKey, ValidateIndexesPartitionResult>>> procPartFutures = new ArrayList<>(partArgs.size());
+        List<Future<Map<String, ValidateIndexesPartitionResult>>> procIdxFutures = new ArrayList<>(idxArgs.size());
+
+        List<T3<CacheGroupContext, GridDhtLocalPartition, Future<CacheSize>>> cacheSizeFutures = new ArrayList<>(partArgs.size());
+        List<T3<GridCacheContext, Index, Future<T2<Throwable, Long>>>> idxSizeFutures = new ArrayList<>(idxArgs.size());
+
         for (T2<CacheGroupContext, GridDhtLocalPartition> t2 : partArgs)
-            procPartFutures.add(processPartitionAsync(t2.get1(), t2.get2()));
+            procPartFutures.add(processPartitionAsync(t2.get1(), t2.get2(), new Runnable() {
```
vmalin: new Runnable() code looks like copy-paste. It should be moved to separate object

modules/indexing/src/test/java/org/apache/ignite/util/GridCommandHandlerIndexingTest.java
```diff
@@ -85,7 +90,111 @@ public void testValidateIndexesFailedOnNotIdleCluster() throws Exception {
...
+        testOut.reset();
+
+        if (persistenceEnable())
+            db.enableCheckpoints(true).get();
+
+        assertEquals(EXIT_CODE_OK, execute("--cache", "validate_indexes", "--check-crc"));
+
+        out = testOut.toString();
+
+        assertNotContains(log, out, IDLE_DATA_ALTERATION_MSG);
```
vmalin: You log group names for cache groups associated with diverged partitions but never check them in log

modules/indexing/src/test/java/org/apache/ignite/util/GridCommandHandlerIndexingTest.java
```diff
@@ -85,7 +90,111 @@ public void testValidateIndexesFailedOnNotIdleCluster() throws Exception {
         String out = testOut.toString();
 
         assertContains(log, out, "Index validation failed");
-        assertContains(log, out, "Checkpoint with dirty pages started! Cluster not idle!");
+        assertContains(log, out, IDLE_DATA_ALTERATION_MSG);
```
vmalin: You added tests for validate_indexes command but not for idle_verify. QA will have question about it

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/VerifyBackupPartitionsTaskV2.java
```diff
@@ -255,93 +256,83 @@ public VerifyBackupPartitionsJobV2(VisorIdleVerifyTaskArg arg) {
 
             completionCntr.set(0);
 
-            AtomicBoolean cpFlag = new AtomicBoolean();
+            /** Update counters per partiton per group. */
```
vmalin: typo: partiton -> partition

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/IdleVerifyUtility.java
```diff
@@ -109,6 +112,70 @@ public static boolean isCheckpointNow(@Nullable IgniteCacheDatabaseSharedManager
...
+    /**
+     * Compares two sets with partitions and upd counters per group distribution.
+     *
+     * @param ign Ignite instance.
+     * @param cntrsIn Set to compare.
+     * @return Diff with grpId info between two sets.
+     */
+    public static List<Integer> compareUpdCounters(
+        IgniteEx ign,
+        T2<Set<Integer>, Map<Integer, Set<Map.Entry<Integer, Long>>>> cntrsIn
```
vmalin: Please extend cntrsIn documentation

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/IdleVerifyUtility.java
```diff
@@ -109,6 +112,70 @@ public static boolean isCheckpointNow(@Nullable IgniteCacheDatabaseSharedManager
...
+    public static T2<Set<Integer>, Map<Integer, Set<Map.Entry<Integer, Long>>>> updCountersSnapshot(
+        IgniteEx ign,
+        Set<Integer> grpIds
+    ) {
+        Map<Integer, Set<Map.Entry<Integer, Long>>> partsWithCountersPerGrp = new HashMap<>();
+
+        for (Integer grpId : grpIds) {
+            CacheGroupContext grpCtx = ign.context().cache().cacheGroup(grpId);
+
+            GridDhtPartitionTopology top = grpCtx.topology();
```
vmalin: Null check is required here

modules/indexing/src/test/java/org/apache/ignite/util/GridCommandHandlerIndexingTest.java
zstan: @vmalin i append **additional** tests for **already**  existing functional, mentioned tests already presented here:  GridCommandHandlerTest

modules/indexing/src/main/java/org/apache/ignite/internal/visor/verify/ValidateIndexesClosure.java
```diff
@@ -283,25 +285,41 @@ private VisorValidateIndexesJobResult call0() {
         totalPartitions = partArgs.size();
         totalIndexes = idxArgs.size();
 
+        List<Future<Map<PartitionKey, ValidateIndexesPartitionResult>>> procPartFutures = new ArrayList<>(partArgs.size());
+        List<Future<Map<String, ValidateIndexesPartitionResult>>> procIdxFutures = new ArrayList<>(idxArgs.size());
+
+        List<T3<CacheGroupContext, GridDhtLocalPartition, Future<CacheSize>>> cacheSizeFutures = new ArrayList<>(partArgs.size());
+        List<T3<GridCacheContext, Index, Future<T2<Throwable, Long>>>> idxSizeFutures = new ArrayList<>(idxArgs.size());
+
+        IdleVerifyUtility.IdleChecker idleChecker = new IdleVerifyUtility.IdleChecker(ignite, partsWithCntrsPerGrp);
```
vmalin: This checker is passed to loops by partArgs which may contain significant number of entries. Although check is light-weighted I would advise to pass it to loops only if you are sure that single check before loops is insufficient.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/VerifyBackupPartitionsTaskV2.java
```diff
@@ -538,29 +504,31 @@ private boolean isCacheMatchFilter(DynamicCacheDescriptor desc) {
...
         private Map<PartitionKeyV2, PartitionHashRecordV2> calculatePartitionHash(
             CacheGroupContext grpCtx,
             GridDhtLocalPartition part,
-            AtomicBoolean cpFlag
+            Runnable idleCheck
         ) {
             if (!part.reserve())
                 return emptyMap();
 
+            idleCheck.run();
```
vmalin: Simple moving this check to the beginning of the method may break logic of complitionCntr. I think complitionCntr increment should be moved to finally block like in validate_indexes

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/verify/IdleVerifyUtility.java
```diff
@@ -85,28 +94,117 @@ public static void checkPartitionsPageCrcSum(
...
+    }
+
+    /**
+     * Gather updateCounters info.
+     *
+     * @param ign Ignite instance.
+     * @param grpIds Group Id`s.
+     * @return Current group id`s and distribution partitions with update counters per group.
+     */
+    public static T2<Set<Integer>, Map<Integer, Set<Map.Entry<Integer, Long>>>> updCountersSnapshot(
```
sergey-chugunov-1985, line 108: If I understand implementation correctly, this method either returns T2 with set containing the same elements as map.keySet or throws GridNotIdleException.
If this set is not used for some other purposes we can decrease complexity of returned data structure.

Another possible simplification: if we use map<partId, updateCounter> instead of Set<Map.Entry<partId, updateCounter> will it break anything?

sergey-chugunov-1985, line 108: Also using abbreviations like `updCountersSnapshot` should be avoided in method names, let use self-explanatory `getUpdateCountersSnapshot`

sergey-chugunov-1985:
There is a performance concern that if we have a lot of caches with a lot of partitions on each node idle_verify utility could eat up a good bunch of memory.
We could mitigate this risk either by using more compact data structures or by checking cache groups one-by-one (which could in turn increase time to finish the job, so it looks like we have to decide here what is our priority).

PR #1011, opened 2020-03-25
GG-28214 Renamed default schema field.

vsisko:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/action/controller/QueryActionsController.java
```diff
@@ -132,9 +133,13 @@ public QueryActionsController(GridKernalContext ctx) {
 
                 SqlFieldsQuery qry = prepareQuery(arg);
 
-                GridCacheContext cctx = F.isEmpty(arg.getCacheName())
+                IgniteInternalCache<Object, Object> cache = F.isEmpty(arg.getDefaultSchema())
+                    ? null
+                    : ctx.cache().cache(arg.getDefaultSchema());
```
nva: org.apache.ignite.internal.processors.cache.GridCacheProcessor#cache use the cache name as an argument, but you pass schema name. How this will work?

vsisko: Expected that cache name will be sent and its context will be used.

modules/management-console-agent/src/test/java/org/apache/ignite/agent/action/controller/QueryActionsControllerTest.java
```diff
@@ -365,7 +453,7 @@ public void shouldReturnRunningQueriesFromAllNodes() {
                         .setQueryId(UUID.randomUUID().toString())
                         .setQueryText("SELECT count(*), sleep() AS \"" + nid + "\" FROM \"TestCache\".STRING")
                         .setPageSize(1)
-                        .setCacheName("TestCache")
+                        .setDefaultSchema("TestCache")
```
nva: Let's add test where schema is removed from query and set we set default schema

vsisko: This variant is already implemented in my test:
getSelectQuery(fullTestTable) + getSelectQuery(testTable)

PR #1015, opened 2020-03-27
Gg 28148

tledkov-gridgain:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/DistributedSqlConfiguration.java
```diff
@@ -0,0 +1,116 @@
...
+    private final SimpleDistributedProperty<HashSet<String>> disabledSqlFuncs
+        = new SimpleDistributedProperty<>("sql.disabledFunctions");
+
+    /**
+     * @param isp Subscription processor.
+     * @param ctx Kernal context.
+     */
+    public DistributedSqlConfiguration(
+        GridInternalSubscriptionProcessor isp,
+        GridKernalContext ctx,
```
ygerzhedovich: what reason to have context here?

tledkov-gridgain: Good point. I guess log parameter must be removed and logger is created on constructo from ctx.log().

Fixed

PR #1021, opened 2020-04-02
GG-26047 Added support of decimal inlining

RealZheka:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/DecimalIndexTest.java
```diff
@@ -0,0 +1,137 @@
...
+        testLog.registerListener(lsnr);
+
+        Ignite node = startGrid(0);
+
+        sql(node, "create table " + TBL_NAME + " (key_pk int, key1 decimal, key2 decimal, key3 decimal, primary key(key_pk), val int)");
+        sql(node, "create index pk_id on " + TBL_NAME + " (key1, key2, key3) inline_size 20");
+        sql(node, "insert into " + TBL_NAME + "(key_pk, key1, key2, key3, val) values (1, 1.0, 2.0, 3.0, 1)");
+        sql(node, "insert into " + TBL_NAME + "(key_pk, key1, key2, key3, val) values (2, 1.0, 2.0, 3.0, 1)");
+
+        assertFalse("Unexpected full inline of the index", lsnr.check());
```
korlov42, line 101: full inline is expected here

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/DecimalIndexTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        return igniteCfg;
+    }
+
+    /**
+     *
+     */
+    @Test
+    @WithSystemProperty(key = IGNITE_THROTTLE_INLINE_SIZE_CALCULATION, value = "1")
+    public void smallInlineSIze() throws Exception {
```
korlov42, line 63: misspell "SIze" here and below

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/DecimalIndexTest.java
```diff
@@ -0,0 +1,137 @@
...
+
+        assertFalse("Unexpected full inline of the index", lsnr.check());
+    }
+
+    /**
+     *
+     */
+    @Test
+    @WithSystemProperty(key = IGNITE_THROTTLE_INLINE_SIZE_CALCULATION, value = "1")
+    public void nullAndOutOfRangeValues() throws Exception {
```
korlov42, line 109: what should this test check?

RealZheka: Was created based on the following comment:
_lack of comparison with null
lack of comparison with different precisions (-0.0000000000000000000001234567f still perfectly fit the float)_
Supposed to test situation when one of the columns is NULL or contains value that is out of Float range.

korlov42: And how this situation is tested? I mean there is only one assertion and it uses the result of query without any condition so scan index will be chosen, and this index has nothing common with inlining.

There should be query with equals and range conditions; whole result should be validated not only size; also there should be validation of a EXPLAIN output (right indexed should be chosen)

korlov42 (approved):

PR #1022, opened 2020-04-02
GG-28388

tkalkirill:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/latch/ExchangeLatchManagerTest.java
```diff
@@ -16,20 +16,124 @@
...
+    /**
+     *
+     */
     private static final String LATCH_NAME = "test";
 
+    /** Message are meaning that node getting a stale acknowledge message. */
+    private static final String STALE_ACK_LOG_MSG = "Ignoring stale latch's acknowledge";
+
+    /** Message happens when assertion was broken. */
+    public static final String ERROR_MSG = "An error occurred processing the message.*LatchAckMessage";
```
tkalkirill, line 48: It looks like a Pattern, please replace it with it, but with the value: "An error occurred processing the message.*LatchAckMessage.*"

vldpyatkov: Thanks, I was fix it.

PR #1023, opened 2020-04-02
GG-26123 Added fallback to full rebalance if historical one has failed.

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -3214,8 +3234,29 @@ private void body0() throws InterruptedException, IgniteCheckedException {
                         if (isCancelled())
                             break;
 
-                        if (task instanceof RebalanceReassignExchangeTask)
-                            exchId = ((RebalanceReassignExchangeTask) task).exchangeId();
+                        if (task instanceof RebalanceReassignExchangeTask) {
```
ascherbakoff, line 3237: You can use here last completed exchange future.
If it's version is higher, reassign task should be immediately cancelled. New exchange will trigger new rebalancing.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -3969,4 +4012,52 @@ public CacheRebalanceOrder(int order, CacheRebalanceMode mode) {
             return result;
         }
     }
+
+    /**
+     * Represents reassign context and stores node ids that cannot be used for historical rebalance for the given exchange.
+     */
+    private static class RebalanceReassignContext {
```
ascherbakoff, line 4019: I think RebalanceReassignContext is not needed.
Just keep a set of failed historical supplier ids until the node has left.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionSupplier.java
```diff
@@ -475,6 +478,11 @@ else if (iter.isPartitionMissing(p)) {
                 log.info("Finished supplying rebalancing [" + supplyRoutineInfo(topicId, nodeId, demandMsg) + "]");
         }
         catch (Throwable t) {
+            boolean fallbackToFullRebalance = (iter == null) ?
```
ascherbakoff, line 481: This code looks fragile. Do we not want to do fallback if this is surely a WAL error, like IO error ?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePreloader.java
```diff
@@ -74,10 +75,13 @@
...
      * @param exchFut Completed exchange future. Can be {@code null} if forced or reassigned generation occurs.
+     * @param historicalExclusions Set of node ids that should not be used for historical rebalance.
      * @return Partition assignments which will be requested from supplier nodes.
      */
-    @Nullable public GridDhtPreloaderAssignments generateAssignments(GridDhtPartitionExchangeId exchId,
-                                                                     @Nullable GridDhtPartitionsExchangeFuture exchFut);
+    @Nullable public GridDhtPreloaderAssignments generateAssignments(
+        GridDhtPartitionExchangeId exchId,
+        @Nullable GridDhtPartitionsExchangeFuture exchFut,
+        Set<UUID> historicalExclusions);
```
ascherbakoff, line 84: This field can be retrieved from exchange future. No need to change method contract.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/IgniteRebalanceIteratorImpl.java
```diff
@@ -96,6 +99,11 @@ private synchronized void advance() throws IgniteCheckedException {
         return historicalIterator != null && historicalIterator.contains(partId);
     }
 
+    /** {@inheritDoc} */
+    @Override public synchronized boolean allHistoricalPartitionsDone() {
+        return historicalIterator == null || reachedHistEnd;
```
ascherbakoff, line 104: Why we need reachedHistEnd ?
Can it be replaced with historicalIterator  != null && !historicalIterator.hasNext() ?

sk0x50: Unfortunately, no. It cannot be replaced. The reason is `WALHistoricalIterator#advance()`. Being more precise, this method at the first step nullifies the field `next`. If this method throws an exception while it iterates data entries, the field `next` still equals to 'null', and so `hasNext()` returns `false`. It looks weird to me. Perhaps we need to rework `WALHistoricalIterator#advance()`. What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
sk0x50: We don't put a new reassign task into the exchange queue if it already contains a server exchange. Anyway, using the last completed exchange future with the condition looks reasonable to me. I will try to fix.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionSupplier.java
sk0x50: Do you mean that we need a special exception (something like as follows: `WalHistoryIteratorInitializationException`) and use it for this case?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
sk0x50: I would prefer the idea to keep this set of nodes in the corresponding exchange future. This approach gives a try to the excluded node to become a historical supplier in the future.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -3216,196 +3207,227 @@ private void body0() throws InterruptedException, IgniteCheckedException {
...
 
-                            exchId = ((ForceRebalanceExchangeTask)task).exchangeId();
+                                        break;
+                                    }
+                                }
+                            }
                         }
-                        else {
-                            assert task instanceof GridDhtPartitionsExchangeFuture : task;
+                    }
```
ascherbakoff, line 3251: This is a lot of code.
I think it can be simplified (if you will pass exchange future in reassigntask)

 to:

```
if (task instanceof RebalanceReassignExchangeTask) {
                            exchId = ((RebalanceReassignExchangeTask) task).exchangeId();

                            GridDhtPartitionsExchangeFuture fut = ((RebalanceReassignExchangeTask) task).future();
                            assert fut.changedAffinity();                            

                            if (fut.hasInapplicableNodesForRebalance()) {
                                  GridDhtPartitionsExchangeFuture lastFut = lastFinishedFut.get();

                                  AffinityTopologyVersion lastAffVer = cctx.exchange()
                                      lastAffinityChangedTopologyVersion(lastFut.topologyVersion());
                            
                                 if (fut.exchangeId().topologyVersion().equals(lastAffVer))
                                      exchFut = fut;
                             }
}
```

I also suggest to avoid iterations over _cctx.exchange().exchangeFutures()_. 
I have plans to remove this method in the future (because we do not need exchange history)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1059,14 +1060,14 @@ private String demandRoutineInfo(UUID supplier, GridDhtPartitionSupplyMessage su
      * Internal states of rebalance future.
      */
     private enum RebalanceFutureState {
-        /** Init. */
+        /** Initial state. */
         INIT,
 
-        /** Started. */
+        /** Rebalance future is started and requests required patitions. */
```
ascherbakoff, line 1066: Spellcheck is required.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -308,6 +308,15 @@
     @GridToStringExclude
     private volatile IgniteDhtPartitionHistorySuppliersMap partHistSuppliers = new IgniteDhtPartitionHistorySuppliersMap();
 
+    /** Set of nodes that cannot be used for wal rebalancing due to some reason. */
+    private Set<UUID> exclusionsFromWalRebalance = Collections.newSetFromMap(new ConcurrentHashMap<>());
```
ascherbakoff, line 312: I suggest to rename it ot exclusionsFromWALRebalance or exclusionsFromHistoricalRebalance.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPreloader.java
```diff
@@ -199,7 +199,11 @@ public boolean disableRebalancingCancellationOptimization() {
 
         AffinityTopologyVersion topVer = top.readyTopologyVersion();
 
-        assert exchFut == null || exchFut.context().events().topologyVersion().equals(top.readyTopologyVersion()) :
+        List<GridDhtPartitionsExchangeFuture> futs = ctx.exchange().exchangeFutures();
```
ascherbakoff, line 202: I suggest not to use this method.
The reasons in the previous comment.
Moreover, looks like call result it is never used.

sk0x50: Unused variable removed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
sk0x50: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionSupplier.java
sk0x50: Introduced `IgniteHistoricalIteratorException`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
```diff
@@ -557,7 +566,70 @@ public ExchangeType exchangeType() {
...
+            Collections.newSetFromMap(new ConcurrentHashMap<>())
+        );
+
+        parts.add(p);
+    }
+
+    /**
+     * @return {@code true} if there are nodes which are inapplicable for wal rebalancing.
+     */
+    public boolean hasInapplicableNodesForWalRebalance() {
```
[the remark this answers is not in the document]
sk0x50: Renamed to `hasInapplicableNodesForHistoricalRebalance`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
sk0x50: Improved in the way as you suggested.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsExchangeFuture.java
sk0x50: Renamed to `exclusionsFromHistoricalRebalance`

PR #1026, opened 2020-04-03
GG-28131

tkalkirill:


modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationSpiConfigSelfTest.java
```diff
@@ -70,4 +98,67 @@ public void testLocalPortRange() throws Exception {
...
+     * Test checks that {@link TcpCommunicationSpi#ATTR_HOST_NAMES} attribute
+     * is empty only if {@link
+     * IgniteSystemProperties#IGNITE_TCP_COMM_SET_ATTR_HOST_NAMES} ==
+     * {@code false} and local host is set via {@link
+     * IgniteConfiguration#setLocalHost}.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    @WithSystemProperty(key = IGNITE_TCP_COMM_SET_ATTR_HOST_NAMES, value = "false")
```
vldpyatkov, line 111: Why is default value need to define explicitly?

tkalkirill: To save default behavior, but may need to divide it into 2 tests.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1394,6 +1395,14 @@
      */
     public static final String IGNITE_INDEX_COST_FUNCTION = "IGNITE_INDEX_COST_FUNCTION";
 
+    /**
+     * Enables setting attribute value of {@link
+     * TcpCommunicationSpi#ATTR_HOST_NAMES ATTR_HOST_NAMES} when value {@link
+     * IgniteConfiguration#getLocalHost getLocalHost} is ip.
+     * By default, {@code false}.
```
akalash, line 1402: It's not clear from javadoc why do we need this property. Can you please expand it?
As I understand, we want to change current behavior to following one -  ignite always sends ATTR_ADDRS and if the user configures local address as hostname, ignite additionally send ATTR_HOST_NAMES. But we also want to keep backward compatibility(when ignite always send ATTR_HOST_NAMES), so this property added exactly for this purpose, am I right?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -2224,8 +2225,11 @@ private void dumpInfo(StringBuilder sb, UUID dstNodeId) {
         else
             connPlc = new FirstConnectionPolicy();
 
+        boolean locAddrIp;
+
         try {
             locHost = U.resolveLocalHost(locAddr);
+            locAddrIp = !F.isEmpty(locAddr) && locHost.toString().startsWith("/");
```
akalash, line 2232: Can you replace this condition to a more readable one? For example, IPAddressUtil already has methods isIPv4LiteralAddress and isIPv6LiteralAddress. You also can build your own one. 
Secondly, your check is pretty dangerous. If somebody invokes locHost.getHostName() before your code, your checking is fail because getHostName has side effect.

PR #1027, opened 2020-04-03
GG-28205 Implement of sending a snapshot informations from the grid through agent

nva:


modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/ClusterInfoProcessor.java
```diff
@@ -129,25 +129,23 @@ void sendClusterInfo(Event evt) {
...
-        populateClusterInfo(clusterInfo);
-
-        mgr.send(buildClusterDest(cluster.id()), clusterInfo);
+        mgr.send(buildClusterDest(cluster.id()), buildClusterInfo());
     }
 
     /**
      * @return Create cluster info.
      */
     protected ClusterInfo createClusterInfo() {
```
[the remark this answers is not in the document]
dmekhanikov: It seems that `buildClusterInfo()` should be made `protected` and overloaded in the ancestor, not `createClusterInfo`.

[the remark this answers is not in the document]
smoldachev: In this case we will copy-paste  follow part of code:

```java
.setId(cluster.id())
            .setTag(cluster.tag())
            .setActive(cluster.active())
            .setPersistenceEnabled(CU.isPersistenceEnabled(ctx.config()))
            .setBaselineParameters(
                new BaselineInfo(
                    cluster.isBaselineAutoAdjustEnabled(),
                    cluster.baselineAutoAdjustTimeout()
                )
            )
            .setSecure(ctx.authentication().enabled() || ctx.security().enabled())
            .setFeatures(getClusterFeatures(ctx, ctx.cluster().get().nodes()));
```

modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
```diff
@@ -226,6 +226,45 @@ public static void quiteStop(GridProcessor proc) {
...
+     * @param args array of objects to be passed as arguments to the constructor call.
+     * @return Agent processor.
+     */
+    public static <T extends GridProcessor> T createProcessor(String clsName, Class<?>[] paramTypes, Object... args) {
+        ClassLoader ldr = U.gridClassLoader();
+
+        try {
+            Class<?> mgrCls = ldr.loadClass(clsName);
+
+            return (T)mgrCls.getConstructor(paramTypes).newInstance(args);
```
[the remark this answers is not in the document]
smoldachev: IDEA disagree with you :)

image

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ManagementConsoleAgent.java
```diff
@@ -87,17 +87,20 @@
...
     protected WebSocketManager mgr;
 
     /** Cluster processor. */
-    private ClusterInfoProcessor clusterProc;
+    protected ClusterInfoProcessor clusterProc;
 
-    /** Span exporter. */
-    private SpanExporter spanExporter;
+    /** Actions dispatcher. */
+    private ActionDispatcher actDispatcher;
```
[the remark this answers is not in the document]
smoldachev: I don't know, let's ask @nva

modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
```diff
@@ -226,6 +226,45 @@ public static void quiteStop(GridProcessor proc) {
...
+        }
+        catch (Exception ignored) {
+            // No-op.
```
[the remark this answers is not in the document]
smoldachev: Add warn log

modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
```diff
@@ -226,6 +226,45 @@ public static void quiteStop(GridProcessor proc) {
...
+
+        return null;
+    }
+
+    /**
+     * Quietly start given processor ignoring possible checked exception.
+     *
+     * @param proc Processor.
+     */
+    public static void quietStart(GridProcessor proc) {
```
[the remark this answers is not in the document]
smoldachev: Add warn log

modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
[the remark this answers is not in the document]
dmekhanikov: Apparently, our custom inspections ignore such warnings: https://github.com/ggprivate/ggprivate/blob/8.7-master/idea/gridgain_inspections.xml#L256

They weren't configured in my environment properly, sorry about that.
Please ignore the comment.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/ClusterInfoProcessor.java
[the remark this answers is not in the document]
dmekhanikov: Just use `super.buildClusterInfo()` to construct an instance of `ClusterInfo` in the ancestor.

modules/management-console-agent/src/main/java/org/apache/ignite/agent/utils/AgentUtils.java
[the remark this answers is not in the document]
dmekhanikov: It looks more like an error to me

modules/management-console-agent/src/main/java/org/apache/ignite/agent/ManagementConsoleAgent.java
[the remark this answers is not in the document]
nva: Revert

modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/ClusterInfoProcessor.java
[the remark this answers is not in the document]
nva: We can't do this because if we call `super.buildClusterInfo()` we construct `ClusterInfo` which doesn't contains additional fields. In this case we need to copy all it's fields to new `GridGainClusterInfo` which may leads to unexpected errors if we add some field to `ClusterInfo` and don't copy it to `GridGainClusterInfo`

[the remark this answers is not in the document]
dmekhanikov: Oh, yeah, you are right. Sorry, I missed that.
But it's still hard to tell the difference between these two methods. What's the difference between "create" and "build"?
One method creates an empty instance of `ClusterInfo`, another one creates **AND** populates it. But `GridGainClusterInfoProcessor#createClusterInfo` actually does both of those things, while it's supposed to do only creating, judging by the name.

I like the previous design when you had two separate methods for creating and populating the ClusterInfo instances more. I suggest falling back to it.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
```diff
@@ -214,14 +215,56 @@ public static void authorizeIfNeeded(IgniteSecurity security, String name, Secur
...
+        }
+        catch (Exception ex) {
+            U.warn(log, "Failed to initialize GridGain Control Center agent processor: " + clsName, ex);
+        }
+
+        return null;
+    }
+
+    /**
+     * Quietly start given processor ignoring possible checked exception.
```
dmekhanikov, line 256: It doesn't do it quietly.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
```diff
@@ -214,14 +215,56 @@ public static void authorizeIfNeeded(IgniteSecurity security, String name, Secur
...
+     * @param proc Processor.
+     * @param log Logger.
+     */
+    public static void startProcessor(GridProcessor proc, IgniteLogger log) {
+        if (proc != null) {
+            try {
+                proc.start();
+            }
+            catch (Exception ex) {
+                U.warn(log, ex.getMessage(), ex);
```
dmekhanikov, line 267: Impossibility to start a processor sounds more like an ERROR to me.

dmekhanikov: Also some message would be nice.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
```diff
@@ -214,14 +215,56 @@ public static void authorizeIfNeeded(IgniteSecurity security, String name, Secur
...
+    public static <T extends GridProcessor> T createProcessor(IgniteLogger log, String clsName, Class<?>[] paramTypes, Object... args) {
+        ClassLoader ldr = U.gridClassLoader();
+
+        try {
+            Class<?> mgrCls = ldr.loadClass(clsName);
+
+            return (T)mgrCls.getConstructor(paramTypes).newInstance(args);
+        }
+        catch (Exception ex) {
+            U.warn(log, "Failed to initialize GridGain Control Center agent processor: " + clsName, ex);
```
dmekhanikov, line 249: Looks more like an ERROR.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
```diff
@@ -214,14 +215,56 @@ public static void authorizeIfNeeded(IgniteSecurity security, String name, Secur
...
-    public static void quiteStop(GridProcessor proc) {
+    public static void stopProcessor(GridProcessor proc, IgniteLogger log) {
         if (proc != null) {
             try {
                 proc.stop(true);
             }
-            catch (Exception ignored) {
-                // No-op.
+            catch (Exception ex) {
+                U.warn(log, ex.getMessage(), ex);
```
dmekhanikov, line 226: I would say, that errors reported during stopping are not that critical. I would change the logging level to DEBUG. What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/management/ManagementConsoleSender.java
```diff
@@ -0,0 +1,34 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.management;
+
+/**
+ * Control Center sender.
+ */
+public interface ManagementConsoleSender {
```
dmekhanikov, line 22: Could you clarify in the javadoc, what this interface is for. Where is it supposed to be used?

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
```diff
@@ -214,14 +215,56 @@ public static void authorizeIfNeeded(IgniteSecurity security, String name, Secur
      * Quietly closes given processor ignoring possible checked exception.
```
dmekhanikov, line 215: It doesn't do it quietly.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
nva: Fixed

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
nva: Fixed

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
nva: Change log level

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
nva: I disagree with you. `IgniteLogger` with logging level to DEBUG can't print stacktrace. I think we need stacktrace for debug.

modules/core/src/main/java/org/apache/ignite/internal/processors/management/ManagementConsoleSender.java
nva: Improve javadoc and reuse this interface in all agent processors

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/utils/AgentUtils.java
nva: Fixed javadoc

modules/management-console-agent/src/main/java/org/apache/ignite/agent/processor/ClusterInfoProcessor.java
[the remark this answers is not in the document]
nva: Partial revert

modules/core/src/main/java/org/apache/ignite/internal/processors/management/ManagementConsoleSender.java
dmekhanikov: OK, I see. So, the issue is that we don't have such method in IgniteLogger. I created a ticket for this improvement: https://ggsystems.atlassian.net/browse/GG-28856
Such error handling will lead to a lot of exceptions printed when a node is stopped. As far as I can see, these errors can only occur when a node is stopped, so the reason behind it is almost always in something else. Let's keep it as warning for now.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/ws/WebSocketManager.java
```diff
@@ -336,7 +331,7 @@ private void addAuthentication(ClientManager mgr) {
             Map<String, String> proxyHeaders = new HashMap<>();
 
             proxyHeaders.put("Proxy-Authorization", "Basic " +
-                encodeToString((user + ':' + pwd).getBytes(Charset.forName("UTF-8"))));
+                encodeToString((user + ':' + pwd).getBytes(StandardCharsets.UTF_8)));
```
akuznetsov-gridgain, line 334: Lets use static import for StandardCharsets.UTF_8 ?

modules/core/src/main/java/org/apache/ignite/internal/processors/management/ControlCenterSender.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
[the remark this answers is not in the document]
nva: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/management/ManagementConsoleProcessor.java
```diff
@@ -35,4 +35,9 @@
      * @return Management configuration instance.
      */
     public ManagementConfiguration configuration();
+
+    /**
+     * @return Control Center sender.
+     */
+    public ControlCenterSender sender();
```
dmekhanikov, line 42: Please clarify, that this sender is responsible for communication with the backend.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/processor/ClusterInfoProcessor.java
```diff
@@ -65,18 +65,18 @@
     protected IgniteClusterEx cluster;
 
     /** Manager. */
```
dmekhanikov, line 67: The javadoc is wrong.

PR #1035, opened 2020-04-06
GG-28445 .NET: Use platform Near Cache for local ScanQuery

ptupitsyn:
Execute `ScanQuery` with `Local = true` and non-null `Partition` on .NET side when platform Near Cache is available, avoiding Java roundtrip. ~40x perf improvement.

Reserves partition same way as Java does in this case, so that correctness is guaranteed.


modules/core/src/test/java/org/apache/ignite/platform/PlatformIsPartitionReservedTask.java
```diff
@@ -0,0 +1,95 @@
...
+
+        /** {@inheritDoc} */
+        @Nullable @Override public Boolean execute() {
+            GridKernalContext ctx = ((IgniteEx) ignite).context();
+
+            GridDhtPartitionTopology top = ctx.cache().cache(cacheName).context().topology();
+
+            GridDhtLocalPartition locPart = top.localPartition(part, top.readyTopologyVersion(), false);
+
+            assert locPart != null;
```
isapego, line 90: This could actually happen if cache was just created and it has no partition for a ready topology version. I've got such exception when dealt with Partition Awareness. Need to handle properly without assertion. Also, I'm not sure, whether null is returned or exception is thrown in this case.

ptupitsyn: Note that this is test code. Are you worried about flaky tests?

PR #1037, opened 2020-04-07
GG-27289: SQLRowCount implemented for SELECT queries

isapego:


modules/platforms/cpp/odbc/src/query/data_query.cpp
```diff
@@ -181,7 +181,11 @@ namespace ignite
             int64_t DataQuery::AffectedRows() const
             {
                 int64_t affected = rowsAffectedIdx < rowsAffected.size() ? rowsAffected[rowsAffectedIdx] : 0;
-                return affected < 0 ? 0 : affected;
+
+                if (affected >= 0)
+                    return affected;
+
+                return connection.GetConfiguration().GetPageSize();
```
ashapkin, line 188: Just in case: different connections could yield different results for the same query.

isapego: This is an expected behavior. Also, this is not a query, this is API call.

modules/platforms/cpp/odbc-test/src/queries_test.cpp
```diff
@@ -1667,7 +1667,41 @@ BOOST_AUTO_TEST_CASE(TestAffectedRows)
...
+
+    // Just selecting everything to make sure everything is OK
+    SQLCHAR selectReq[] = "SELECT _key, strField FROM TestType ORDER BY _key";
+
+    SQLRETURN ret = SQLExecDirect(stmt, selectReq, sizeof(selectReq));
+
+    if (!SQL_SUCCEEDED(ret))
+        BOOST_FAIL(GetOdbcErrorMessage(SQL_HANDLE_STMT, stmt));
+
+    for (int i = 0; i < 200; ++i)
```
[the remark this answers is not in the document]
isapego: 123 is the number of records which transferred by a single network request-response and thus stay cached in driver. But records still fetched one by one by application. I've intentionally made number of fetches more than page size (to be sure paging is working properly with this feature), but less than the whole result set.

PR #1039, opened 2020-04-08
GG-27288: Implemented column-wise binding for result rows

isapego:


modules/platforms/cpp/odbc-test/include/odbc_test_suite.h
```diff
@@ -129,13 +139,176 @@ namespace ignite
...
+            /**
+             * Check floatField test value.
+             * @param idx Index.
+             * @param value Value to test.
+             */
+            static void CheckTestFloatValue(int idx, float value);
+
+            /**
+             * Get test doubleField.
+             *
```
ptupitsyn, line 219: This blank line is sometimes present, sometimes isn't

PR #1048, opened 2020-04-09
GG-26430 Collect TOP of N long SQL queries

korlov42:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/RunningQueryManager.java
```diff
@@ -109,6 +119,12 @@
      */
     private final AtomicLongMetric oomQrsCnt;
 
+    /** */
+    private final List<Consumer<GridQueryStartedInfo>> qryStartedListeners = new CopyOnWriteArrayList<>();
+
+    /** */
+    private final List<Consumer<GridQueryFinishedInfo>> qryFineshedListeners = new CopyOnWriteArrayList<>();
```
mshonichev, line 126: qryFin*i*shedListeners

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/IgniteSqlQueryStartFinishListenerTest.java
```diff
@@ -0,0 +1,495 @@
...
+            qryStarted.set(null);
+            qryFinished.set(null);
+        }
+
+        {
+            final String qry = "text query";
+
+            IgniteCache<String, Object> cache = grid(CLIENT_NODE_NAME).cache(DEFAULT_CACHE_NAME);
+
+            cache.query(new TextQuery<String, String>(String.class, "text query"));
```
mshonichev, line 237: I don't quite understand, if we issue `TextQuery` over cache got from client node, why do we later check that `startedInfo.nodeId()` equals to server node id?

korlov42: it is because nodeId represents an id of the node which starts execute the query, not an id of the node which sends request. Semantic is the same as for running query info (for running queries view)

PR #1049, opened 2020-04-10
GG-28417

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/internal/marshaller/optimized/OptimizedObjectStreamRegistry.java
```diff
@@ -30,27 +30,18 @@
  */
 class OptimizedObjectStreamRegistry {
     /** Holders. */
-    private static final ThreadLocal<StreamHolder> holders = new ThreadLocal<>();
+    private final ThreadLocal<StreamHolder> holders = new ThreadLocal<>();
```
AMashenkov, line 33: 'holders' are never cleaned-up from thread local map.
Is there any guarantee thread-local field is used within Ignite thread-pool's threads?

Recreate registry in same thread will result in creating new entry in thread-locals table, 
but seems, it can be reused.

Should we keep this field as static or cleanup thread local map? What do you think?

agoncharuk: @AMashenkov Agree that the slots can be reused in theory (this is how it worked before). I would not like though to mix up static and non-static approaches in the same class. Perhaps, we should have an abstract class with defined methods and two implementations: static cached and pooled non-static? What do you think?

AMashenkov: @agoncharuk , yes.
Having 2 different implementation will make code cleaner. Let's do it.

PR #1052, opened 2020-04-10
GG-26945

tkalkirill:


modules/core/src/test/java/org/apache/ignite/testframework/wal/record/RecordUtils.java
```diff
@@ -564,4 +564,14 @@ public static MvccTxRecord buildMvccTxRecord() {
...
         return new ConsistentCutRecord();
     }
+
+    /**
+     * Return {@code true} if append to write-ahead log.
+     *
+     * @param walRecord Instance of {@link WALRecord}.
+     * @return {@code True} if append to write-ahead log.
+     */
+    public static boolean isLogEnabled(WALRecord walRecord) {
```
denis-chudov, line 574: It's better to change this method's name and javadoc, now it's purpose is not obvious (e.g. "isIncludedInLog").

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/DataEntryWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+        Object value = unwrappedDataEntry.unwrappedValue();
+
+        if (HASH == sensitiveData) {
+            key = valueOf(key).hashCode();
+            value = valueOf(value).hashCode();
+        }
+        else if (MD5 == sensitiveData) {
+            key = md5(valueOf(key));
+            value = md5(valueOf(value));
+        }
```
denis-chudov, line 83: Don't you think that ProcessSensitiveData should be interface for strategy pattern?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -53,8 +74,12 @@ public static void main(String[] args) throws Exception {
         H2ExtrasInnerIO.register();
         H2ExtrasLeafIO.register();
 
-        boolean printRecords = IgniteSystemProperties.getBoolean("PRINT_RECORDS", false); //TODO read them from argumetns
-        boolean printStat = IgniteSystemProperties.getBoolean("PRINT_STAT", true); //TODO read them from argumetns
+        boolean printRecords = getBoolean(PRINT_RECORDS, false); //TODO read them from argumetns
+        boolean printStat = getBoolean("PRINT_STAT", true); //TODO read them from argumetns
+        ProcessSensitiveData sensitiveData = getEnum(SENSITIVE_DATA, SHOW); //TODO read them from argumetns
```
denis-chudov, line 79: Please create a ticket for this TODOs and link it to yours.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -34,10 +37,28 @@
...
+    /**
+     * System property for printing {@link WALRecord}. By default, {@code false}.
+     */
+    static final String PRINT_RECORDS = "PRINT_RECORDS";
+
+    /**
+     * System property for setting {@link ProcessSensitiveData strategy} of output sensitive data.
+     * By default, {@link ProcessSensitiveData#SHOW}.
+     */
+    static final String SENSITIVE_DATA = "SENSITIVE_DATA";
```
denis-chudov, line 60: Why do we need another system property for sensitive data? We can just rework IGNITE_TO_STRING_INCLUDE_SENSITIVE by adding new possible values.

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterSensitiveDataTest.java
```diff
@@ -0,0 +1,307 @@
...
+
+        GridKernalContext kernalCtx = crd.context();
+        IgniteWriteAheadLogManager wal = kernalCtx.cache().context().wal();
+
+        for (WALRecord walRecord : withSensitiveData()) {
+            if (isLogEnabled(walRecord))
+                wal.log(walRecord);
+        }
+
+        wal.flush(null, true);
```
denis-chudov, line 107: These tests look too synthetic with this manual write to WAL. Could you please rewrite it and add some real cache puts and SQL insert queries that include sensitive data?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/DataEntryWrapper.java
tkalkirill: We discussed it with @akalash and decided that we will define each type of record in our own way.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
tkalkirill: https://ggsystems.atlassian.net/browse/GG-28576

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
tkalkirill: We decided with @akalash that we would do something different and not change the current logic. This property is only used here.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1441,7 +1441,16 @@ private IgniteSystemProperties() {
         if (val == null)
             return dflt;
 
-        return Enum.valueOf(enumCls, val);
+        E e;
```
ibessonov, line 1444: Why do you declare this variable? You can just use 2 return statements instead.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/DataEntryWrapper.java
```diff
@@ -0,0 +1,88 @@
...
+            dataEntry.nearXidVersion(),
+            dataEntry.writeVersion(),
+            dataEntry.expireTime(),
+            dataEntry.partitionId(),
+            dataEntry.partitionCounter()
+        );
+
+        this.sensitiveData = sensitiveData;
+
+        this.unwrappedDataEntry = UnwrappedDataEntry.class.isInstance(dataEntry) ?
```
ibessonov, line 64: Why not "instanceof"?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -53,8 +74,12 @@ public static void main(String[] args) throws Exception {
         H2ExtrasInnerIO.register();
         H2ExtrasLeafIO.register();
 
-        boolean printRecords = IgniteSystemProperties.getBoolean("PRINT_RECORDS", false); //TODO read them from argumetns
-        boolean printStat = IgniteSystemProperties.getBoolean("PRINT_STAT", true); //TODO read them from argumetns
+        boolean printRecords = getBoolean(PRINT_RECORDS, false); //TODO read them from argumetns
+        boolean printStat = getBoolean("PRINT_STAT", true); //TODO read them from argumetns
```
ibessonov, line 78: Please make "PRINT_STAT" a constant as well.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
```diff
@@ -110,4 +135,29 @@ public static void main(String[] args) throws Exception {
...
+     *
+     * @param walRecord Instance of {@link WALRecord}.
+     * @param sensitiveData Strategy for processing of sensitive data.
+     * @return String representation of {@link WALRecord}.
+     */
+    private static String toString(WALRecord walRecord, ProcessSensitiveData sensitiveData) {
+        if (SHOW == sensitiveData || HIDE == sensitiveData)
+            return walRecord.toString();
+
+        if (MetastoreDataRecord.class.isInstance(walRecord))
```
ibessonov, line 150: Same here, why not "instanceof"?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/MetastoreDataRecordWrapper.java
```diff
@@ -0,0 +1,48 @@
...
+     * Constructor.
+     *
+     * @param metastoreRecord Instance of {@link MetastoreDataRecord}.
+     * @param sensitiveData Strategy for the processing of sensitive data.
+     */
+    public MetastoreDataRecordWrapper(MetastoreDataRecord metastoreRecord, ProcessSensitiveData sensitiveData) {
+        super(
+            HASH == sensitiveData ? valueOf(metastoreRecord.key().hashCode()) :
+                MD5 == sensitiveData ? md5(metastoreRecord.key()) : metastoreRecord.key(),
+            metastoreRecord.value()
```
ibessonov, line 40: Value is not obscured, why? It can have encryption keys and all this stuff...

modules/dev-utils/src/test/java/org/apache/ignite/development/utils/IgniteWalConverterSensitiveDataTest.java
```diff
@@ -0,0 +1,382 @@
...
+/**
+ * Class for testing sensitive data when reading {@link WALRecord} using
+ * {@link IgniteWalConverter}.
+ */
+public class IgniteWalConverterSensitiveDataTest extends GridCommonAbstractTest {
+    /** Sensitive data prefix. */
+    private static final String SENSITIVE_DATA_VALUE_PREFIX = "must_hide_it_";
+
+    /** Path to directory where WAL is stored. */
+    private static String WAL_DIR_PATH;
```
ibessonov, line 75: This naming convention is not typical, upper case is for constants only, right?

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/DataEntryWrapper.java
tkalkirill: There's no overhead, it's just my style.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverter.java
tkalkirill: There's no overhead, it's just my style.

modules/dev-utils/src/main/java/org/apache/ignite/development/utils/DataEntryWrapper.java
ibessonov: Well, technically there is an overhead until the moment when JIT realizes that this code can be replaced with "instanceof", but that wasn't the reason why I mentioned this code :)

PR #1062, opened 2020-04-13
GG-26434 Security permissions for VIEWS and KILL QUERY command

korlov42:


modules/core/src/main/java/org/apache/ignite/internal/jdbc2/JdbcStatement.java
```diff
@@ -142,7 +143,7 @@ else if (!conn.isMultipleStatementsAllowed() && conn.isMultipleStatementsTaskV2S
             }
         }
         catch (Exception e) {
-            throw convertToSqlException(e, "Failed to query Ignite.");
+            throw convertToSqlException(e, e instanceof SecurityException ? e.getMessage() : "Failed to query Ignite.");
```
AMashenkov, line 146: Can we move "e instanceof SecurityException ? e.getMessage()" into convertToSqlException()  ?

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientListenerAbstractConnectionContext.java
```diff
@@ -162,6 +162,9 @@ protected void initClientDescriptor(String prefix) {
 
         if (authCtx != null)
             clientDesc += "@" + authCtx.userName();
+
```
AMashenkov, line 165: Empty line

PR #1067, opened 2020-04-14
GG-28558 Return the default data region name when the cache data region is not configured.

vsisko:


modules/core/src/main/java/org/apache/ignite/internal/visor/cache/VisorCacheConfiguration.java
```diff
@@ -225,7 +225,9 @@ public VisorCacheConfiguration(IgniteEx ignite, CacheConfiguration ccfg, IgniteU
         evictFilter = compactClass(ccfg.getEvictionFilter());
         lsnrConfigurations = compactIterable(ccfg.getCacheEntryListenerConfigurations());
         loadPrevVal = ccfg.isLoadPreviousValue();
-        dataRegName = ccfg.getDataRegionName();
+        dataRegName = ccfg.getDataRegionName() != null
+            ? ccfg.getDataRegionName()
+            : ignite.configuration().getDataStorageConfiguration().getDefaultDataRegionConfiguration().getName();
```
MaxLosevskoy, line 230: Is there a chance for NPE?
e.g. if `getDataStorageConfiguration()` returns `null`

vsisko: It has default configuration. See org/apache/ignite/internal/IgnitionEx.java:2383

PR #1076, opened 2020-04-15
GG-27581 .NET: Add partition-based AffinityCall and AffinityRun overloads

ptupitsyn:
* Add AffinityCall and AffinityRun overloads that take multiple cache names and partition
* Fix existing AffinityCall and AffinityRun: use proper Java APIs that reserve the partition instead of custom Task-based implementation


modules/platforms/dotnet/Apache.Ignite.Core/Compute/ICompute.cs
```diff
@@ -532,6 +567,38 @@ public interface ICompute
         /// <returns>Task.</returns>
         Task AffinityRunAsync(string cacheName, object affinityKey, IComputeAction action, 
             CancellationToken cancellationToken);
+        
+        /// <summary>
+        /// Executes given job on primary node for the given partition.
+        /// (a.k.a. affinity co-location).
```
ashapkin, line 573: Let's be consistent with AffinityCall

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/compute/PlatformBalancingSingleClosureAffinityTask.java
```diff
@@ -31,6 +31,8 @@
 
 /**
  * Interop single-closure task with node balancing.
+ *
+ * NOTE: This class is not used, but preserved for compat during rolling upgrades.
```
ashapkin, line 35: This class doesn't transfer through the wire, why should we take care of it during RU?

modules/platforms/dotnet/Apache.Ignite.Core/Compute/ICompute.cs
ptupitsyn: Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/compute/PlatformBalancingSingleClosureAffinityTask.java
ptupitsyn: Good catch, fixed

PR #1079, opened 2020-04-16
GG-23133

zstan:


modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+import static org.apache.ignite.internal.commandline.cache.CacheCommands.usageCache;
+import static org.apache.ignite.internal.commandline.cache.CacheSubcommands.CHECK_INDEX_INLINE_SIZES;
+
+/**
+ * Command for check secondary indexes inline size on the different nodes.
+ */
+public class CheckIndexInlineSizes implements Command<Void> {
+    /** Warn message format. */
+    public static final String NOT_ALL_NODES_SUPPORT_FEATURE_WARN_MSG_FMT =
+        "Indexes inline size were not checked on all nodes because the few nodes aren't support this feature. Skipped nodes: %s";
```
zstan, line 56: too long as for me: Inline size has been completed on limited cluster nodes .... something like this.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /** Success message. */
+    public static final String INDEXES_INLINE_SIZE_ARE_SAME =
+        "All secondary indexes have the same effective inline size on all cluster nodes.";
```
zstan, line 60: we have no secondary indexes, if i`m wrong: show mw documentation plz.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+
+    /** Predicate to filter server nodes. */
+    private static final Predicate<GridClientNode> SRV_NODES = node ->
+        Objects.equals(node.attribute(ATTR_CLIENT_MODE), false) && !"true".equals(node.attribute(ATTR_DAEMON));
```
zstan, line 64: it`s ok, but as for me would be more clear to use : DaemonFilter , as in org.apache.ignite.cluster.ClusterGroup#forDaemons, u need simle to exclude them.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+            CheckIndexInlineSizesResult res = client.compute().projection(supportedServerNodes).execute(
+                CheckIndexInlineSizesTask.class.getName(),
+                new VisorTaskArgument<>(supportedServerNodeIds, false)
+            );
+
+            Set<UUID> unsupportedNodes = serverNodeIds.stream()
+                .filter(n -> !supportedServerNodeIds.contains(n))
+                .collect(toSet());
+
+            analizeResults(log, unsupportedNodes, res);
```
zstan, line 89: typo: analyze

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+        }
+
+        return null;
+    }
+
+    /** */
```
zstan, line 95: sorry, but informative comment need to be here )

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+
+        for (Map.Entry<UUID, Map<String, Integer>> nodeRes : res.inlineSizes().entrySet()) {
+            for (Map.Entry<String, Integer> index : nodeRes.getValue().entrySet()) {
+                Map<Integer, Set<UUID>> sizeToNodes = indexToSizeNode.computeIfAbsent(index.getKey(), x -> new HashMap<>());
+
+                sizeToNodes.computeIfAbsent(index.getValue(), x -> new HashSet<>()).add(nodeRes.getKey());
+            }
+        }
+
+        log.info("Found " + indexToSizeNode.size() + " secondary indexes.");
```
zstan, line 114: 1. one more "secondary" word )
2. and what about grid with PK indexes only ? empty result will be printing, isn`t it ?

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+
+        Map<String, Map<Integer, Set<UUID>>> problems = indexToSizeNode.entrySet().stream()
+            .filter(e -> e.getValue().size() > 1)
+            .collect(Collectors.toMap(Map.Entry::getKey, Map.Entry::getValue));
+
+        if (F.isEmpty(problems))
+            log.info(INDEXES_INLINE_SIZE_ARE_SAME);
```
zstan, line 121: are THE same, looks like correct.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+        else
+            printProblemsAndShowRecommendations(problems, log);
+    }
+
+    /** */
+    private void printProblemsAndShowRecommendations(Map<String, Map<Integer, Set<UUID>>> problems, Logger log) {
+        log.info(problems.size() +
+            " indexes have different effective inline size on nodes. It can lead to performance degradation in SQL queries.");
```
zstan, line 129: and what if it only one index ?

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+        log.info("Indexes:");
+
+        for (Map.Entry<String, Map<Integer, Set<UUID>>> entry : problems.entrySet()) {
+            SB sb = new SB();
+
+            sb.a("Full index name: ").a(entry.getKey());
+
+            for (Integer size : entry.getValue().keySet())
+                sb.a(" nodes: ").a(entry.getValue().get(size)).a(" inline size: ").a(size).a(",");
```
zstan, line 138: column view here looks like more readable than string representation.

zstan: can someone understand what nodes are talking about ?

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,176 @@
...
+
+            sb.setLength(sb.length() - 1);
+
+            log.info(INDENT + sb);
+        }
+
+        log.info("");
+
+        log.info("Recommendations:");
+        log.info(INDENT + "Check that value of property " + IgniteSystemProperties.IGNITE_MAX_INDEX_PAYLOAD_SIZE + " are the same on all nodes.");
```
zstan, line 148: 1. which low or high inline size need to be applied ? 
2. can u describe recreation procedure ? may be more user friendly here ?

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -340,27 +348,121 @@ public void onCacheReconnect() throws IgniteCheckedException {
...
+    }
+
+    /**
+     * @return Serialiazible information about secondary indexes inline size.
+     * @see #secondaryIndexesInlineSize()
+     */
+    private Serializable collectSecondaryIndexesInlineSize() {
+        Map<String, Integer> map = secondaryIndexesInlineSize();
+
+        return map instanceof Serializable ? (Serializable)map : new HashMap<>(map);
```
zstan, line 465: not serializable map ? plz clarify here ?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
```diff
@@ -4382,20 +4384,11 @@ private UUID safeLocalNodeId() {
      * @return {@code True} if remote nodes support {@link HandshakeWaitMessage}.
      */
     private boolean isHandshakeWaitSupported() {
```
zstan, line 4386: what is the purpose for this change ?

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
antonovsergey93: A secondary index means not a primary or affinity key index. Look at org/apache/ignite/internal/processors/query/h2/database/H2Tree.java:699

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
antonovsergey93: Yes, we check only secondary indexes. primary and affinity key indexes will be skipped during the check.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
antonovsergey93: Do you mean a plural form of index**es**?

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
antonovsergey93: > column view here looks like more readable than string representation.
I'm not sure. I made a usability shield in the ticket's comments. CS, QA and Core teams are ok with current output. I'd prefer to keep it as is.

antonovsergey93: > can someone understand what nodes are talking about ?

It can be helpful for investigation problems. Also, the user can check IGNITE_MAX_INDEX_PAYLOAD_SIZE property on the selected nodes.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
antonovsergey93: >which low or high inline size need to be applied ?

It depends on the user's choice. The property disabled by default, if the user set it you made some calculations and he knows right property value.

antonovsergey93: > can u describe recreation procedure ? may be more user friendly here ?

Done

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
antonovsergey93: Map interface doesn't extend java.io.Serializable interface. We could get here some custom no serializable implementation

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/TcpCommunicationSpi.java
antonovsergey93: It's common logic for check feature support. I move it to the IgniteFeatures class.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
zstan: https://www.google.ru/search?client=opera&q=apache+ignite+secondary+index&sourceid=opera&ie=UTF-8&oe=UTF-8 google show nothing )

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
zstan: yep )

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -340,27 +348,121 @@ public void onCacheReconnect() throws IgniteCheckedException {
...
-                for (SchemaProposeDiscoveryMessage activeProposal : data0.values())
+        // Process proposals as if they were received as regular discovery messages.
+        if (!F.isEmpty(activeProposals)) {
+            synchronized (stateMux) {
+                for (SchemaProposeDiscoveryMessage activeProposal : activeProposals.values())
                     onSchemaProposeDiscovery0(activeProposal);
             }
         }
+
+        if (!F.isEmpty(data.nodeSpecificData())) {
```
zstan, line 392: voice calling, just remind u, give opportunity for further usage this stuff.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -340,27 +348,121 @@ public void onCacheReconnect() throws IgniteCheckedException {
...
+            Map<String, Integer> indexesInlineSize = secondaryIndexesInlineSize();
+
+            if (!F.isEmpty(indexesInlineSize)) {
+                Map<UUID, Serializable> nodeSpecific = data.nodeSpecificData();
+
+                for (UUID nodeId : nodeSpecific.keySet()) {
+                    if (nodeSpecific.get(nodeId) instanceof Map)
+                        checkInlineSizes(indexesInlineSize, (Map<String, Integer>)nodeSpecific.get(nodeId), nodeId);
```
zstan, line 400: && !(Map)nodeSpecific.get(nodeId).isEmpty() add little gc reduce fix

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
```diff
@@ -0,0 +1,170 @@
...
+    /** {@inheritDoc} */
+    @Override public Object execute(GridClientConfiguration clientCfg, Logger log) throws Exception {
+        try (GridClient client = Command.startClient(clientCfg)) {
+            Set<GridClientNode> serverNodes = client.compute().nodes().stream()
+                .filter(SRV_NODES)
+                .collect(toSet());
+
+            Set<GridClientNode> supportedServerNodes = serverNodes.stream()
+                .filter(n -> n.supports(CHECK_INDEX_INLINE_SIZES))
+                .collect(toSet());
```
MaxLosevskoy, line 68: Is that fine that we omit server nodes that do not support the feature? Than we don't check inline sizes there, isn't that a risk of inconsistency with false-positive check?

MaxLosevskoy: Though I can see `log.info()` call for that case

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -130,6 +135,12 @@
  * Indexing processor.
  */
 public class GridQueryProcessor extends GridProcessorAdapter {
+    /** */
+    private static final String INLINE_SIZES_DISCO_BAG_KEY = "inline_sizes";
+
+    /** Warn message if some indexes have different inline sizes on the nodes. */
+    public static final String INLINE_SIZES_DIFFER_WARN_MSG_FORMAT = "Inline sizes on local node and node %s are different. Please drop and create again these indexes, for avoinding performance problems with SQL queries. Problem indexes: %s";
```
MaxLosevskoy, line 142: Typo: `avoiding`
I propose to **replace** with `...indexes to avoid performance problems...`

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -340,27 +351,147 @@ public void onCacheReconnect() throws IgniteCheckedException {
...
+
+    /**
+     * Checks that server node with {@code nodeId} is supported {@link IgniteFeatures#CHECK_INDEX_INLINE_SIZES}.
+     *
+     * @param nodeId Node id.
+     * @return {@code true} if node with {@code nodeId} is server node and the node supports feature, {@code false}
+     * otherwise.
+     */
+    private boolean nodeSupportedCheckIndexInlineSize(UUID nodeId) {
+        IgnitePredicate<ClusterNode> nodeFilter = n -> !n.isClient() && !n.isDaemon() && nodeId.equals(n.id());
```
ibessonov, line 449: "nodeSupportedCheckIndexInlineSize" will return "true" for client nodes, because "allNodesSupport" for empty collection of nodes is "true". Please change javaDoc or fix the method.

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CheckIndexInlineSizes.java
antonovsergey93: Already done. Look at CheckIndexInlineSizes#analyzeResults:101 line.

PR #1082, opened 2020-04-17
GG-28549 Switch off property IGNITE_DISABLE_REBALANCING_CANCELLATION_…

ascherbakoff:
…OPTIMIZATION


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPreloader.java
```diff
@@ -72,7 +72,7 @@
 
     /** Disable rebalancing cancellation optimization. */
     private final boolean disableRebalancingCancellationOptimization = IgniteSystemProperties.getBoolean(
-        IGNITE_DISABLE_REBALANCING_CANCELLATION_OPTIMIZATION, true);
+        IGNITE_DISABLE_REBALANCING_CANCELLATION_OPTIMIZATION, false);
```
ascherbakoff, line 75: You don't need to specify false - it is default.

PR #1085, opened 2020-04-20
GG-25705: Add kill query action.

smoldachev:


modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
```diff
@@ -89,6 +89,27 @@ public QueryActionsController(GridKernalContext ctx) {
         return new IgniteFinishedFutureImpl<>();
     }
 
+    /**
+     * @param globalQryId Global query id.
+     * @return Kill query result.
+     */
+    public IgniteInternalFuture<QueryResult> kill(String globalQryId) {
+        return ctx.closure().callLocalSafe(() -> {
+            SqlFieldsQuery qryFields = new SqlFieldsQuery("KILL QUERY '" + globalQryId + "';");
```
nva, line 98: Can we receive SQL injections here?

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/dto/action/query/RunningQuery.java
```diff
@@ -23,9 +23,12 @@
  * Descriptor of running query.
  */
 public class RunningQuery {
-    /** */
+    /** ID. */
```
nva, line 26:
```suggestion
    /** Query ID from request. */
```

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
dmekhanikov: I don't think we should care too much about this here. If the user has access to killing queries, she probably can execute them as well.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
```diff
@@ -89,6 +89,27 @@ public QueryActionsController(GridKernalContext ctx) {
...
+
+            try (
+                FieldsQueryCursor<List<?>> cur = qryProc.querySqlFields(qryFields, true);
+                CursorHolder cursorHolder = new CursorHolder(cur)
+            ) {
+                QueryResult res = fetchSqlQueryResult(cursorHolder, 1);
+
+                res.setResultNodeId(ctx.localNodeId().toString());
+
+                return res;
```
dmekhanikov, line 108: Is the result going to be used the frontend?

modules/management-console-agent/src/test/java/org/apache/ignite/internal/agent/action/controller/QueryActionsControllerTest.java
```diff
@@ -388,6 +391,62 @@ public void shouldCancelLongQuery() {
         );
     }
 
+    /**
+     * 1. Execute a query with sleep function.
+     * 2. Send kill query action with global query id for a query from 1.
+     * 3. Assert that action was completed and no one running queries exists in a cluster.
+     */
+    @Test
+    public void shouldKillRunningQuery() {
```
dmekhanikov, line 400: Please add a test that kill a query started from Ignite directly without using GGCC API.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
smoldachev: We have a special action for execution any SQL query, :)

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
smoldachev: I think no, make it as cancel action :)

modules/management-console-agent/src/test/java/org/apache/ignite/internal/agent/action/controller/QueryActionsControllerTest.java
smoldachev: Added

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/dto/action/query/RunningQuery.java
smoldachev: Done.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
```diff
@@ -89,6 +89,19 @@ public QueryActionsController(GridKernalContext ctx) {
         return new IgniteFinishedFutureImpl<>();
     }
 
+    /**
+     * @param globalQryId Global query id.
+     * @return Kill query result.
+     */
+    public IgniteInternalFuture<?> kill(String globalQryId) {
+        return ctx.closure().runLocalSafe(() -> {
```
akuznetsov-gridgain, line 97: Should we have checks features that "KILL QUERY" is supported by node?

nva: May be we should define some interceptor for all actions that can validate feature/version from annotation on action method?

smoldachev: Why we should check it on the agent? I think we should check that we can run some action on UI as we do it with baseline auto-adjust.

modules/management-console-agent/src/main/java/org/apache/ignite/internal/agent/action/controller/QueryActionsController.java
```diff
@@ -89,6 +89,19 @@ public QueryActionsController(GridKernalContext ctx) {
...
+            SqlFieldsQuery qryFields = new SqlFieldsQuery("KILL QUERY '" + globalQryId + "';")
+                .setLazy(false);
+
+            qryProc.querySqlFields(qryFields, true).close();
+        }, MANAGEMENT_POOL);
+    }
```
akuznetsov-gridgain, line 103: May be it make sense to introduce base class for ActionsControllers?
And declare protected method "safeExecute(lambda)" that will accept function to execute?

smoldachev: @nva What do you think about it? If yes then i create a separate ticket for this and change in all places  where it need.

akuznetsov-gridgain: It will reduce number of boilerplate code. IMHO.

nva: Let's do this in separated ticket.

smoldachev: Ticket: GG-28865

PR #1091, opened 2020-04-21
GG-28826 [IGNITE-12468] Java thin: deserialization of collections fixed

isapego:
(cherry picked from commit 4433485d74adc109ac5f435569e18578bc9f351f)

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
@@ -48,12 +48,18 @@ import org.apache.ignite.cache.query.SqlFieldsQuery;
 import org.apache.ignite.client.ClientCacheConfiguration;
 import org.apache.ignite.internal.binary.BinaryFieldMetadata;
 import org.apache.ignite.internal.binary.BinaryMetadata;
+import org.apache.ignite.internal.binary.BinaryObjectImpl;
 import org.apache.ignite.internal.binary.BinaryRawWriterEx;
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
```diff
@@ -495,12 +501,73 @@ void writeObject(BinaryOutputStream out, Object obj) {
...
+            BinaryReaderHandles hnds = new BinaryReaderHandles();
+
+            return (T)unwrapBinary(marsh.deserialize(in, hnds), hnds);
+        }
+    }
+
+    /**
+     * Unwrap binary object.
+     */
+    private Object unwrapBinary(Object obj, BinaryReaderHandles hnds) {
```
ptupitsyn, line 516: This does not look correct to me:

1. Why do we use special unmarshal logic for thin client? It is supposed to be reused from thick, right?
2. We should not need to unwrap things like that. When `keepBinary` is `false`, we should deserialize correctly right away, without intermediate `BinaryObject`s that need to be unwrapped.

isapego: Well, now we need to decide what should we do with this, considering this is backport from Ignite.

isapego: I propose to properly fix this in separate patch so it can be donated in Ignite. What do you think?

ptupitsyn: Agree.

PR #1093, opened 2020-04-21
GG 27804

zstan:


modules/core/src/test/java/org/apache/ignite/internal/GridDiscoveryManagerChangeCoordinatorTest.java
```diff
@@ -0,0 +1,197 @@
...
+     * Checks that there are no messages like "coordinator changed" after client left topology.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testChangeCoordinatorLogging() throws Exception {
+        //Start 2 server nodes
+        IgniteEx srv1 = startGrid("server1");
+        startGrid("server2");
+
```
zstan, line 75: hope would be a little bit simpler to start client between 2 servers run ? s1 -> c1 -> s2 isn`t it ?

modules/core/src/test/java/org/apache/ignite/internal/GridDiscoveryManagerChangeCoordinatorTest.java
```diff
@@ -0,0 +1,197 @@
...
+        stopGrid("server1");
+        srv1 = startGrid("server1");
+
+        stopGrid("server2");
+        startGrid("server2");
+
+        UUID clientClusterNode = client.localNode().id();
+        UUID srv1ClusterNode = srv1.localNode().id();
+
+        Pattern ptrn = Pattern.compile(String.format(CRD_CHANGE_MSG, srv1ClusterNode, clientClusterNode));
```
zstan, line 89: purpose of ticket - not to trigger : "Coordinator changed" at all, thus would be more clear to change regep into to more common pattern : "Coordinator changed" and enough.

modules/core/src/test/java/org/apache/ignite/internal/GridDiscoveryManagerChangeCoordinatorTest.java
```diff
@@ -0,0 +1,197 @@
...
+
+        srv1 = startGrid("server1");
+
+        crdUUID = ((TcpDiscoverySpi) srv3.context().config().getDiscoverySpi()).getCoordinator();
+
+        // Server2 is still the coordinator
+        assertEquals(srv2.localNode().id(), crdUUID);
+
+        // Check that there are no messages like "coordinator changed server2 -> server1"
+        assertFalse(lsnr.check());
```
zstan, line 155: same as upper: no one occurrence of "Coordinator changed" need to be detected here.

PR #1095, opened 2020-04-21
Gg 27840 2

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
```diff
@@ -1624,12 +1626,18 @@ private boolean shouldOverridePartitionMap(GridDhtPartitionMap currentMap, GridD
                         else if (state == MOVING) {
                             GridDhtLocalPartition locPart = locParts.get(p);
 
+                            if (locPart.state() == OWNING)
```
glukos, line 1629: Do we need to calculate fut.affinityReassign flag in a such complex way?
There's a special condition on which we decide on coordinator that we should assign partition state explicitly in case of TX DR (see finishExchangeOnCoordinator):

                else if (discoveryCustomMessage instanceof SnapshotDiscoveryMessage
                        && ((SnapshotDiscoveryMessage)discoveryCustomMessage).needAssignPartitions())
                    assignPartitionsStates();

If above is true, we may save a flag in exchange future that related GridDhtPreloaderAssignments should be incompatible.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPreloaderAssignments.java
```diff
@@ -38,16 +38,31 @@
     /** */
     private boolean cancelled;
 
+    /** Some of owned by affinity partitions were changed state to moving. */
+    private volatile boolean affinityReassign;
```
glukos, line 42: Can be final

vldpyatkov: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/GridDhtPartitionTopologyImpl.java
vldpyatkov: If we do that, a case when partition changing state (from OWNING to MOVING) by coordinator's reason would not be handled.

PR #1120, opened 2020-04-29
GG-28961 Actions can't be executed on cluster without ignite-opencensus in classpath

nva:


modules/management-console-agent/pom.xml
```diff
@@ -105,15 +105,9 @@
         </dependency>
 
         <dependency>
-            <groupId>org.reflections</groupId>
-            <artifactId>reflections</artifactId>
+            <groupId>net.oneandone.reflections8</groupId>
+            <artifactId>reflections8</artifactId>
```
akuznetsov-gridgain, line 109: 👍

modules/management-console-agent/src/test/java/org/apache/ignite/internal/agent/dto/topology/TopologySnapshotTest.java
```diff
@@ -28,6 +27,7 @@
 import org.apache.ignite.spi.discovery.tcp.internal.TcpDiscoveryNode;
 import org.junit.Test;
 
+import static org.apache.ignite.internal.util.lang.GridFunc.asList;
```
akuznetsov-gridgain, line 30: May be it will be more natural to "import F" ?
And use in code `F.asList(....)` ?

modules/management-console-agent/src/test/java/org/apache/ignite/internal/agent/processor/export/SpanExporterTest.java
```diff
@@ -41,6 +40,7 @@
 
 import static io.opencensus.trace.TraceOptions.DEFAULT;
 import static org.apache.ignite.internal.agent.ManagementConsoleAgent.TOPIC_MANAGEMENT_CONSOLE;
+import static org.apache.ignite.internal.util.lang.GridFunc.asList;
```
akuznetsov-gridgain, line 43: Same as previous, how about "import F"?

PR #1128, opened 2020-05-04
GG-28494

sanpwc:


modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
```diff
@@ -0,0 +1,303 @@
...
+        DistributedMetaStorage metaStore;
+
+        try {
+            metaStore = ctx.distributedMetastorage();
+        }
+        catch (Exception e) {
+            log.warning("Failed to restore tracing configuration for coordinates=[" + coordinates +
+                "] to default  — meta storage is not available.");
+
+            return false;
```
glukos, line 256: I'd make all API methods void and wrap original exception into IgniteException and throw it out of the method.
I think it would be more transparent if user could observe the original reason why API call failed instead of searching it in the logs.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
```diff
@@ -0,0 +1,303 @@
...
+                    return true;
+                }
+
+                HashMap<String, TracingConfigurationParameters> updatedScopeSpecificTracingConfiguration =
+                        new HashMap<>(existingScopeSpecificTracingConfiguration);
+
+                if (coordinates.label() != null)
+                    updatedScopeSpecificTracingConfiguration.remove(coordinates.label());
+                else
+                    updatedScopeSpecificTracingConfiguration.remove(null);
```
glukos, line 286: Null mappings in Map always look suspicious.
Also, there's no simple way to drop all label-specific configurations in case we created lots of them.
I have a proposal. We can use two preset label names: DEFAULT (meaning that this configuration coordinate is applicable to all labels, for which label-specific configuration wasn't set) and ANY (meaning that this configuration coordinate is applicable to the whole scope).
DEFAULT will be stored as map key instead of null, ALL will be never present in configuration, but could be utilized for mass updates.
Examples:


        new TracingConfigurationCoordinates.Builder(Scope.TX).withLabel(DEFAULT).build();
        new TracingConfigurationCoordinates.Builder(Scope.TX).build();
        // ^ these two configuration coordinates are equal

        client.tracingConfiguration().addConfiguration(
            new TracingConfigurationCoordinates.Builder(Scope.TX).withLabel(ANY).build(),
            new TracingConfigurationParameters.Builder().withSamplingRate(0.5).build()
        );
        // ^ this code will remove all label-specific configurations and will set sampling rate for the whole TX scope to 0.5

        client.tracingConfiguration().restoreDefaultConfiguration(
            new TracingConfigurationCoordinates.Builder(Scope.TX).withLabel(ANY).build()
        );
        // ^ this code will restore defaults to the whole TX scope

        client.tracingConfiguration().retrieveConfiguration(
            new TracingConfigurationCoordinates.Builder(Scope.TX).withLabel(ANY).build()
        );
        // ^ this code will throw exception. "ANY" label can't be retrieved!

What do you think?
Probably, I'm over-engineering it, but I think need to wipe all label-specific configurations with a single command is likely to appear at least once.

modules/opencensus/src/test/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusTxTracingConfigurationTest.java
```diff
@@ -0,0 +1,234 @@
...
+        handler().flush();
+
+        java.util.List<SpanData> gotSpans = handler().allSpans()
+            .filter(span -> SpanType.TX.traceName().equals(span.getName()))
+            .collect(Collectors.toList());
+
+        // Cause of probability nature of sampling it's not possible to check that 0.5 sampling rate will end with
+        // 5 sampling transactions out of {@code txAmount},
+        // so we just check that some and not all transactions were traced.
+        assertTrue(!gotSpans.isEmpty() && gotSpans.size() < txAmount);
```
glukos, line 101: This test will have ~0.2% flaky rate.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
```diff
@@ -0,0 +1,303 @@
...
+     * @param ctx Context.
+     */
+    public GridTracingConfiguration(@NotNull GridKernalContext ctx) {
+        this.ctx = ctx;
+
+        log = ctx.log(getClass());
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean addConfiguration(
```
glukos, line 85: Personally, I'd name it "applyConfiguration"

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
```diff
@@ -0,0 +1,303 @@
...
+            catch (IgniteCheckedException e) {
+                LT.warn(log, "Failed to retrieve tracing configuration");
+            }
+        }
+
+        return res;
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean restoreDefaultConfiguration(@NotNull TracingConfigurationCoordinates coordinates) {
```
glukos, line 246: Personally, I'd name it "resetConfiguration"

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
sanpwc: I believe that it's less intuitive and more complicated.

- Originally DEFAULT has different meaning - scope specific parameters that were predefined by us, not the user. So, using DEFAULT as scope specific,as you suggest, leads us to the necessity of introducing some default-default label.

- There's also a disadvantage in using predefined labels cause we actually change existing behavior, cause users may already have 'DEFAULT, ANY, ALL' labels for transactions.

- To be honest, I don't think that using null key is bad practice here. I believe that we might add some comment to clarify null's meaning as key.

- However I agree that it'll be useful to have an ability to drop all label-specific configurations for some scope, but I'd rather implement that with new api method.

All in all, it's discussable.

modules/opencensus/src/test/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusTxTracingConfigurationTest.java
sanpwc: Agreed, let it be 100 transactions instead of 10.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
sanpwc: Agreed.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
sanpwc: Agreed.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfiguration.java
sanpwc: It's not possible to make all of them void, cause 2 out of 4 return configuration, however void instead of boolean definitely has sense.
> wrap original exception into IgniteException and throw it out of the method.
Agreed.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/TracingConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+ *      https://www.gridgain.com/products/software/community-edition/gridgain-community-edition-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
```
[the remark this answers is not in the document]
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/TracingConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+
+import org.apache.ignite.IgniteException;
+import org.apache.ignite.internal.processors.tracing.Scope;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Allows to configure tracing, read the configuration and restore it to the defaults.
+ */
+public interface TracingConfiguration {
```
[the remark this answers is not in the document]
sanpwc: Agreed. Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/TracingConfiguration.java
```diff
@@ -0,0 +1,169 @@
...
+     * Apply new tracing configuration for the specific tracing coordinates (scope, label, etc.).
+     * If tracing configuration with specified coordinates already exists it'll be overrided,
+     * otherwise new one will be created.
+     *
+     * @param coordinates {@link TracingConfigurationCoordinates} Specific set of locators like {@link Scope} and label,
+     *  that defines subset of traces and/or spans that'll use given configuration.
+     * @param parameters {@link TracingConfigurationParameters} e.g. sampling rate, set of supported scopes etc.
+     * @throws IgniteException If failed to apply tracing configuration.
+     */
+    void apply(@NotNull TracingConfigurationCoordinates coordinates,
```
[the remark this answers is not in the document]
sanpwc: I don't think it's a good idea, cause it's exactly Tracing**Configuration**Coordinates.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/NoopTracingSpi.java
```diff
@@ -23,6 +23,8 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
+import java.util.Set;
+
```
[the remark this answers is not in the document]
sanpwc: I'm not sure. Currently there are 1.5 OpenCensus and Noop. However it's not the part of tracing configuration))

modules/opencensus/src/test/java/org/apache/ignite/TracingConfigurationValidationTest.java
```diff
@@ -0,0 +1,123 @@
...
+
+import static org.apache.ignite.internal.processors.tracing.Scope.COMMUNICATION;
+import static org.apache.ignite.internal.processors.tracing.Scope.DISCOVERY;
+import static org.apache.ignite.internal.processors.tracing.Scope.EXCHANGE;
+import static org.apache.ignite.internal.processors.tracing.Scope.TX;
+
+/**
+ * Tests for tracing configuration validation rules.
+ */
+@SuppressWarnings("ThrowableNotThrown") public class TracingConfigurationValidationTest extends GridCommonAbstractTest {
```
glukos, line 40: @SuppressWarning would look better on a separate line

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfigurationManager.java
```diff
@@ -0,0 +1,358 @@
...
+import org.apache.ignite.internal.util.tostring.GridToStringExclude;
+import org.apache.ignite.internal.util.typedef.internal.LT;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Tracing configuration manager implementation that uses distributed meta storage
+ * in order to store tracing configuration.
+ */
+public class GridTracingConfigurationManager implements TracingConfigurationManager {
```
glukos, line 38: Suffix "Manager" makes this class look like lifecycle-aware manager component, which is not true.
What about jut TracingConfiguration (for interface) and TracingConfigurationImpl (for this class)?
Another option (to clarify that this class is not part of static configuration): DistributedTracingConfiguration  and DistributedTracingConfigurationImpl.
See also: DistributedBaselineConfiguration

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/TracingConfigurationParameters.java
```diff
@@ -0,0 +1,143 @@
...
+    public static final double SAMPLING_RATE_NEVER = 0d;
+
+    /** Max valid sampling rate with special meaning that span will be always created. */
+    public static final double SAMPLING_RATE_ALWAYS = 1d;
+
+    /**
+     * Number between 0 and 1 that more or less reflects the probability of sampling specific trace.
+     * 0 and 1 have special meaning here, 0 means never 1 means always. Default value is 0 (never).
+     */
+    private final double samplingRate;
```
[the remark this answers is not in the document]
sanpwc: Yep, that will be few hundreds of such samplingRates at most.

modules/opencensus/src/test/java/org/apache/ignite/TracingConfigurationValidationTest.java
sanpwc: Xopowo

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/configuration/GridTracingConfigurationManager.java
sanpwc: Was discussed in separate thread. As a result it'll be TracingConfigurationManager.

PR #1139, opened 2020-05-12
GG-28536 Added new format of snapshot file

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/MarkerPageIO.java
```diff
@@ -0,0 +1,65 @@
...
+    /**
+     * @param type Page type.
+     * @param ver  Page format version.
+     */
+    protected MarkerPageIO(int type, int ver) {
+        super(type, ver);
+    }
+
+    /** */
+    public int markerType(long pageAddr) {
```
sk0x50, line 52: Please add Javadoc here and below.

PR #1140, opened 2020-05-12
GG-29192 Fully working version of GridGain Control Center agent implemented as plugin.

akuznetsov-gridgain:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -314,8 +314,9 @@ public static boolean allNodesSupport(
                 continue;
 
             // Add only when Control Center is enabled.
-            if (MANAGEMENT_CONSOLE == value && !IgniteComponentType.MANAGEMENT_CONSOLE.inClassPath())
-                continue;
+            // TODO GG-29192 How to implement feature check in case of plugin?
```
nva, line 317: For what propose do you need this feature?

PR #1142, opened 2020-05-13
GG-29108 Synchronization problems when different classloaders are use…

tkalkirill:
…d for deployment of same class

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
@@ -183,40 +183,43 @@ class GridDeploymentLocalStore extends GridDeploymentStoreAdapter {
                 return null;
             }
 
-            try {
-                // Check that class can be loaded.
-                String clsName = meta.className();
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -183,40 +183,43 @@
                 return null;
             }
 
-            try {
-                // Check that class can be loaded.
-                String clsName = meta.className();
+            while (true) {
```
mshonichev, line 186: Is it safe to loop forever like this? Maybe add some limit counter to the loop?

mshonichev (commented):
Should not we replace assertion to check at line no 162 too?

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
vldpyatkov: I agree to add a limit, but when we can't find deployment here node will only have to stop by FH.

PR #1150, opened 2020-05-14
GG-28399 [IGNITE-12399' Java thin client: add cache expiry policies

isapego:


modules/core/src/main/java/org/apache/ignite/client/ClientCache.java
```diff
@@ -335,12 +337,19 @@
      */
     public <K1, V1> ClientCache<K1, V1> withKeepBinary();
 
+    /**
+     * Returns cache with the specified expired policy set. This policy will be used for each operation invoked on
+     * the returned cache.
+     *
+     * @return Cache instance with the specified expiry policy set.
+     */
+    public <K1, V1> ClientCache<K1, V1> withExpirePolicy(ExpiryPolicy expirePlc);
```
ashapkin, line 346: I believe that this method should be consistent with a regular cache API and be named as "withExpir**y**policy". But since we are making a port from Ignite, let's keep the single code base for now.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
```diff
@@ -338,15 +343,32 @@ void cacheConfiguration(ClientCacheConfiguration cfg, BinaryOutputStream out, Pr
...
+                        w.writeBoolean(false);
+                    else {
+                        w.writeBoolean(true);
+                        w.writeLong(convertDuration(expiryPlc.getExpiryForCreation()));
+                        w.writeLong(convertDuration(expiryPlc.getExpiryForUpdate()));
+                        w.writeLong(convertDuration(expiryPlc.getExpiryForAccess()));
+                    }
+                });
+            } else if (cfg.getExpiryPolicy() != null) {
+                throw new ClientProtocolError(String.format("Expire policies have not supported by the server " +
```
ashapkin, line 368: "Expiry", but see comment above

PR #1153, opened 2020-05-15
GG-28799 Added support of atomic caches

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -6696,6 +6723,8 @@ private void versionCheck(@Nullable IgniteBiTuple<Object, Exception> invokeRes)
                             cctx.store().put(null, entry.key, val, entry.ver);
                     }
                     else {
+                        outOfOrderUpdate = ATOMIC_VER_COMPARATOR.compare(entry.ver, newVer) > 0;
```
glukos, line 6726: comparator is executed three times in this method, can we optimize it?

PR #1156, opened 2020-05-18
GG-29287

sanpwc:
Splitting tracing manager and tracing spi hierarchies.


modules/core/src/main/java/org/apache/ignite/internal/managers/tracing/GridTracingManager.java
```diff
@@ -179,8 +211,127 @@ private Span enrichWithLocalNodeParameters(@Nullable Span span) {
...
     /** {@inheritDoc} */
-    @Override public Span create(@NotNull SpanType spanType, @Nullable byte[] serializedSpan) {
-        return enrichWithLocalNodeParameters(getSpi().create(spanType, serializedSpan));
+    @Override public Span create(@NotNull SpanType spanType, @Nullable byte[] serializedParentSpan) {
+        // 1 byte: special flags;
+        // 1 bytes: spi type;
+        // 2 bytes: major protocol version;
+        // 2 bytes: minor protocol version;
+        // 4 bytes: spi specific serialized span length;
+        // n bytes: spi specific serializes span length;
```
glukos, line 220: spi specific serialize**d** span body ?

modules/core/src/main/java/org/apache/ignite/internal/managers/tracing/GridTracingManager.java
```diff
@@ -179,8 +211,127 @@ private Span enrichWithLocalNodeParameters(@Nullable Span span) {
...
+            // Fall through.
+            switch (minProtoVer) {
+                case 0 : {
+                    // Deserialize parent span type.
+                    parentSpanType = SpanType.fromIndex(
+                        bytesToInt(
+                            Arrays.copyOfRange(
+                                serializedParentSpan,
+                                8 + spiSpecificSpanSize,
+                                12 + spiSpecificSpanSize),
```
glukos, line 269: It would be a bit safer to create constants for previous items lengths and calculate prefix size with "+"
See: PageIO

modules/core/src/main/java/org/apache/ignite/internal/managers/tracing/GridTracingManager.java
```diff
@@ -179,8 +211,127 @@ private Span enrichWithLocalNodeParameters(@Nullable Span span) {
...
+                // create new span as child span for parent span, using parents span included scopes.
+
+                Set<Scope> mergedIncludedScopes = new HashSet<>(includedScopes);
+                mergedIncludedScopes.add(parentSpanType.scope());
+                mergedIncludedScopes.remove(spanType.scope());
+
+                span = new SpanImpl(
+                    getSpi().create(
+                        spanType.spanName(),
+                        Arrays.copyOfRange(serializedParentSpan,8,  8 + spiSpecificSpanSize)),
```
glukos, line 306: " , "

modules/core/src/main/java/org/apache/ignite/internal/managers/tracing/GridTracingManager.java
glukos: Applicable to all next code lines with hardcoded constants

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/Span.java
```diff
@@ -26,64 +26,60 @@
  * status, time events, attributes, links.
  * You can use tracing to debug errors and latency issues in your applications.
  */
-public interface Span {
+public interface Span extends SpiSpecificSpan {
```
glukos, line 29: Do we need inheritance here?
- We already delegate all calls to spiSpecificSpan 
- Looks like it would be more clear to distinguish that regular Span is not SPI specific (Ignite logic will work independent of SPI choice)

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/SpanManager.java
```diff
@@ -29,7 +29,7 @@
      *
      * @param spanType Type of span to create.
      */
-    public default Span create(@NotNull SpanType spanType) {
```
glukos, line 32: By the way, do we need two separate abstractions: Tracing and SpanManager?
As long as Tracing SPI implements neither of them, looks like we can keep only one.

sanpwc: I think yes. Tracing is general one that encapsulates both SpanManager and TracingConfiguratoin. We might prefer composition over inheritance here, but I'm not sure that it's necessary. Probably when will will add some mote stuff to Tracing we might move SpanManager inside Tracing it instead of inheriting it.

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/Span.java
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/managers/tracing/GridTracingManager.java
sanpwc: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/managers/tracing/GridTracingManager.java
sanpwc: Fixed.

PR #1180, opened 2020-05-22
GG-29165 Improvement of page replacement, reduce replaced pages.

akalash:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryImpl.java
```diff
@@ -2319,92 +2317,45 @@ private long removePageForReplacement(PageStoreWriter saveDirtyPage) throws Igni
 
                     boolean pinned = PageHeader.isAcquired(absPageAddr);
 
-                    boolean skip = ignored != null && ignored.contains(rndAddr);
-
-                    if (relRmvAddr == rndAddr || pinned || skip) {
+                    if (pinned || ignored != null && ignored.contains(rndAddr) ||
+                        !electedPages.add(new PageWithAttrHolder(absPageAddr, rndAddr, fullId))) {
```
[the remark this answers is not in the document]
zstan: good catch ! i did it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryImpl.java
```diff
@@ -2319,92 +2317,46 @@ private long removePageForReplacement(PageStoreWriter saveDirtyPage) throws Igni
 
                     boolean pinned = PageHeader.isAcquired(absPageAddr);
 
-                    boolean skip = ignored != null && ignored.contains(rndAddr);
-
-                    if (relRmvAddr == rndAddr || pinned || skip) {
+                    if (pinned || ignored != null && ignored.contains(rndAddr) ||
+                        !electedPages.add(new PageWithAttrHolder(absPageAddr, rndAddr, fullId)) ||
+                        fullId.pageId() == storeMgr.metaPageId(fullId.groupId())) {
```
akalash, line 2322: You I checks 'fullId.pageId() == storeMgr.metaPageId(fullId.groupId())' only after page was added to electedPages, which is not correct because we should add to electedPages only valid pages. So you should swap condition or you should remove this page from electedPages if 'fullId.pageId() == storeMgr.metaPageId(fullId.groupId())' is true.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryImplTest.java
```diff
@@ -678,6 +680,65 @@ private PageMemoryImpl createPageMemory(
...
+     *  Right comparator and further iteration need to be sorted by ts and flags from lower to higher:
+     *  [(ts = 100, dirty=false, meta=false),
+     *  (ts = 200, dirty=false, meta=false),
+     *  (ts = 80, dirty=true, meta=false),
+     *  (ts = 90, dirty=true, meta=false),
+     *  (ts = 50, dirty=false|true, meta=true),
+     *  (ts = 300, dirty=false|true, meta=true)]
+     */
+    @Test
+    public void testPageWithAttrComparator() {
```
akalash, line 694: Maybe it better to rewrite this test to more explicit form because honest to say it pretty hard to understand what happens here? When I hear about sorting I expected to see something like:
```
Set<Integer> checkedSet = new TreeSet<>();

checkedSet.add(2);
checkedSet.add(1);
checkedSet.add(3);

List<Integer> sorted = new ArrrayList<>(checkedSet);

assertEquals(1, sorted.get(0));
assertEquals(2, sorted.get(1));
assertEquals(3, sorted.get(2));
```
But of course, it's my personal point of view.

But anyway as I can understand there are 16 possible combinations of comparing(true, true - true, true | true, true - true, false | etc.). But I don't see all of that in your test(even if some of the combinations sorted equally I want to see it explicitly)

zstan: @akalash make sense ) i check this test after some time pass and found that it is a little bit incomprehensible for me too, i add some helpful code, may be it became more clear now ?
what about combination, i describe all possible flags in javadoc, i think we have less combinations here:
first of all we need to see only dirty=false, meta=false pages with sorted ts,
after dirty=true with sorted ts,
after meta=true with any dirty=true|false with sorted ts
thus sequence need to be like:
ts = 100, dirty = false , meta = false
ts = 200, dirty = false , meta = false
ts = 1000, dirty = false , meta = false
---
ts = 200, dirty = true , meta = false
ts = 300, dirty = true , meta = false
ts = 1000, dirty = true , meta = false
---
ts = 200, dirty = true|false , meta = true
ts = 300, dirty = true|false , meta = true
ts = 1000, dirty = true|false , meta = true
did i miss something ?

PR #1184, opened 2020-05-25
GG-26360 Remove NoLGPL edition

vveider:


modules/geospatial/README.txt
```diff
@@ -32,17 +32,3 @@ interested in):
     ...
 </project>
 
-
```
alamar, line 35: I don't think we should remove this clause.

modules/hibernate-4.2/README.txt
```diff
@@ -32,17 +32,3 @@ interested in):
     ...
 </project>
 
-
-LGPL dependencies
```
alamar, line 36: and this one

modules/hibernate-5.1/README.txt
```diff
@@ -32,17 +32,3 @@ interested in):
     ...
 </project>
 
-
```
alamar, line 35: and this one

modules/hibernate-5.3/README.txt
```diff
@@ -32,17 +32,3 @@ interested in):
     ...
 </project>
 
-
-LGPL dependencies
```
alamar, line 36: and this one

modules/kubernetes/config/Dockerfile
```diff
@@ -20,7 +20,7 @@ FROM java:8
 ENV IGNITE_VERSION 2.0.0-SNAPSHOT
 
 # Set IGNITE_HOME variable.
-ENV IGNITE_HOME /opt/ignite/apache-ignite-lgpl-${IGNITE_VERSION}-bin
```
alamar, line 23: let's change it to /opt/gridgiain/gridgain-community-${IGNITE_VERSION}, also below

modules/ml/README.txt
```diff
@@ -8,10 +8,4 @@ GridGain ML Grid module provides machine learning features, along with involved
 Run from project root:
 mvn clean package -DskipTests -pl modules/ml -am
 
-GridGain binary releases cannot include LGPL dependencies. If you would like to activate native BLAS optimizations
```
alamar, line 11: Here, I'm not sure.

modules/schedule/README.txt
```diff
@@ -31,17 +31,3 @@ interested in):
     ...
 </project>
 
-
```
alamar, line 34: And this

modules/kubernetes/config/Dockerfile
vveider: Agreed. Done.

modules/hibernate-4.2/README.txt
slukyano: On this and other LGPL-related comments.
The text being removed was never true. GridGain of all editions can include LGPL, and always included it in regular editions. Also, we never had any source releases. This text seems to be just a copy-paste from Apache Ignite which was never altered in the fork.

I suggest we proceed with removing LGPL clauses.

alamar:
OK, looks good given the input from @slukyano

PR #1203, opened 2020-05-27
gg-27070

vmalin:
pull request for review


vmalin (commented):
General comments:
1. Message with cluster name should be same for all commands and should be printed at the same place. For example right after client connection establishing.
2. Not all commands are covered in this branch. For example validate_indexes command wasn't changed.
3. If you make commits via IDEA you can set up optimizing imports before commit so you won't have to think about unused imports

modules/core/src/main/java/org/apache/ignite/internal/commandline/ClusterChangeTagCommand.java
```diff
@@ -66,7 +66,7 @@
             );
 
             if (res.success())
-                logger.info("Cluster tag updated successfully, old tag was: " + res.tag());
+                logger.info( getFullClusterInfo(client.state()) + " tag updated successfully, old tag was: " + res.tag());
```
vmalin, line 69: Excess space in info call.

modules/core/src/main/java/org/apache/ignite/internal/commandline/Command.java
```diff
@@ -51,6 +53,29 @@ public static GridClient startClient(GridClientConfiguration clientCfg) throws E
         return client;
     }
 
+    /**
+     * @return Cluster information to show user for.
+     */
+    default String getFullClusterInfo(GridClientClusterState clientCfg){
```
vmalin, line 59: This method doesn't actually return full cluster info. Method name should be changed

modules/core/src/main/java/org/apache/ignite/internal/commandline/Command.java
```diff
@@ -51,6 +53,29 @@ public static GridClient startClient(GridClientConfiguration clientCfg) throws E
...
+        String clusterInfo = "Cluster";
+        if (clusterName != null)
+            clusterInfo = "Cluster \"" + clusterName + "\"";
+        return clusterInfo;
+    }
+
+    /**
+     * @return Cluster information to show user for.
+     */
+    default String getClusterInfo(GridClientClusterState clientCfg){
```
vmalin, line 70: getClusterInfo -> getClusterName

modules/core/src/main/java/org/apache/ignite/internal/commandline/Command.java
```diff
@@ -51,6 +53,29 @@ public static GridClient startClient(GridClientConfiguration clientCfg) throws E
...
+        String clusterName = null;
+        try{
+            clusterName = clientCfg.clusterName();
+        } catch (GridClientException ignored){
```
vmalin, line 74: Try-catch block doesn't comply with our code style. Also there should be some more blank lines. Check out this link: https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines

modules/core/src/test/java/org/apache/ignite/util/GridCommandHandlerClusterByClassTest.java
```diff
@@ -1673,6 +1674,7 @@ public void testErrIllegalStateWithVerbose() {
      * without {@link CommonArgParser#CMD_VERBOSE} flag.
      */
     @Test
+    @WithSystemProperty(key = IGNITE_CLUSTER_NAME, value = "TEST_CLUSTER_NAME")
```
vmalin, line 1677: This particular test checks very specific case so I don't think it's a good idea to add extra checks here. Better to choose 3-5 most popular commands and add your checks to some common cases. Also best way to check control script output is to use testOut.toString().contains(%your_pattern%)

modules/core/src/main/java/org/apache/ignite/internal/commandline/cache/CacheViewer.java
```diff
@@ -416,7 +421,8 @@ private void cachesConfig(
      * @param infos Caches info.
      * @param cmd Command.
      */
-    private void printCacheInfos(Collection<CacheInfo> infos, VisorViewCacheCmd cmd, Logger logger) {
+    private void printCacheInfos(Collection<CacheInfo> infos, VisorViewCacheCmd cmd, Logger logger, String clusterName) {
+        logger.info(clusterName);
```
vmalin, line 425: blank line after logger.info

modules/core/src/main/java/org/apache/ignite/internal/commandline/Command.java
```diff
@@ -51,6 +53,29 @@ public static GridClient startClient(GridClientConfiguration clientCfg) throws E
         return client;
     }
 
+    /**
+     * @return Cluster information to show user for.
+     */
+    default String getFullClusterInfo(GridClientClusterState clientCfg){
+        String clusterName = getClusterInfo(clientCfg);
+        String clusterInfo = "Cluster";
+        if (clusterName != null)
```
vmalin, line 62: Better to use if-else logic here. If clusterName == null do A, else do B.

modules/core/src/main/java/org/apache/ignite/internal/commandline/ClusterChangeTagCommand.java
VladimirGoncharov-GG: Fixed

PR #1204, opened 2020-05-27
GG-28995

tkalkirill:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -151,6 +151,9 @@
     /** Distributed propagation of tx collisions dump interval. */
     DISTRIBUTED_TX_COLLISIONS_DUMP(37),
 
+    /** Support for compatibility of command-line utility for snapshots. */
+    SNAPSHOT_UTILITY_COMPATIBILITY_SUPPORT(40),
```
sk0x50, line 155: It looks like this feature was added only for testing purposes. Am I right?

tkalkirill: I think so, I'll try to do it without it.

PR #1211, opened 2020-05-28
GG-29480 WIP

tkalkirill:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -2082,6 +2082,9 @@ public UpdateResult executeUpdateOnDataNode(
 
         GridFutureAdapter<Void> rebuildCacheIdxFut = new GridFutureAdapter<>();
 
+        //to avoid possible data race
+        GridFutureAdapter<Void> outRebuildCacheIdxFut = new GridFutureAdapter<>();
+
         cctx.group().metrics().addIndexBuildCountPartitionsLeft(cctx.topology().localPartitions().size());
 
         rebuildCacheIdxFut.listen(fut -> {
```
ygerzhedovich, line 2090: could we use chain future here instead of explicit create new feature?

tkalkirill: At first glance, the same thing is done here, I think you can leave it as it is.
org.apache.ignite.internal.util.future.GridFutureAdapter#chain(org.apache.ignite.lang.IgniteClosure<? super org.apache.ignite.internal.IgniteInternalFuture<R>,T>)

PR #1217, opened 2020-06-03
GG-29610 - fixed the problem when UnknownHostException in TcpDiscover…

aealeksandrov:
…yVmIpFinder fails node start if one of the addresses can't be resolved to IP

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
@@ -219,10 +219,11 @@ public class TcpDiscoveryVmIpFinder extends TcpDiscoveryIpFinderAdapter {
             if (portStr.contains("..")) {
                 port1 = Integer.parseInt(portStr.substring(0, portStr.indexOf("..")));
                 port2 = Integer.parseInt(portStr.substring(portStr.indexOf("..") + 2, portStr.length()));
-            } else
+            }
+            else
                 port1 = port2 = Integer.parseInt(portStr);
 
-            if (port1 != port2 && port2 < port1 || port1 <= 0 || port2 <= 0)
+            if ((port1 != port2 && port2 < port1) || port1 <= 0 || port2 <= 0)
                 throw new IgniteSpiException(errMsg);
 
             try {
--- a/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
@@ -225,7 +225,18 @@ public class TcpDiscoveryVmIpFinder extends TcpDiscoveryIpFinderAdapter {
 
                     Collection<InetSocketAddress> res = new ArrayList<>();
 
-                    InetAddress[] inetAddresses = InetAddress.getAllByName(addrStr);
+                    InetAddress[] inetAddresses;
+
+                    try{
+                        inetAddresses = InetAddress.getAllByName(addrStr);
+                    }
+                    catch (UnknownHostException e) {
+                        //ignoring
+                        for (int i = port1; i <= port2; i++)
+                            res.add(new InetSocketAddress(addrStr, i));
+
+                        return res;
+                    }
 
                     for (InetAddress curAddr : inetAddresses) {
                         // Upper bound included.
@@ -235,7 +246,7 @@ public class TcpDiscoveryVmIpFinder extends TcpDiscoveryIpFinderAdapter {
 
                     return res;
                 }
-                catch (IllegalArgumentException | UnknownHostException e) {
+                catch (IllegalArgumentException e) {
                     throw new IgniteSpiException(errMsg, e);
                 }
             }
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
```diff
@@ -225,7 +225,18 @@ else if (ipStr.indexOf(':') >= 0)
 
                     Collection<InetSocketAddress> res = new ArrayList<>();
 
-                    InetAddress[] inetAddresses = InetAddress.getAllByName(addrStr);
+                    InetAddress[] inetAddresses;
+
+                    try{
```
alamar, line 230: Whitespace is desired

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
```diff
@@ -235,15 +246,28 @@ else if (ipStr.indexOf(':') >= 0)
...
+                catch (IllegalArgumentException e) {
                     throw new IgniteSpiException(errMsg, e);
                 }
             }
             else {
                 try {
                     int port = Integer.parseInt(portStr);
 
-                    return Collections.singleton(new InetSocketAddress(addrStr, port));
+                    Collection<InetSocketAddress> col = new LinkedHashSet<>();
```
alamar, line 257: I wonder if we really need this if-else here. Can we have same logic block for range and single port?

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
```diff
@@ -235,15 +246,28 @@ else if (ipStr.indexOf(':') >= 0)
...
+
+                        for (InetAddress addrs : inetAddresses)
+                            col.add(new InetSocketAddress(addrs, port));
+                    }
+                    catch (UnknownHostException ignored) {
+                        //save previous behavior on UnknownHostException
+                        col = Collections.singleton(new InetSocketAddress(addrStr, port));
+                    }
+
+                    return col;
```
alamar, line 270: What about multicast finder, btw? It has the same addresses input.

alamar (changes requested):

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
aealeksandrov: The current fix is about VmIpFinder. Probably we should fix multicast too but ut should be done in a separate ticket.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
aealeksandrov: fixed

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
aealeksandrov: The code was updated. Please check one more time.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
```diff
@@ -214,40 +214,43 @@ else if (ipStr.indexOf(':') >= 0)
...
-                    int port1 = Integer.parseInt(portStr.substring(0, portStr.indexOf("..")));
-                    int port2 = Integer.parseInt(portStr.substring(portStr.indexOf("..") + 2, portStr.length()));
+                port1 = Integer.parseInt(portStr.substring(0, portStr.indexOf("..")));
+                port2 = Integer.parseInt(portStr.substring(portStr.indexOf("..") + 2, portStr.length()));
+            } else
+                port1 = port2 = Integer.parseInt(portStr);
 
-                    if (port2 < port1 || port1 == port2 || port1 <= 0 || port2 <= 0)
-                        throw new IgniteSpiException(errMsg);
+            if (port1 != port2 && port2 < port1 || port1 <= 0 || port2 <= 0)
```
alamar, line 225: Can you put the first two into ()? if ((port1 != port2 && port2 < port1) || port1 <= 0 || port2 <= 0) for better readability

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
```diff
@@ -214,40 +214,43 @@ else if (ipStr.indexOf(':') >= 0)
...
             String portStr = tokens[1];
+            int port1, port2;
 
             if (portStr.contains("..")) {
-                try {
-                    int port1 = Integer.parseInt(portStr.substring(0, portStr.indexOf("..")));
-                    int port2 = Integer.parseInt(portStr.substring(portStr.indexOf("..") + 2, portStr.length()));
+                port1 = Integer.parseInt(portStr.substring(0, portStr.indexOf("..")));
+                port2 = Integer.parseInt(portStr.substring(portStr.indexOf("..") + 2, portStr.length()));
+            } else
```
alamar, line 222: Need line feed before else

alamar (changes requested):
Otherwise, LGTM

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
aealeksandrov: Fixed.

modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ipfinder/vm/TcpDiscoveryVmIpFinder.java
aealeksandrov: Fixed

PR #1235, opened 2020-06-11
GG-29779 Spark 2.4 module was added.

aealeksandrov:


modules/clients/src/test/bin/start-nodes.cmd
```diff
@@ -39,7 +39,7 @@ set MVN_EXEC=mvn
 
 if defined M2_HOME set MVN_EXEC=%M2_HOME%\bin\%MVN_EXEC%
 
-call %MVN_EXEC% -P+test,-scala,-examples,-release clean package -DskipTests -DskipClientDocs
+call %MVN_EXEC% -P+test,-scala,-spark-2.4,-examples,-release clean package -DskipTests -DskipClientDocs
```
alamar, line 42: So it won't build scala but would build spark 2.3? How should it work?

alamar (changes requested):

PR #1245, opened 2020-06-17
GG-29781 Get rid of release pinned page AssertionError in case of hug…

tkalkirill:
…e rebuild index workers are in progress.


modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaOperationWorker.java
```diff
@@ -177,8 +178,17 @@ public boolean cacheRegistered() {
      * Cancel operation.
      */
     @Override public void cancel() {
-        if (cancelToken.cancel())
+        if (cancelToken.cancel()) {
+            try {
+                fut.get();
+            }
+            catch (IgniteCheckedException e) {
+                if (log.isDebugEnabled())
```
vldpyatkov, line 186: I think when you use isDebugEnable() need printing via debug(...).
But here warning will be better.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/schema/SchemaOperationWorker.java
```diff
@@ -177,8 +178,17 @@ public boolean cacheRegistered() {
      * Cancel operation.
      */
     @Override public void cancel() {
-        if (cancelToken.cancel())
+        if (cancelToken.cancel()) {
+            try {
+                fut.get();
```
akalash, line 183: Do you have any guarantees that this future will finish eventually? I just afraid that this line can lead to the hung of the node.

PR #1247, opened 2020-06-17
GG-29772

akorensh:


akorensh:
not needed.

slukyano (changes requested):
See comments in the line comments. Also, maybe add some simple tests for this after all? See GridMBeansTest for example.

modules/indexing/src/main/java/org/apache/ignite/internal/mxbean/SqlQueryMXBeanImpl.java
```diff
@@ -73,19 +74,30 @@ public SqlQueryMXBeanImpl(GridKernalContext ctx) {
         h2idx.longRunningQueries().setResultSetSizeThresholdMultiplier(rsSizeThresholdMultiplier);
     }
 
+    /** {@inheritDoc} */
+    @Override public Long getSqlGlobalMemoryQuotaBytes() {
+        return U.parseBytes(h2idx.memoryManager().getGlobalQuota());
```
slukyano, line 79: We already have an integer value of the quota, right? So we probably don't need `parseBytes`.

May change the `memoryManager().getGlobalQuota()` to return `long` instead of String? It is internal and doesn't seem to be used anywhere else, so we can change it easily.

akorensh:
Agreed. Implemented as such.

PR #1256, opened 2020-06-22
GG-29869 Improving snapshot check for exchangeless snapshots

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/serializer/RecordV1Serializer.java
```diff
@@ -384,6 +384,9 @@ static WALRecord readWithCrc(
             try {
                 size = in0.io().size();
             }
+            catch (UnsupportedOperationException unsupported) {
```
sk0x50, line 387: Looks weird to me. Is it possible to refactor this catch block or even delete it?

denis-chudov: It's related to implementation of org.gridgain.grid.persistentstore.snapshot.file.FileSnapshotInputStream.InputStreamFileIO#size. We can't calculate size there. Any suggestions?

sk0x50: Well, let's change `InputStreamFileIO ` as follows:
`
private class InputStreamFileIO extends AbstractFileIO {
        /** {@inheritDoc} */
        @Override public long size() {
            throw new IOException("The size cannot be calculated", new UnsupportedOperationException());
        }
        ....
}
`

WDYT?

PR #1263, opened 2020-06-29
GG-29764 Removed default cache from REST APIs

RealZheka:


aealeksandrov (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/query/QueryCommandHandler.java
```diff
@@ -227,6 +227,13 @@ private static void removeQueryCursor(Long qryId, ConcurrentHashMap<Long, QueryC
                 );
         }
 
+        if (req.command() != FETCH_SQL_QUERY && req.command() != CLOSE_SQL_QUERY) {
```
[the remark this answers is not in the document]
RealZheka: They have different logic and don't have cache name as parameter.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cache/GridCacheCommandHandler.java
```diff
@@ -363,15 +392,16 @@ private static Object appendOrPrepend(Object origVal, Object appendVal,
...
-        final String cacheName = req0.cacheName() == null ? DFLT_CACHE_NAME : req0.cacheName();
-
         final Object key = req0.key();
 
         final Set<GridClientCacheFlag> cacheFlags = GridClientCacheFlag.parseCacheFlags(req0.cacheFlags());
 
         try {
             GridRestCommand cmd = req0.command();
 
+            if (req0.cacheName() == null && CACHE_NAME_REQUIRED_REQUESTS.contains(cmd))
```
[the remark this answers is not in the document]
RealZheka: For some requests absence of cache name is correct scenario, so we should check them separately. I was following the logic of GridCacheCommandHandler with its KEY_REQUIRED_REQUESTS. But you are right, it will be more safe to create array with requests that should be processed differently. In such case the default behavior will be to check cache name.

aealeksandrov (approved):

aealeksandrov:
LGTM

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cache/GridCacheCommandHandler.java
```diff
@@ -180,6 +180,12 @@
         CACHE_UPDATE_TLL
     );
 
+    /** Cache requests that can have null cache name {@code cacheName}. */
+    private static final EnumSet<GridRestCommand> CACHE_NAME_NOT_REQUIRED_REQUESTS = EnumSet.of(
+        CACHE_CLEAR,
```
slukyano, line 185: This looks dangerous actually. Imagine someone forgets to specify a cache name and removes all data in the grid accidentally. How about not allowing to call `clear` without a cache name in this same patch?

slukyano (commented):
See comment on CACHE_CLEAR.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cache/GridCacheCommandHandler.java
RealZheka: Yes, I also paid attention to this, but CACHE_CLEAR is not part of the public REST API (https://apacheignite.readme.io/docs/rest-api#api-reference). Was not sure if it was used by some inner tool or by some customer, so just kept it.

slukyano: It is not documented, yes, but a lot of things are not documented, unfortunately.

I did a git annotate, and it seems CACHE_CLEAR was added as a part of a Redis command support here https://issues.apache.org/jira/browse/IGNITE-5706, but it is invokable on its own too.

Then, I looked into the usages of CACHE_CLEAR in the Redis protocol implementation. Based on what I see in GridRedisFlushCommandHandler, Redis always specifies the caches explicitly. It leads me to thinking that it should be safe to disallow CACHE_CLEAR without a cache name.

RealZheka: Fixed that. CACHE_CLEAR remained in CACHE_NAME_NOT_REQUIRED_REQUESTS because it uses "values" field. "CacheName" is just ignored. By the way, corresponding test didn't not know that and was working only because it cleared all caches.

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/handlers/cache/GridCacheCommandHandler.java
```diff
@@ -563,18 +570,14 @@ private static Object appendOrPrepend(Object origVal, Object appendVal,
...
                     break;
                 }
 
                 case CACHE_CLEAR: {
-                    Map<Object, Object> map = req0.values();
-
-                    // HashSet wrapping for correct serialization
-                    Set<Object> cacheNames = map == null ?
-                        new HashSet<Object>(ctx.cache().publicCacheNames()) : new HashSet<>(map.keySet());
+                    Set<Object> cacheNames = getKeys(req0);
```
slukyano, line 580: It was values and know it's keys? Why?

RealZheka: Do you mean that now I check keys? GetKeys() function does exactly the same as code that I've removed (except getting of all public caches when values() are null).

RealZheka: Do you mean that now I check keys? GetKeys() function does exactly the same as code that I've removed (except getting of all public caches when values() are null)

PR #1276, opened 2020-07-01
GG-29956

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -4341,22 +4341,42 @@ protected void logUpdate(GridCacheOperation op, CacheObject val, GridCacheVersio
...
      *
-     * @param gapStart Update counter corresponds to out-of-order update.
+     * @param op Update operation.
+     * @param val Write value.
+     * @param writeVer Write version.
+     * @param expireTime Expire time.
+     * @param updCntr Update counter.
      */
-    protected void logOutOfOrderUpdate(Long gapStart) throws IgniteCheckedException {
+    protected void logOutOfOrderUpdate(
```
glukos, line 4353: We should check whether WAL is enabled when we call this method.

sk0x50: This check is already added below.

PR #1283, opened 2020-07-03
Gg 28742

ingvard:


modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -1842,19 +1861,23 @@ private boolean inverseTcpConnectionFeatureIsSupported(ClusterNode node) {
         int connIdx
     ) throws IgniteCheckedException {
         if (ctx.security().enabled() &&
-            IgniteFeatures.allNodesSupports(ctx, ctx.discovery().aliveServerNodes(), IGNITE_SECURITY_PROCESSOR)) {
-            UUID secSubjId = null;
+            !(msg instanceof GridDhtPartitionsAbstractMessage) && !(msg instanceof GridJobExecuteResponse)) {
```
glukos, line 1864: Why exactly these messages? Requires explanation.

ascherbakoff: This is a necessary minimum. I think any response can be exluded, bu we do not have a means to check if a message is a response. Exclusion of GridJobExecuteResponse is necessary to avoid NPEs on deactivated cluster.

PR #1284, opened 2020-07-06
GG-29905 Fixes

ingvard:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheVersionGenerationWithCacheStorageTest.java
```diff
@@ -210,7 +210,7 @@ private void checkGridCacheVersionsGenerationOrder(Consumer<IgniteEx> actions, S
                     latch.await();
                 }
                 catch (InterruptedException e) {
-                    e.printStackTrace();
+                    throw new RuntimeException(e);
```
glukos, line 213: Probably it works, but it's too hardcore to kill node with FH it's being stopped.
Thread.currentThread().interrupt(); would do the trick.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteTxCacheWriteSynchronizationModesMultithreadedTest.java
```diff
@@ -205,121 +341,111 @@ private void multithreaded(CacheWriteSynchronizationMode syncMode,
...
-                                break;
-                            }
+                        catch (TransactionOptimisticException ignored) {
+                            // Retry.
+                        }
+                        catch (CacheException | IgniteException ignored) {
+                            break;
                         }
                     }
                 });
```
glukos, line 451: Why we still keep section with cache destroy?
Hang on destroy was the reason of suite timeout previous time.
Let's get rid of destroy, nodes are stopped after test anyway.

PR #1286, opened 2020-07-06
GG-30055: Ab Initio related fixes to ODBC

isapego:


modules/platforms/cpp/odbc-test/src/utility_test.cpp
```diff
@@ -89,4 +89,58 @@ BOOST_AUTO_TEST_CASE(TestUtilityWriteReadString)
...
+{
+    CheckDecimalWriteRead("0");
+    CheckDecimalWriteRead("1");
+    CheckDecimalWriteRead("-1");
+    CheckDecimalWriteRead("42");
+    CheckDecimalWriteRead("-42");
+    CheckDecimalWriteRead("160");
+    CheckDecimalWriteRead("-160");
+    CheckDecimalWriteRead("34729864879625196");
+    CheckDecimalWriteRead("-34729864879625196");
```
ptupitsyn, line 143: Should we also test values with fractional parts?

ptupitsyn (approved):

modules/platforms/cpp/odbc-test/src/utility_test.cpp
isapego: Good point. Added.

PR #1287, opened 2020-07-07
GG-29619 Change eviction order to be compared by rebalance mode and b…

alievmirza:
…y partition size

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
@@ -1,0 +1,214 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ *
+ * Licensed under the GridGain Community Edition License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.gridgain.com/products/software/community-edition/gridgain-community-edition-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.cache.distributed.dht.topology;
+
+import java.util.ArrayList;
+import java.util.Arrays;
+import java.util.Collections;
+import java.util.List;
+import java.util.concurrent.PriorityBlockingQueue;
+import java.util.concurrent.atomic.AtomicBoolean;
+import java.util.concurrent.locks.ReentrantLock;
+import org.apache.ignite.IgniteCache;
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.IgniteSystemProperties;
+import org.apache.ignite.cache.CacheRebalanceMode;
+import org.apache.ignite.configuration.CacheConfiguration;
+import org.apache.ignite.configuration.DataRegionConfiguration;
+import org.apache.ignite.configuration.DataStorageConfiguration;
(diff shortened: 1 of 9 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
```diff
@@ -0,0 +1,199 @@
...
+        stopAllGrids();
+
+        cleanPersistenceDir();
+    }
+
+    /**
+     * Tests that {@link CacheRebalanceMode#SYNC} caches are evicted at first.
+     */
+    @Test
+    @WithSystemProperty(key = IgniteSystemProperties.IGNITE_EVICTION_PERMITS, value = "1")
```
glukos, line 89: Let's also disable historical rebalancing for this scenario. In case we change threshold, this start will start to fail.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
```diff
@@ -0,0 +1,199 @@
...
+                    return (T)(new TestBucketQueue(1, latch));
+
+                return instance;
+            }
+        });
+
+        startGrid(0, resolver);
+
+        //We start not from 0 because the first partition that comes to queue is from async cache.
+        for (int i = 1; i < 101; i++)
```
glukos, line 142: This check looks too strong. Is there a guarantee that first partition will be from ASYNC cache?
By the way, why first partition should be from ASYNC cache, if we wait for 5 seconds before first poll?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
```diff
@@ -0,0 +1,199 @@
...
+        TestBucketQueue(int buckets, CountDownLatch latch) {
+            super(buckets);
+
+            this.latch = latch;
+        }
+
+        /** {@inheritDoc} */
+        @Override PartitionsEvictManager.PartitionEvictionTask poll(int bucket) {
+            try {
+                if (!latch.await(5, TimeUnit.SECONDS))
```
glukos, line 169: Looks a bit fragile. 
Can we wait until all partitions from sys-cache and default cache are scheduled for eviction instead? We can check that size of the queue is equal to total number of partitions.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
```diff
@@ -50,6 +51,9 @@
  * Tests that {@link CacheRebalanceMode#SYNC} caches are evicted at first.
  */
 public class PartitionEvictionOrderTest extends GridCommonAbstractTest {
+    /** Condition check. */
+    boolean condCheck;
```
glukos, line 55: Are you sure that there's happen-before between
condCheck = GridTestUtils.waitForCondition(() -> queue.size() == expEvictQueueSize, 100_000); **(system thread)**
and 
assertTrue(condCheck); **(test-runner thread)**
?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionEvictionOrderTest.java
```diff
@@ -131,6 +135,8 @@ public void testSyncCachesEvictedAtFirst() throws Exception {
 
         ReentrantLock lock = new ReentrantLock();
 
+        CopyOnWriteArrayList<Integer> evictionOrder = new CopyOnWriteArrayList<>();
```
glukos, line 138: Why CopyOnWriteArrayList ?
Complexity of its fill is O(N^2), where N is number of partitions. Collections.synchronizedList(new ArrayList<>()) would just O(N).

PR #1288, opened 2020-07-07
GG-21923 Fix nodes not joining baseline topology after enabling auto adjust

SammyVimes:


modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/baseline/autoadjust/ChangeTopologyWatcher.java
```diff
@@ -99,17 +99,26 @@ public ChangeTopologyWatcher(GridKernalContext ctx) {
...
+        triggerBaselineUpdate(discoEvt.topologyVersion());
+    }
 
-            if (isLocalNodeCoordinator(discoveryMgr)) {
-                exchangeManager.affinityReadyFuture(new AffinityTopologyVersion(discoEvt.topologyVersion()))
+    /**
+     * Schedule update of the baseline topology
+     * @param topologyVersion version of topology
+     */
+    public synchronized void triggerBaselineUpdate(long topologyVersion) {
```
akalash, line 109: I believe it is better not to add second method here. You can rename onEvent -> triggerBaselineUpdate and class ChangeTopologyWatcher to something like BaselineTopologyUpdater.
And then registration of events(EVT_NODE_FAILED, EVT_NODE_LEFT, EVT_NODE_JOINED) should be modified according to using triggerBaselineUpdate(long topologyVersion)

modules/core/src/test/java/org/apache/ignite/internal/processors/cluster/BaselineAutoAdjustTest.java
```diff
@@ -585,6 +586,39 @@ public void shouldJoinFailedBecauseCoordinatorIsInMemoryNodeAndEnabledAutoAdjust
...
+
+        ignite0.cluster().baselineAutoAdjustTimeout(100);
+
+        ignite0.cluster().baselineAutoAdjustEnabled(false);
+
+        ignite0.cluster().active(true);
+
+        startGrid(1);
+
+        assertTrue(GridTestUtils.waitForCondition(() -> {
```
akalash, line 611: You can add awaitPartitionMapExchange() after startGrid(1); then you can check ignite0.cluster().nodes().size() without waitForCondition. Also you must check  that currentBaselineTopology().size() equal to 1 before auto-adjust enabled.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/baseline/autoadjust/BaselineTopologyUpdater.java
```diff
@@ -152,6 +150,9 @@ public BaselineAutoAdjustStatus getStatus() {
             if (lastBaselineData.isAdjusted())
                 return BaselineAutoAdjustStatus.notScheduled();
 
+            if (baselineAutoAdjustScheduler.isExecutionExpired(lastBaselineData))
```
akalash, line 153: Do you add separate 'if' consciously or by mistake? ('if' which is above has the same return value)

PR #1293, opened 2020-07-08
GG-29779 Spark 2.4 module was added.

aealeksandrov:


examples-ml/pom-standalone.xml
```diff
@@ -163,6 +157,60 @@
         </dependency>
     </dependencies>
 
+    <profiles>
+        <profile>
+            <id>scala</id>
```
alamar, line 162: Why spark profile is called `scala` but spark 2.4 profile is called `spark-2.4'? I assume scala should include both.

pom.xml
```diff
@@ -616,6 +617,19 @@
             </modules>
         </profile>
 
+        <profile>
+            <id>spark-2.4</id>
```
alamar, line 621: Again, why do we need two different profiles here?

alamar (changes requested):

examples-ml/pom-standalone.xml
aealeksandrov: According to naming, it's not my desition. I took it from Ignite. As the next change, I will move spark-2.3 to separate modules. It was discussed with Stan Lukyanov (as you remember it was done in first variant of PR but it also means the changes that differ from Ignite codebase).

Scala can't use both of them because this module will be used for example. Examples at the moment can use only one version of the spark. If I added them both to scala then they will not be able to build.  Otherwise, I can change this behavior but it also will mean the changes that differ from Ignite codebase). It should be done in Ignite first.

pom.xml
aealeksandrov: I answered in the previous question.  These profiles used for examples. Spark 2.3 and 2.4 aren't compatible and should be separated.

alamar (approved):
LGTM

PR #1294, opened 2020-07-09
GG-30033

ascherbakoff:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtCacheAdapter.java
```diff
@@ -112,76 +105,29 @@
...
-            assert !loc.id().equals(n.id());
-
-            for (GridDhtForceKeysFuture<?, ?> f : forceKeyFuts.values())
-                f.onDiscoveryEvent(e);
-        }
-    };
-
     /**
-     * Empty constructor required for {@link Externalizable}.
+     * TODO GG-30033 remove this method.
```
gvvinblade, line 109: Should we remove it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtCacheAdapter.java
```diff
@@ -112,76 +105,29 @@
...
      */
-    protected final void processForceKeyResponse(ClusterNode node, GridDhtForceKeysResponse msg) {
-        GridDhtForceKeysFuture<?, ?> f = forceKeyFuts.get(msg.futureId());
-
-        if (f != null)
-            f.onResult(msg);
-        else if (log.isDebugEnabled())
-            log.debug("Receive force key response for unknown future (is it duplicate?) [nodeId=" + node.id() +
-                ", res=" + msg + ']');
+    protected GridDhtCacheAdapter() {
```
gvvinblade, line 129: Let's create a ticket to unimplement `Externalizable`, from my point of view it's impossible to serialize cache adapter and I can't imagine a case when we need it.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/GridCacheAbstractFullApiSelfTest.java
```diff
@@ -191,7 +191,7 @@ public void beforeGridCacheAbstractFullApiSelfTest() {
 
     /** {@inheritDoc} */
     @Override protected long getTestTimeout() {
-        return TEST_TIMEOUT;
+        return 10_000_000;
```
gvvinblade, line 194: let's return it back

gvvinblade (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtCacheAdapter.java
agoncharuk: https://issues.apache.org/jira/browse/IGNITE-13241

agoncharuk: We only need proxies to be externalizable, no need to have the cache adapter to use this interface. I put this into the ticket.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtGetFuture.java
```diff
@@ -340,7 +285,7 @@ private boolean map(KeyCacheObject key, boolean forceKeys) {
             if (parts == null || !F.contains(parts, part.id())) {
                 // By reserving, we make sure that partition won't be unloaded while processed.
                 if (part.reserve()) {
-                    if (forceKeys || (part.state() == OWNING || part.state() == LOST)) {
+                    if (part.state() == OWNING || part.state() == LOST) {
```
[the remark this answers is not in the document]
agoncharuk: @ascherbakoff We do reserve each partition only once (see the contains call above). I can replace this with IntHashMap, not sure if will make any difference though. What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtCacheAdapter.java
ascherbakoff: Ok, we can do this safely for Ignite but not for GG, or it will break compatibility.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtGetFuture.java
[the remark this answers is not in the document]
ascherbakoff: Missed a check. Still a check does linear scan on a array,  and we have resize on each append.
I think it would make sense to try.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/GridDhtGetFuture.java
```diff
@@ -259,77 +267,14 @@ public GridCacheVersion version() {
...
-                // to avoid useless remapping to other non-owning partitions. For non-mvcc caches the force key request
-                // is also useless in the such situations, so the same flow is here: allegedly we've made a force key
-                // request with no results and therefore forceKeys flag may be set to true here.
-                if (noOwners)
-                    forceKeys = true;
-            }
-
             GridDhtLocalPartition part = topVer.topologyVersion() > 0 ?
                 cache().topology().localPartition(keyPart, topVer, true) :
                 cache().topology().localPartition(keyPart);
```
ascherbakoff, line 280: I don't think we should create a partition on a get in any circumstances.
This should be simplified to GridDhtLocalPartition part = cache().topology().localPartition(keyPart);

PR #1296, opened 2020-07-09
GG-29802 [IGNITE-13125] SQL regressions framework: Random query generation

korlov42:


modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/SqlQueryRegressionsTest.java
```diff
@@ -0,0 +1,359 @@
...
+import org.apache.ignite.spi.discovery.tcp.ipfinder.TcpDiscoveryIpFinder;
+import org.apache.ignite.spi.discovery.tcp.ipfinder.vm.TcpDiscoveryVmIpFinder;
+import org.apache.ignite.testframework.junits.multijvm.IgniteProcessProxy;
+import org.junit.Test;
+
+/**
+ * Test for SQL queries regressions detection.
+ * It happens in the next way:
+ * 
+ * 1. Test starts two different Ignite versions: current version and the old one.
```
AMashenkov, line 66: "the old one" means some specific old version. Seems, you mean "one of previous versions".
So, how old version will be chosen? 
Would you please give few words on this.

korlov42: It should be passed as System property. See https://github.com/gridgain/gridgain/pull/1296/files#diff-5d8bc8f077dd7b97c909e7fef5b24756R238

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/SqlQueryRegressionsTest.java
```diff
@@ -0,0 +1,359 @@
...
+
+    /** */
+    private static final Schema SCHEMA = new Schema();
+
+    static {
+        MODEL_FACTORIES.stream().map(ModelFactory::queryEntity).map(Table::new).forEach(SCHEMA::addTable);
+    }
+
+    /** */
+    private boolean baseIsIgnite;
```
AMashenkov, line 139: {@true} if base is Ignite, {@false} if it is GG ?

korlov42: Yep. Will note this in javadoc

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/ModelUtil.java
```diff
@@ -0,0 +1,53 @@
...
+
+    /**
+     * Generate random alphabetical string.
+     *
+     * @param rnd Random object.
+     * @param minLen Minimum length of string.
+     * @param maxLen Maximal length of string.
+     * @return Random string object.
+     */
+    public static String randomString(Random rnd, int minLen, int maxLen) {
```
AMashenkov, line 34: Is it copy-pasted from GridTestUtil?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/City.java
```diff
@@ -0,0 +1,135 @@
...
+        return countryId;
+    }
+
+    /** */
+    public int population() {
+        return population;
+    }
+
+    /** */
+    public static class Factory implements ModelFactory {
```
AMashenkov, line 79: Factory classes looks identically, let's extract base\abstract class.

BaseFactory<T>
```suggestion
    public static class BaseFactory implements ModelFactory {
   
   protected Random rnd;
   protected final qryEntity;
   
    public BaseFactory(QueryEntity ent, String tableName, int count) {
    ...
    }
   
    public abstract T createRandom();
   
    public void init(int seed) {
            rnd = new Random(seed);
     }
   
     public QueryEntity queryEntity() {
            return qryEntity;
     }

   } 
```

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/ModelFactory.java
```diff
@@ -0,0 +1,39 @@
...
+
+/**
+ * Factory for model.
+ */
+public interface ModelFactory {
+    /** Inits factory with a random seed. */
+    public void init(int seed);
+
+    /** Creates random model objects. */
+    public Object createRandom();
```
AMashenkov, line 29: Why can't this class be a generic?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -0,0 +1,246 @@
...
+        return tbls.get(0);
+    }
+
+    /**
+     * Returns random column from provided table.
+     *
+     * @param tbl Table.
+     * @return Random column.
+     */
+    private ColumnRef rndColumn(TableRef tbl) {
```
AMashenkov, line 222: Actually this return only "numeric" column.
Let's rename the method to be more specific.

Also, I'd suggest to add "string" column support (or may be some another varlen type) as well. 
I'm ok to add new types support within separate ticket.

AMashenkov (approved):

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/ModelUtil.java
korlov42: Have no idea, it's kinda legacy code from Roman. Also it may be artefact of porting from AI codebase.
It's better to get rig of ModelUtils at all =)

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/City.java
korlov42: 👌

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/ModelFactory.java
korlov42: Think it can, so let's do it

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
korlov42: > Let's rename the method to be more specific

Ok 

> I'd suggest to add "string" column support

Actually this PR is like MVP for regression test. There are plans to improve generator not only with supporting different types, but with generating more meaningful queries, where conditions not totally random, but generated with regard to the value bounds and logic relations.

AMashenkov (approved):

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/SqlQueryRegressionsTest.java
AMashenkov: Would you please describe here how to run the test? or refer to sys properties? as it is still unclear how  "old version" is evaluated.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/ModelUtil.java
korlov42: Well I've got the answer: the AI has not such method, so if we are wanted to run benchmarks against AI, we have to provide it explicitly

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/model/ModelUtil.java
```diff
@@ -0,0 +1,73 @@
...
+ *     https://www.gridgain.com/products/software/community-edition/gridgain-community-edition-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+/*
```
AMashenkov, line 17: Double license

PR #1301, opened 2020-07-13
GG-28024 Improvement of index inline defaults

RealZheka:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -101,13 +111,26 @@ static int computeInlineSize(List<InlineIndexColumn> inlineIdxs, int cfgInlineSi
         int size = 0;
 
         for (InlineIndexColumn idxHelper : inlineIdxs) {
-            if (idxHelper.size() <= 0) {
-                size = propSize;
-                break;
+            // for variable types - default variable size, for other types - type's size + byte type
+            int sizeInc = idxHelper.size() <= 0 ? IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE : idxHelper.size() + 1;
```
ygerzhedovich, line 115: why we use LE or EQ to zero? I understand that is copy past from previous code, however why we should reserve DEFAULT_SIZE for zero size?
@korlov42 could we have zero type at all? may be we need additional assertion in AbstractInlineIndexColumn constructor?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -38,8 +41,15 @@
  * H2 tree index base.
  */
 public abstract class H2TreeIndexBase extends GridH2IndexBase {
+
     /** Default value for {@code IGNITE_MAX_INDEX_PAYLOAD_SIZE} */
-    static final int IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT = 10;
+    static final int IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT = 64;
```
ygerzhedovich, line 46: need to be carefully and understand that effective size will be less by 3. May be good to add such comment here.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -101,13 +111,26 @@ static int computeInlineSize(List<InlineIndexColumn> inlineIdxs, int cfgInlineSi
...
+
+            if (idxHelper instanceof StringInlineIndexColumn) {
+                String sql = idxHelper.columnSql();
+
+                if (sql != null) {
+                    Matcher m = STRING_WITH_LENGTH_SQL_PATTERN.matcher(sql);
+
+                    if (m.find())
+                        // if column has defined length we use it as default
+                        sizeInc = Integer.parseInt(String.valueOf(m.group().subSequence(1, m.group().length() - 1)));
```
RealZheka, line 125: Just realized that I lost my latest update. sizeInc should be equal to 
Integer.parseInt(String.valueOf(m.group().subSequence(1, m.group().length() - 1))) + 3
3 bytes should reserved by default. May be something similar should be done for string 115 for IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
RealZheka: Yes, just wrote the same in my comment. Does it worth to add extra bytes to provided IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT  in computeInlineSize() to be sure that it will be working with any value?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -101,13 +111,26 @@ static int computeInlineSize(List<InlineIndexColumn> inlineIdxs, int cfgInlineSi
         int size = 0;
 
         for (InlineIndexColumn idxHelper : inlineIdxs) {
-            if (idxHelper.size() <= 0) {
-                size = propSize;
-                break;
+            // for variable types - default variable size, for other types - type's size + byte type
+            int sizeInc = idxHelper.size() <= 0 ? IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE : idxHelper.size() + 1;
+
+            if (idxHelper instanceof StringInlineIndexColumn) {
```
ygerzhedovich, line 117: is it possible to add such optimization for Bytes too?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/SqlSystemViewsSelfTest.java
```diff
@@ -264,13 +264,13 @@ public void testIndexesView() throws Exception {
...
             {"1374144180", "SQL_PUBLIC_DFLT_AFF_CACHE", "1374144180", "SQL_PUBLIC_DFLT_AFF_CACHE", "PUBLIC", "DFLT_AFF_CACHE", "_key_PK_hash", "HASH", "\"ID1\" ASC, \"ID2\" ASC, \"ID1\" ASC", "false", "true", "null"},
 
-            {"1102275506", "SQL_PUBLIC_DFLT_CACHE", "1102275506", "SQL_PUBLIC_DFLT_CACHE", "PUBLIC", "DFLT_CACHE", "IDX_1", "BTREE", "\"ID2\" DESC, \"ID1\" ASC, \"MY_VAL\" DESC, \"ID1\" ASC, \"ID2\" ASC", "false", "false", "10"},
-            {"1102275506", "SQL_PUBLIC_DFLT_CACHE", "1102275506", "SQL_PUBLIC_DFLT_CACHE", "PUBLIC", "DFLT_CACHE", "IDX_3", "BTREE", "\"MY_VAL\" ASC, \"ID1\" ASC, \"ID2\" ASC, \"ID1\" ASC, \"ID2\" ASC", "false", "false", "10"},
+            {"1102275506", "SQL_PUBLIC_DFLT_CACHE", "1102275506", "SQL_PUBLIC_DFLT_CACHE", "PUBLIC", "DFLT_CACHE", "IDX_1", "BTREE", "\"ID2\" DESC, \"ID1\" ASC, \"MY_VAL\" DESC, \"ID1\" ASC, \"ID2\" ASC", "false", "false", "25"},
+            {"1102275506", "SQL_PUBLIC_DFLT_CACHE", "1102275506", "SQL_PUBLIC_DFLT_CACHE", "PUBLIC", "DFLT_CACHE", "IDX_3", "BTREE", "\"MY_VAL\" ASC, \"ID1\" ASC, \"ID2\" ASC, \"ID1\" ASC, \"ID2\" ASC", "false", "false", "25"},
```
korlov42, line 273: Could you explain please why we get 25 here? We have index with 4 int values and one string, so expected size is 4 * 5 + 1 * 10 = 30 bytes

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: why do you choose 64 for default?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -101,13 +111,26 @@ static int computeInlineSize(List<InlineIndexColumn> inlineIdxs, int cfgInlineSi
...
+                        sizeInc = Integer.parseInt(String.valueOf(m.group().subSequence(1, m.group().length() - 1)));
+                }
             }
 
-            // 1 byte type + size
-            size += idxHelper.size() + 1;
+            size += sizeInc;
+
+            // total index size is limited by the property
+            if (size > propSize)
```
korlov42, line 132: Why do you decide to limit max index size?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/database/H2ComputeInlineSizeTest.java
```diff
@@ -0,0 +1,158 @@
...
+     *
+     * Steps:
+     * 1) Call H2TreeIndexBase.computeInlineSize() function for 2 StringInlineIndexColumns, 1 BytesInlineIndexColumns
+     * and 1 LongInlineIndexColumn
+     * 2) Check that computed size is equal to 3 * default length for variable types + constant length of long column
+     */
+    @Test
+    public void testDefaultSizeForCompositeIndex() {
+        Column c1 = new Column("c1", new TypeInfo(Value.STRING, Integer.MAX_VALUE, 0, Integer.MAX_VALUE, null));
+        Column c2 = new Column("c2", new TypeInfo(Value.STRING, Integer.MAX_VALUE, 0, Integer.MAX_VALUE, null));
```
korlov42, line 126: let's make one string with defined length

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: Just found "Change of index inline defaults" thread, so question is not relevant any more

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: Found answer in "Change of index inline defaults" thread

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
RealZheka: For me support of 0-sized types seems incorrect. But I also don't see where some assertion can be added. Don't think that computeInlineSize(), which is being called for customer indexes, should check our inner implementation of sql type. Size is a hardcoded value and it is being set by ourselves for the each type. What if I will just add some description for the property that will contain expected values?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
RealZheka: Guys, what do you think about increasing of the actual index size by adding of 3 bytes to the IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT value? This property can be changed by user, so it will be more safe to add them by ourselves.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -101,13 +111,26 @@ static int computeInlineSize(List<InlineIndexColumn> inlineIdxs, int cfgInlineSi
         int size = 0;
 
         for (InlineIndexColumn idxHelper : inlineIdxs) {
-            if (idxHelper.size() <= 0) {
-                size = propSize;
-                break;
+            // for variable types - default variable size, for other types - type's size + byte type
```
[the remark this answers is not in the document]
RealZheka: Copy-pasted this from the previous description. Fixed.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/SqlSystemViewsSelfTest.java
RealZheka: Don't know, just fixed values for the tests. Will try to debug.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/database/H2ComputeInlineSizeTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+    /**
+     * Default variable length of the column's index
+     */
+    private static final int DEFAULT_VARIABLE_LENGTH = 10;
+
+    /**
+     * Default max length of the default index
+     */
+    private static final int DEFAULT_MAX_LENGTH = 64;
```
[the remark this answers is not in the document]
RealZheka: Maximum default allowed size for the index with variable type. Replaced it by H2TreeIndexBase.IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT to make this more clear

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
RealZheka: Also what do you think about increasing of IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE to 13 (3 bytes are taken for our info)?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -38,8 +41,15 @@
...
+
     /** Default value for {@code IGNITE_MAX_INDEX_PAYLOAD_SIZE} */
-    static final int IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT = 10;
+    static final int IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT = 64;
+
+    /** Default sql index size for types with variable length (such as String or byte[]). */
+    static final int IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE = 10;
+
+    /** SQL pattern for the String with defined length. */
+    static final Pattern STRING_WITH_LENGTH_SQL_PATTERN = Pattern.compile("\\([0-9]+\\)");
```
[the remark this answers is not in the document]
RealZheka: Checked suggested pattern, it returns VARCHAR(XX). With it I will need to cut type name + braces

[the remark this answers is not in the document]
korlov42: Try to get `matcher.group(1)`

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: Assertion should be added to constructor of AbstractInlineIndexColumn

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
korlov42: >  what do you think about increasing of the actual index size by adding of 3 bytes to the IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT value?

I think it should be left as is (64) because we have no information about types of the fields the index will consists of.

> Also what do you think about increasing of IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE to 13 

The same. Let it be 10 as it was discussed previously with the arch-team.

RealZheka: Ok, but note that IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT can be set by user. I think that we should add at least some assertion.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -38,8 +41,15 @@
  * H2 tree index base.
  */
 public abstract class H2TreeIndexBase extends GridH2IndexBase {
+
     /** Default value for {@code IGNITE_MAX_INDEX_PAYLOAD_SIZE} */
-    static final int IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT = 10;
+    static final int IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT = 64;
+
+    /** Default sql index size for types with variable length (such as String or byte[]). */
+    static final int IGNITE_VARIABLE_TYPE_DEFAULT_INDEX_SIZE = 10;
```
korlov42, line 49: By the way, is there a way to tune this? IMO, it worth to provide a way to adjust this param somehow.

RealZheka: @ygerzhedovich Do you agree with such update?

ygerzhedovich: yes, however it should be provide through decision approval procedures - I think it could be done in separate ticket in case it will require significant changes

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/SqlSystemViewsSelfTest.java
RealZheka: Why do you think that we should have 4 int values in the index? There are only 2 int columns:

_CREATE TABLE PUBLIC.DFLT_CACHE (ID1 INT, ID2 INT, MY_VAL VARCHAR, PRIMARY KEY (ID1 DESC, ID2))_

User indexes are:

_CREATE INDEX IDX_1 ON PUBLIC.DFLT_CACHE(ID2 DESC, ID1, MY_VAL DESC)
CREATE INDEX IDX_3 ON PUBLIC.DFLT_CACHE(MY_VAL)_

Both of them will contain 4 columns: 2 int (ID1 and ID2), 1 String (MY_VAL), 1 Object hash (_KEY)
5 * 2 + 10 + 5 = 25

Previous index size was equal to 10 because default maximum limit was equal to 10.

RealZheka:
@ygerzhedovich, @korlov42 ready for re-review

RealZheka:
@korlov42 fixed according last comments

ygerzhedovich (approved):

korlov42 (approved):

PR #1304, opened 2020-07-13
GG-29910

glukos:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/IgniteCacheNearRestartRollbackSelfTest.java
```diff
@@ -166,13 +174,15 @@ public void testRestarts() throws Exception {
...
+                    doRestarts.set(false);
+
                     throw e;
                 }
             }
-
-            fut.get();
         }
         finally {
+            fut.get();
```
glukos, line 184: What if fut.get() will throw an exception for some reason?
It's anyway better to stop grids in afterTest() to avoid any risks in test method code.

sergeyuttsel: got it

PR #1305, opened 2020-07-14
GG-30083 [IGNITE-13226] .NET: Fix ClientNotificationHandler leak in ClientSocket

ptupitsyn:
Fix race condition in `ClientSocket.AddNotificationHandler`: `SetHandler` call was causing `RemoveNotificationHandler` call before `ConcurrentDictionary.AddOrUpdate` has finished, so `AddOrUpdate` added the handler back, causing the leak.


isapego (approved):
Looks good.

PR #1307, opened 2020-07-14
GG-30084 [IGNITE-13214] .NET: Fix TransactionScope for read operations

ptupitsyn:
* Add `StartTxIfNeeded` to all cache operations (it was missing for Get, GetAll, TryGet)
* Add `OptimisticTransactionTest` with explicit and ambient tests
* Add exception mapping for `IgniteTxOptimisticCheckedException`
* Fix error handling in ambient transactions - close Ignite tx on Prepare failure


modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/OptimisticTransactionTest.cs
```diff
@@ -0,0 +1,133 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
isapego, line 2: It should be 2020, not 2019

isapego (approved):
Overall looks good. See my minor comment

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Cache/OptimisticTransactionTest.cs
ptupitsyn: Fixed

PR #1318, opened 2020-07-21
GG-30206: Enabled CPP Cluster API

isapego:


modules/platforms/cpp/binary/src/impl/binary/binary_object_impl.cpp
```diff
@@ -40,6 +40,20 @@ namespace ignite
             {
                 if (idRslvr)
                     this->idRslvr = idRslvr->Clone();
+
+                while (true)
```
ptupitsyn, line 44: Can you please explain this loop logic? Our BinaryObject format does not require looping like this, looks confusing.

modules/platforms/cpp/core/include/ignite/impl/compute/compute_impl.h
```diff
@@ -48,11 +48,13 @@ namespace ignite
                 {
                     enum Type
                     {
-                        AFFINITY = 1,
-
                         BROADCAST = 2,
 
                         UNICAST = 5,
+
+                        OP_AFFINITY_CALL = 13,
```
ptupitsyn, line 55: `OP_` prefix is not consistent

modules/platforms/cpp/core/src/impl/ignite_environment.cpp
```diff
@@ -659,18 +685,13 @@ namespace ignite
...
-            compute::ComputeJobHolder* job = job0.Get();
-
             SharedPointer<compute::ComputeTaskHolder> task0 =
                 StaticPointerCast<compute::ComputeTaskHolder>(registry.Get(taskHandle));
 
             compute::ComputeTaskHolder* task = task0.Get();
 
-            if (task && job)
-                return task->JobResultRemote(*job, reader);
+            if (task)
```
ptupitsyn, line 693: Combine with the condition below to make it `if-else`. The logic is confusing with the last error: is `Job is not registred for handle` even possible?

modules/platforms/cpp/core/src/impl/ignite_environment.cpp
```diff
@@ -789,6 +810,103 @@ namespace ignite
...
+
+            SharedPointer<compute::ComputeTaskHolder> task0 =
+                StaticPointerCast<compute::ComputeTaskHolder>(registry.Get(handle));
+
+            registry.Release(handle);
+
+            compute::ComputeTaskHolder* task = task0.Get();
+
+            task->JobResultError(err);
+            task->Reduce();
```
ptupitsyn, line 852: Should we reduce in case of error?

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
```diff
@@ -0,0 +1,160 @@
...
+        std::cout << ">>> Cluster compute example started." << std::endl;
+        std::cout << std::endl;
+
+        // Get binding instances and register our classes as a compute functions.
+        node1.GetBinding().RegisterComputeFunc<PrintMsg>();
+        node2.GetBinding().RegisterComputeFunc<PrintMsg>();
+        node3.GetBinding().RegisterComputeFunc<PrintMsg>();
+        client.GetBinding().RegisterComputeFunc<PrintMsg>();
+
+        // Create cluster groups splitted up by demo attribute value.
```
ptupitsyn, line 125: `splitted` -> `split`

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
```diff
@@ -0,0 +1,160 @@
...
+        ClusterGroup localGroup = client.GetCluster().AsClusterGroup();
+        ClusterGroup group1 = localGroup.ForAttribute("DemoAttribute", "Value0");
+        ClusterGroup group2 = localGroup.ForAttribute("DemoAttribute", "Value1");
+
+        // Making calls.
```
ptupitsyn, line 130: "Broadcast compute jobs"

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
```diff
@@ -0,0 +1,160 @@
...
+        client.GetCompute(group1).Broadcast(PrintMsg("DemoAttribute=Value0"));
+        client.GetCompute(group2).Broadcast(PrintMsg("DemoAttribute=Value1"));
+
+        // Waiting the compute jobs.
```
ptupitsyn, line 134: "Waiting for the compute jobs to complete".

By the way, this is not clear: C++ Compute API is synchronous, so we don't have to wait, as I understand.

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
```diff
@@ -0,0 +1,160 @@
...
+        std::cout << std::endl;
+        std::cout << ">>> Waiting all compute jobs done, before shutting down the cluster, press 'Enter to continue ..." << std::endl;
+        std::cout << std::endl;
+        std::cin.get();
+
+        // Stop client node first.
```
ptupitsyn, line 140: Why stop client first?

ptupitsyn (changes requested):

modules/platforms/cpp/examples/cluster-compute-example/src/cluster_compute_example.cpp
isapego: This is not completely true. We do have sync and async Compute API in C++. But in this case sync version is used, so you are right, this is not necessary, but it helps to see result of the example for a user. I think, only a message should be fixed.

modules/platforms/cpp/binary/src/impl/binary/binary_object_impl.cpp
```diff
@@ -36,10 +36,23 @@ namespace ignite
...
-                metaMgr(metaMgr)
+                metaMgr(metaMgr),
+                binary(false)
             {
                 if (idRslvr)
                     this->idRslvr = idRslvr->Clone();
+
+                int8_t hdr = BinaryUtils::ReadInt8(mem, this->start);
+
+                if (hdr == IGNITE_TYPE_BINARY)
```
ptupitsyn, line 47: What if this condition is false? Is it a valid case or should we throw an error?

ptupitsyn (changes requested):

modules/platforms/cpp/binary/src/impl/binary/binary_object_impl.cpp
isapego: It's OK, the class works on normal objects (hdr = 103) too. Maybe we should throw in other cases though.

PR #1322, opened 2020-07-24
GG-30304 Fix partition awarenes for thin clients.

SammyVimes:


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs
```diff
@@ -553,6 +553,12 @@ private object ReadDistributionMapResponse(IBinaryStream s)
 
                 // Count partitions to avoid reallocating array.
                 int maxPartNum = 0;
+
+                if (grp.PartitionMap == null)
+                {
+                    continue;
+                }
```
isapego, line 560: Not sure the code is following code style here. Also, will this change break backward compatibility for .NET?

SammyVimes: I'm not sure if our java code style applies here, so I just looked up other null check examples and copied it
image

PR #1338, opened 2020-08-03
GG-30394 False start of cacheGrpStates initialization(in CheckpointEntry)

ibessonov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
```diff
@@ -240,21 +242,38 @@ public CheckpointEntry lastCheckpointMarkingAsInapplicable(Integer grpId) {
      * Add last checkpoint to map of the earliest checkpoints.
      *
      * @param entry Checkpoint entry.
+     * @param cacheGrpStates
      */
-    private void addCpToEarliestCpMap(CheckpointEntry entry) {
+    private void addCpToEarliestCpMap(CheckpointEntry entry, Map<Integer, CacheState> cacheGrpStates) {
```
akalash, line 247: to avoid mistakes in further refactoring, I suggest splitting this method into two with the following signature:
* void addCpToEarliestCpMap(CheckpointEntry entry, Map<Integer, CacheState> cacheGrpStates)
* void addCpToEarliestCpMap(CheckpointEntry entry, Map<Integer, CheckpointEntry.GroupState> cacheGrpStates)

And extract following code to extra method:
```
if (!earliestCp.containsKey(grpPartKey))
         earliestCp.put(grpPartKey, entry);
```
because if tomorrow somebody changes the condition there are no guarantees that they do it in both places.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
```diff
@@ -174,9 +175,10 @@ public WALPointer firstCheckpointPointer() {
      * is not finished yet.
      *
      * @param entry Entry to add.
+     * @param cacheGrpStates
```
akalash, line 178: I believe it makes sense to add some comment why using cacheGrpStates from entry is a bad idea(to prevent further mistakes)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
ibessonov: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
ibessonov: I doubt that someone will reuse this method, and if they do, they can find current fix I guess.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointEntry.java
```diff
@@ -356,9 +356,12 @@ private void initIfNeeded(
 
                         grpStates = remap(stateRec);
                     }
-                    else
+                    else {
                         initEx = new IgniteCheckedException(
                             "Failed to find checkpoint record at the given WAL pointer: " + ptr);
+
+                        throw initEx;
```
akalash, line 363: As I understand you can skip the assignment of initEx here, because if you just throw an exception from here, initEx will be assignmented in the catch block.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
```diff
@@ -211,7 +213,7 @@ private void updateEarliestCpMap(CheckpointEntry entry) {
                     iter.remove();
             }
 
-            addCpToEarliestCpMap(entry);
+            addCpGroupStatesToEarliestCpMap(entry, entry.groupState(cctx));
```
akalash, line 216: You can use variable 'states' here because entry.groupState(cctx) was called in this method a little earlier.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointEntry.java
ibessonov: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointHistory.java
ibessonov: Done.

PR #1340, opened 2020-08-04
GG-30311 Fixed deployment remote filter of CQ on a client node

ibelyakov:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeploymentToClientTest.java
```diff
@@ -0,0 +1,155 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
sk0x50, line 2: 2020

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeploymentToClientTest.java
```diff
@@ -0,0 +1,155 @@
...
+    /** Cache name. */
+    private static final String CACHE_NAME = "test_cache";
+
+    /** {@inheritDoc} */
+    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);
+
+        cfg.setPeerClassLoadingEnabled(true);
+
+        cfg.setDiscoverySpi(new TcpDiscoverySpi().setIpFinder(ipFinder));
```
sk0x50, line 55: I don't think you need explicitly set shared TcpDiscoverySpi here, this step should be done at `GridAbstractTest#getConfiguration(String igniteInstanceName, IgniteTestResources rsrcs)`.
Please take a look at `GridAbstractTest.sharedStaticIpFinder` and its usages.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeploymentToClientTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        cfg.setCacheConfiguration(new CacheConfiguration<>(CACHE_NAME));
+
+        ((TcpCommunicationSpi)cfg.getCommunicationSpi()).setSharedMemoryPort(-1);
```
sk0x50, line 59: This property relates to communication via shared memory (aka SHMEM). You can just remove this line, I think.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryDeploymentToClientTest.java
```diff
@@ -0,0 +1,155 @@
...
+
+        cache.query(qry);
+
+        IgniteEx client2 = startClientGrid(2);
+
+        GridContinuousProcessor proc = client2.context().continuous();
+
+        assertInfosMap(U.field(proc, "locInfos"));
+        assertInfosMap(U.field(proc, "rmtInfos"));
+        assertEquals(0, ((Map)U.field(proc, "startFuts")).size());
```
sk0x50, line 101: It seems to me, using `Map<?, ?>` would be better (here and below).

modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/GridContinuousProcessor.java
```diff
@@ -660,13 +660,34 @@ private void registerHandlerOnJoin(UUID srcNodeId, UUID routineId, IgnitePredica
...
+                                hnd.p2pUnmarshal(srcNodeId, ctx);
+                            }
+                            catch (IgniteCheckedException | IgniteException e) {
+                                U.error(log, "Failed to unmarshal continuous routine handler [" +
+                                    "routineId=" + routineId +
+                                    ", srcNodeId=" + srcNodeId + ']', e);
+
+                                ctx.failure().process(new FailureContext(FailureType.CRITICAL_ERROR, e));
+
+                                unregisterHandler(routineId, hnd, false);
```
sk0x50, line 686: Why is FailureHandler called before unregistering the CQ handler?
I don't think, this is an issue, but could you please clarify?

PR #1358, opened 2020-08-11
GG-28760 Deprecate DataRegionMetrics#OffheapUsedSize, add TotalUsedSize

ibessonov:


modules/platforms/dotnet/Apache.Ignite.Core/IDataRegionMetrics.cs
```diff
@@ -43,6 +43,11 @@ public interface IDataRegionMetrics
         /// </summary>
         long TotalAllocatedSize { get; }
 
+        /// <summary>
+        /// Gets a total number of pages used for storing the data.
```
ptupitsyn, line 47: Please put here the same detailed comment that we have in `DataRegionMetrics.java`

ptupitsyn (changes requested):
Can we add tests for the new property? Ideally both in Java and C#.

modules/platforms/dotnet/Apache.Ignite.Core/IDataRegionMetrics.cs
ibessonov: Done

ptupitsyn (approved):
I've added a test to C# part, now we are good to go

PR #1360, opened 2020-08-12
GG-30465 control-center-agent not included in gridgain-zos libs/optional.

nva:


vveider (approved):
It is not very good that we have to duplicate parts of code, but currently the solution is acceptable.

alamar (approved):

PR #1362, opened 2020-08-12
GG-30403 RollingUpgrade enabled

ingvard:


modules/core/src/main/java/org/apache/ignite/internal/processors/ru/IgniteRollingUpgradeStatus.java
```diff
@@ -108,6 +107,7 @@ public IgniteRollingUpgradeStatus(
         out.writeBoolean(forcedModeEnabled);
         out.writeObject(initVer);
         out.writeObject(updateVer);
+        out.writeObject(new HashSet<>());
```
glukos, line 110: This will work, but it's better to override getProtocolVersion() and use it in the code like in other data transfer objects. It will raise less questions this way.

modules/core/src/test/resources/org.apache.ignite.util/GridCommandHandlerClusterByClassTest_help.output
```diff
@@ -63,10 +63,10 @@ This utility can do the following commands:
     control.(sh|bat) --diagnostic
 
   Enable rolling upgrade:
-    control.(sh|bat) --rolling-upgrade on
+    control.(sh|bat) --rolling-upgrade on [--force] [--yes]
```
sk0x50, line 66: Looks like it should be updated in accordance with new commands: start/finish/force
Please check `GridCommandHandlerClusterByClassTest.testHelp`

modules/core/src/main/java/org/apache/ignite/internal/visor/ru/VisorRollingUpgradeStatus.java
```diff
@@ -62,34 +58,30 @@ public VisorRollingUpgradeStatus() {
...
-        this.supportedFeatures = supportedFeatures;
     }
 
     /**
      * Creates a wrapper class that represents a value of {@link RollingUpgradeStatus}.
      *
      * @param status Rolling upgrade status to be copied.
      */
-    public VisorRollingUpgradeStatus(RollingUpgradeStatus status) {
+    public VisorRollingUpgradeStatus(IgniteRollingUpgradeStatus status) {
```
sk0x50, line 79: It seems that we can use RollingUpgradeStatus here. This will allow avoiding the following:
`status = new VisorRollingUpgradeStatus((IgniteRollingUpgradeStatus)changeModeRes.status());`
(see `VisorRollingUpgradeChangeModeResult`)
What do you think?

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -205,8 +206,8 @@ public static boolean nodeSupports(GridKernalContext ctx, ClusterNode clusterNod
         if (ctx != null) {
             RollingUpgradeStatus status = ctx.rollingUpgrade().getStatus();
 
-            if (status.enabled() && !status.forcedModeEnabled())
-                return status.supportedFeatures().contains(feature);
+            if (status.enabled() && !status.forcedModeEnabled() && status instanceof IgniteRollingUpgradeStatus)
```
sk0x50, line 209: Is this cast to `IgniteRollingUpgradeStatus` still actual?

modules/core/src/main/java/org/apache/ignite/internal/visor/ru/VisorRollingUpgradeChangeModeResult.java
```diff
@@ -54,8 +56,11 @@ public VisorRollingUpgradeChangeModeResult() {
      * @param changeModeRes Status of the operation.
      */
     public VisorRollingUpgradeChangeModeResult(RollingUpgradeModeChangeResult changeModeRes) {
+        if (!(changeModeRes.status() instanceof IgniteRollingUpgradeStatus))
```
sk0x50, line 59: Is this cast to `IgniteRollingUpgradeStatus` still actual?

PR #1368, opened 2020-08-13
GG-30489 Rename arguments in WAL reader and change default for printing sensitive data

vmalin:


modules/dev-utils/src/main/java/org/apache/ignite/development/utils/IgniteWalConverterArguments.java
```diff
@@ -113,33 +113,33 @@
...
      * @param skipCrc                       Skip CRC calculation/check flag.
      */
     public IgniteWalConverterArguments(File walDir, File walArchiveDir, int pageSize,
-        File binaryMetadataFileStoreDir, File marshallerMappingFileStoreDir, boolean keepBinary,
-        Set<WALRecord.RecordType> recordTypes, Long fromTime, Long toTime, String recordContainsText,
-        ProcessSensitiveData processSensitiveData,
+        File binaryMetadataDir, File marshallerMappingDir, boolean keepBinary,
+        Set<WALRecord.RecordType> recordTypes, Long fromTime, Long toTime, String hasText,
+        ProcessSensitiveData includeSensitive,
         boolean printStat, boolean skipCrc) {
```
vmalin, line 131: In this method list of arguments and method's body aren't separated. Although it's not connected to the ticket I think it's better to move '{' to separate line since you are changing this place.

vmalin (commented):
There is one minor style comment

PR #1371, opened 2020-08-14
GG-30524 unified interface to open files (from configuration, especially)

Berkof:
Add U.openFileInputStream and use it to find files

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
@@ -3601,7 +3601,6 @@ public abstract class IgniteUtils {
      * @throws IOException If stream could not be opened.
      */
     public static InputStream openFileInputStream(String filePath) throws IOException {
-
         File abs = new File(filePath);
 
         if (abs.exists())
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
@@ -3594,6 +3594,28 @@ public abstract class IgniteUtils {
     }
 
     /**
+     * Find file (by path or in resources) and open input stream to it.
+     *
+     * @param filePath Path to the file.
+     * @return Opened input stream.
+     * @throws IOException If stream could not be opened.
+     */
+    public static InputStream openFileInputStream(String filePath) throws IOException {
+
+        File abs = new File(filePath);
+
+        if (abs.exists())
+            return new FileInputStream(abs);
+
+        URL clsPthRes = IgniteUtils.class.getClassLoader().getResource(filePath);
+
+        if (clsPthRes != null)
+            return clsPthRes.openStream();
+
+        throw new FileNotFoundException("File " + filePath + " not found");
+    }
+
+    /**
      * Copies input byte stream to output byte stream.
      *
      * @param in Input byte stream.
--- a/modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
+++ b/modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
@@ -16,6 +16,7 @@
 
 package org.apache.ignite.ssl;
 
+import java.io.FileInputStream;
 import java.io.FileNotFoundException;
 import java.io.IOException;
 import java.io.InputStream;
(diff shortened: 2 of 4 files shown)


modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1056,6 +1058,52 @@ public void testDoInParallelBatch() {
...
+        // 1. by relative path
+        InputStream relative = U.openFileInputStream("org.apache.ignite.util/bigUtf.txt");
+
+        assertNotNull("Can't open input stream by relative path org.apache.ignite.util/bigUtf.txt", relative);
+
+        relative.close();
+
+        // 2. by absolute path
+        String absolutePath = U.getIgniteHome() + "/pom.xml";
+        InputStream absolute = U.openFileInputStream(absolutePath);
```
AMashenkov, line 1081: Let's use try-finally for streams.

Berkof: Switch to try with resources

AMashenkov (approved):

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
```diff
@@ -476,6 +477,7 @@ private void checkNullParameter(Object param, String name) throws SSLException {
      * @return Opened input stream.
      * @throws IOException If stream could not be opened.
      */
```
AMashenkov, line 479: '@deprecated Use {@link = U#openFileInputStream(String)} instead

AMashenkov, line 479:
```suggestion
     * @deprecated Use {@link = U#openFileInputStream(String)} instead.
     */
```

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1056,6 +1058,41 @@ public void testDoInParallelBatch() {
...
     }
 
+    /**
+     * Test open file:
+     * 1. by relative path
+     * 2. by absolute path
+     * 3. by classpath path
+     * 4. by non existing path
+     *
+     * @throws Throwable
```
AMashenkov, line 1068:
```suggestion
     * @throws Throwable If fails.
```

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -3593,6 +3593,28 @@ public static void onGridStop() throws InterruptedException {
...
 
+    /**
+     * Find file (by path or in resources) and open input stream to it.
+     *
+     * @param filePath Path to the file.
+     * @return Opened input stream.
+     * @throws IOException If stream could not be opened.
+     */
+    public static InputStream openFileInputStream(String filePath) throws IOException {
+
```
AMashenkov, line 3604:
```suggestion
```

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
Berkof: we can't offer to use internal class from public API

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
Berkof: DONE

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
Berkof: DONE

korlov42 (commented):
Could we provide a few more tests where we verify that keys will actually be loaded from both FS and classpath?

modules/core/src/main/java/org/apache/ignite/ssl/SslContextFactory.java
```diff
@@ -476,6 +477,7 @@ private void checkNullParameter(Object param, String name) throws SSLException {
...
+    @Deprecated
```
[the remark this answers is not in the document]
Berkof: Only because I suppose we should use the same way to load any configuration parts, not only for SslContextFactory. But now we haven't such mehanism.

PR #1372, opened 2020-08-17
GG-30384

akalash:


modules/control-utility/src/test/java/org/apache/ignite/util/GridCommandHandlerMetadataTest.java
```diff
@@ -105,6 +108,96 @@ public void testMetadataList() {
...
+
+        Collection<BinaryType> metadata = crd.context().cacheObjects().metadata();
+
+        assertEquals(EXIT_CODE_OK, execute("--meta", "list"));
+
+        assertEquals(metadata.toString(), 1, metadata.size());
+
+        assertContains(log, testOut.toString(), "typeName=" + TestValue.class.getTypeName());
+
+        grid(0).destroyCache(DEFAULT_CACHE_NAME);
```
akalash, line 186: I think it makes sense to add one more assert here that metadata is not empty yet after cache destroy(assertEquals(metadata.toString(), 1, metadata.size());)

sergey-chugunov-1985: Good point, I modified test as suggested and confirmed locally that it passes with the change.

PR #1386, opened 2020-08-24
GG-30549 Extend query generator with ability to generate single-table queries

korlov42:


modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -248,13 +320,51 @@ private ColumnRef rndNumericColumn(TableRef tbl) {
...
+         * @return Random value within interval [1..6].
+         */
+        public int roll() {
+            return 1 + rnd.nextInt(SIDES_COUNT);
+        }
+
+        /**
+         * Roll this dice twice.
+         *
+         * @return Random value within interval [1..12].
```
Berkof, line 351: 2..12

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -248,13 +320,51 @@ private ColumnRef rndNumericColumn(TableRef tbl) {
...
+         */
+        public int rollTwice() {
+            return rollNTimes(2);
+        }
+
+        /**
+         * Roll this dice {@code N} times.
+         *
+         * @param n Amount of time to roll dice. Should be greater than 0.
+         * @return Random value within interval [1..N*6].
```
Berkof, line 361: N..N*6

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -80,151 +92,211 @@ public RandomQuerySupplier(Schema schema, int seed) {
...
+        if (allowDublicates)
+            tbls.addAll(schema.tables()); // add all tables again to get duplicates
 
-            from = new InnerJoin(from, rightTbl,
-                new BiCondition(rndNumericColumn(rndScopeTable(rndQryCtx)), rndNumericColumn(rightTbl), Operator.EQUALS));
+        Collections.shuffle(tbls, rnd);
 
-            rndQryCtx.addScopeTable(rightTbl); // add a table to the scope AFTER we created the join condition
-        }
+        for (Table tbl : tbls.subList(0, tblCnt)) {
```
Berkof, line 260: can we require more than 6 tables (with duplicates) from schema with only 3 table?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -80,151 +92,211 @@ public RandomQuerySupplier(Schema schema, int seed) {
...
+            String alias = String.valueOf(tbl.name().toLowerCase().charAt(0)) + aliasIdGen.incrementAndGet();
 
-        return from;
+            rndQryCtx.addScopeTable(new TableRef(tbl, alias));
+        }
     }
 
     /**
-     * Returns random table which was used in FROM expression.
+     * Picks random item from provided collection that met specified condition.
```
Berkof, line 268: m**ee**t specified condition?

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -80,151 +92,211 @@ public RandomQuerySupplier(Schema schema, int seed) {
...
-        rndQryCtx.addScopeTable(leftTbl);
+    /**
+     * Picks N random table from the schema and adds them to the scope.
+     *
+     * @param rndQryCtx Random query context.
+     * @param tblCnt Tables count.
+     * @param allowDublicates Whether to allow dublicates or not.
+     */
+    private void pickRndTables(RandomisedQueryContext rndQryCtx, int tblCnt, boolean allowDublicates) {
+        List<Table> tbls = new ArrayList<>(schema.tables());
```
Berkof, line 253: Why we use local schema instead of tndQryCtx.schema? And why we need same schema in RandomisedQueryContext if we never use it?
It hard to track which schema is preferable in different situations and I hope we can remove one or comment the difference.

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
korlov42: yes, we can. So I've add handling for this case, thx

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
korlov42: schema should be part of RandomQuerySupplier only. So I've remove obsolete field inside RandomisedQueryContext

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
korlov42: yes, of course

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -2587,6 +2589,20 @@ public static long sleep_and_can_fail() {
...
 
+    /**
+     * @param lowerBound Lower bound.
+     * @param upperBound Upper bound.
+     */
+    public static <T extends Comparable<? super T>> Matcher<T> inRange(T lowerBound, T upperBound) {
+        return new CustomMatcher<T>("should be in range [" + lowerBound + ", " + upperBound + "]") {
+            @SuppressWarnings({"unchecked", "rawtypes"})
+            @Override public boolean matches(Object item) {
+                return lowerBound != null && upperBound != null && item instanceof Comparable
```
Berkof, line 2600: Lead to unexpected behaviour when called with nulls, for example assertThat(5, inRange(null, 10) should return True (for me)... Can we check arguments here or annotate its with \@NotNull  ?

Berkof (approved):
There are still a few magic constants, but in general it looks good to me

modules/compatibility/src/test/java/org/apache/ignite/compatibility/sql/randomsql/RandomQuerySupplier.java
```diff
@@ -80,151 +92,211 @@ public RandomQuerySupplier(Schema schema, int seed) {
...
-                tblCnt++;
-        }
+    /**
+     * Generates random SELECT expression from several tables.
+     *
+     * @param rndQryCtx Context of randomised query.
+     * @return Ast representing SELECT expression.
+     */
+    private Select rndMultiTableSelect(RandomisedQueryContext rndQryCtx) {
+        pickRndTables(rndQryCtx, 2 + rnd.nextInt(3), true);
```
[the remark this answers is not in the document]
korlov42: I added a comment to shed a light on these numbers

modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
korlov42: Added not null check. BTW we don't use `@NotNull` annotation in the project

PR #1396, opened 2020-08-26
GG-26550  Implement a common manager for managing runtime properties of the grid.

tledkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/metastorage/persistence/DmsLocalMetaStorageLock.java~!master
```diff
@@ -0,0 +1,26 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.metastorage.persistence;
+
+/** */
+interface DmsLocalMetaStorageLock {
+    /** Lock. */
+    public void lock();
```
Berkof, line 22: Why do we need access modifier here? https://docs.oracle.com/javase/specs/jls/se11/html/jls-9.html#jls-9.4 not recommend to use public in interface methods.

PR #1397, opened 2020-08-27
GG-28561

tkalkirill:


modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -502,10 +502,10 @@ public static void assertContains(@Nullable IgniteLogger log, String str, String
      *      and this message should be equal.
      * @return Thrown throwable.
      */
-    public static Throwable assertThrows(
+    public static <T> T assertThrows(
```
denis-chudov, line 505: Why have you removed wildcard? In this case we except that something that extends Throwable will be thrown.

PR #1401, opened 2020-08-31
GG-30665

tkalkirill:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -2006,47 +2006,43 @@ private void onChainFinished() {
...
+                ", bytesRcvd=" + U.humanReadableByteCount(fullBytes + histBytes) +
+                ", bandwidth=" + U.humanReadableByteCount((fullBytes + histBytes) / durationSec) + "/sec" +
+                ", histPartitions=" + histParts +
+                ", histEntries=" + histEntries +
+                ", histBytesRcvd=" + U.humanReadableByteCount(histBytes) +
+                ", fullPartitions=" + fullParts +
+                ", fullEntries=" + fullEntries +
+                ", fullBytesRcvd=" + U.humanReadableByteCount(fullBytes) +
+                ", topVer=" + topologyVersion() +
+                ", progress=" + (routines - remainingRoutines) + "/" + routines + "]");
```
ascherbakoff, line 2045: Are you 100% sure this code never throws any exception ? Let's keep the try - catch to avoid node shutdown on logging error. I suggest to do it for each complex logging block.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -2006,47 +2006,43 @@ private void onChainFinished() {
...
-            }
-            catch (Throwable t){
-                U.error(log, "Completed " + ((remainingRoutines == 0 ? "(final) " : "") +
-                    "rebalancing [grp=" + grp.cacheOrGroupName() +
-                    ", supplier=" + nodeId +
-                    ", topVer=" + topologyVersion() +
-                    ", progress=" + (routines - remainingRoutines) + "/" + routines + "]"), t);
-            }
+            // Node can left and by discovery we will not receive it.
+            Set<Integer> parts = nonNull(ctx.node(nodeId)) ? rebalancingParts.get(ctx.node(nodeId)) :
```
ascherbakoff, line 2010: This is incorrect. ctx.node(nodeId) can become null after the first null check and throw NPE on subsequent get(null).
I would suggest to make rebalancingParts = Map<UUID, Set<Integer>> instead and use 
rebalancingParts.getOrDefault(nodeId, Collections.emptySet())

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
ascherbakoff, line 2045: Are you 100% sure this code never throws any exception ? Let's keep the try - catch to avoid node shutdown on logging error. I suggest to do it for each complex logging block.

PR #1405, opened 2020-09-02
GG-30669 Fixed partition clearing while holding a group reservation.

sk0x50:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/EvictionWhilePartitionGroupIsReservedTest.java
```diff
@@ -0,0 +1,167 @@
...
+    private static final int PARTS = 1024;
+
+    /** {@inheritDoc} */
+    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);
+
+        cfg.setConsistentId(igniteInstanceName);
+
+        cfg.setCacheConfiguration(new CacheConfiguration(DEFAULT_CACHE_NAME).
+            setBackups(1).setAffinity(new RendezvousAffinityFunction(false, PARTS)).
```
sk0x50, line 55: It seems, `RendezvousAffinityFunction(false, 1024)` is the default, so there is no need to override.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/RentingPartitionIsOwnedDuringEvictionTest.java
```diff
@@ -137,25 +172,45 @@ private void testOwnedAfterEviction(boolean persistence, int backups) throws Exc
...
+            else if (reservation == 1)
+                evicting.reserve();
+
             IgniteEx joining = startGrid(backups + 1);
 
             if (persistence)
                 resetBaselineTopology();
 
-            GridDhtLocalPartition evicting = g0.cachex(DEFAULT_CACHE_NAME).context().topology().localPartition(p0);
+            doSleep(500);
```
sk0x50, line 196: Can we use `GridTestUtils.waitForCondition` instead of `doSleep`?
`
if (reservations > 0)
    boolean res = GridTestUtils.waitForCondition(() -> evicting.state() == OWNING)...);
`

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/EvictionWhilePartitionGroupIsReservedTest.java
ascherbakoff: I prefer to keep as is for readability,

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/RentingPartitionIsOwnedDuringEvictionTest.java
ascherbakoff: The sleep is here to wait a bit before checking for OWNING state. It is expected to not change, so waiting for condition is meaningless.

PR #1417, opened 2020-09-07
Gg 30693

vldpyatkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndex.java
```diff
@@ -152,6 +153,9 @@
     /** Query context registry. */
     private final QueryContextRegistry qryCtxRegistry;
 
+    /** If {code true} then this index is already marked as destroyed. */
+    private final AtomicBoolean destroyed = new AtomicBoolean(false);
```
vldpyatkov, line 157: False not need to write here.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndex.java
```diff
@@ -528,23 +532,15 @@ private static boolean isExpired(@NotNull H2Row row) {
...
-     * @param async Destroy asynchronously.
      */
-    private void destroy0(boolean rmvIdx, boolean async) {
+    @Override public void destroy(boolean rmvIdx) {
+        if (!markDestroyed())
+            return;
+
         try {
             if (cctx.affinityNode() && rmvIdx) {
                 assert cctx.shared().database().checkpointLockIsHeldByThread();
```
AMashenkov, line 546: Why we expects checkpoint is held by thread?

denis-chudov: Read lock moved inside of #destroy method, to delete all index segments from metatree under single read lock.

PR #1419, opened 2020-09-08
GG-30699 [IGNITE-13296] .NET: Fix transaction disposal and finalization

ptupitsyn:
- Implement TransactionRollbackOnlyProxy for GetLocalActiveTransactions to handle them differently than regular transactions
- Handle exceptions in transaction finalizers and Dispose methods


isapego (approved):
Looks good to me.

PR #1427, opened 2020-09-14
GG-29613 IgniteCache#isClosed() returns false on server node even if …

sk0x50:
…the cache had been closed before.

Signed-off-by: Ilya Kasnacheev <ilya.kasnacheev@gmail.com>
(cherry picked from commit 4b8ddd9bf4b7acf12dc05c1c657662b821f6e42f)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/IgniteCacheClusterReadOnlyModeSelfTest.java
```diff
@@ -515,11 +515,6 @@ public void testLocalClearAllDenied() {
     //@Test
     public void testCloseAllowed() {
```
sk0x50, line 516: Why is this test still disabled?

konstantins304: fixed

PR #1431, opened 2020-09-17
GG-27731 Add assertion message to assert in CachedDeploymentInfo cons…

JKonSir:
…tructor (#8231)

(cherry picked from commit f0d9b03276bf3fe080c234247f2e0ee3b26e79d3)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheDeploymentManager.java
```diff
@@ -947,7 +947,8 @@ private boolean isLocallyExcluded(String name) {
          */
         private CachedDeploymentInfo(UUID sndId, IgniteUuid ldrId, String userVer, DeploymentMode depMode,
             Map<UUID, IgniteUuid> participants) {
-            assert sndId.equals(ldrId.globalId()) || participants != null;
+            assert sndId.equals(ldrId.globalId()) || participants != null : "[senderId=" + sndId +
```
sk0x50, line 950: I would prefer to add a small message before detailed info as follows:
`Unable to create the deployment info [senderId....]`

PR #1432, opened 2020-09-18
GG-30359 Fix performance drop when there are many thin clients per server

gvvinblade:


modules/core/src/main/java/org/apache/ignite/internal/binary/streams/BinaryMemoryAllocator.java
```diff
@@ -16,41 +16,258 @@
...
+
+                if (data == null || size > data.length)
+                    data = new byte[size];
+
+                return data;
+            }
+
+            /** {@inheritDoc} */
+            @Override public byte[] reallocate(byte[] data, int size) {
+                byte[] newData = new byte[size];
```
isapego, line 106: Should not we check if the `size` fits `data` before allocating new memory?

modules/core/src/main/java/org/apache/ignite/internal/binary/streams/BinaryMemoryAllocator.java
```diff
@@ -16,41 +16,258 @@
...
-    /**
-     * Checks whether a thread-local array is acquired or not.
-     * The function is used by Unit tests.
-     *
-     * @return {@code true} if acquired {@code false} otherwise.
-     */
-    public boolean isAcquired() {
-        BinaryMemoryAllocatorChunk holder = holders.get();
+    /** */
+    private static class PooledAllocator extends BinaryMemoryAllocator {
```
isapego, line 146: Don't we have ready-to-use classes like this already?

isapego (approved):
Overall, looks good. I believe, we should implement the same client-side changes on other clients (C++, .NET at least)

modules/core/src/main/java/org/apache/ignite/internal/binary/streams/BinaryMemoryAllocator.java
gvvinblade: All allocations are checked by corresponding binary writer, the method won't be called if allocation isn't needed

modules/core/src/main/java/org/apache/ignite/internal/binary/streams/BinaryMemoryAllocator.java
gvvinblade: Nope

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
```diff
@@ -634,47 +635,80 @@ public ByteCountingDataInput(InputStream in) {
...
-                }
+            try {
+                while (offset < len) {
+                    int toRead;
 
-                if (bytesNum < 0)
-                    throw handleIOError(null);
+                    if (tryReadCnt == 0)
+                        toRead = len - offset;
+                    else if ((toRead = Math.min(in.available(), len - offset)) == 0) {
```
ptupitsyn, line 671: The condition with assignment is way too clever, can you please make it more readable?

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
```diff
@@ -634,47 +635,80 @@ public ByteCountingDataInput(InputStream in) {
...
+        public byte[] read(int len) throws ClientConnectionException {
+            byte[] bytes = new byte[len];
+
+            read(bytes, len, 0);
+
+            return bytes;
+        }
+
+        /** Read bytes from the input stream. */
+        public byte[] readNonBlocking(int len) {
```
ptupitsyn, line 648: Let's rename to `readSpinWait` to better reflect the logic. `nonBlocking` is misleading - the method can block the thread.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/TcpClientChannel.java
```diff
@@ -634,47 +635,80 @@ public ByteCountingDataInput(InputStream in) {
...
+            byte[] bytes = new byte[len];
+
+            read(bytes, len, Integer.MAX_VALUE);
```
ptupitsyn, line 651: I'm not sure this will scale well. It may be fine to spin a few times, but 2 million times is a waste of CPU.

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientMessage.java
```diff
@@ -0,0 +1,188 @@
...
+    /** {@inheritDoc} */
+    @Override public boolean writeTo(ByteBuffer buf, MessageWriter ignored) {
+        assert stream != null || data != null;
+
+        byte[] data = stream != null ? stream.array() : this.data;
+        int msgSize = stream != null ? stream.position() : data.length;
+
+        if (cnt < 0) {
+            for (; cnt < 0 && buf.hasRemaining(); cnt++)
+                buf.put((byte) ((msgSize >> (8 * (4 + cnt))) & 0xFF));
```
ptupitsyn, line 70: 1. When `cnt` is negative, it is always `-4`, can we use `putInt` here?
2. Does this handle endianness correctly?

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/ClientMessage.java
```diff
@@ -0,0 +1,188 @@
...
+        }
+
+        return false;
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean readFrom(ByteBuffer buf, MessageReader reader) {
+        if (cnt < 0) {
+            for (; cnt < 0 && buf.hasRemaining(); cnt++)
+                msgSize |= (buf.get() & 0xFF) << (8 * (4 + cnt));
```
ptupitsyn, line 112: Same as above, msgSize is 4 bytes, can we use `buf.getInt()`?

ptupitsyn (changes requested):

PR #1438, opened 2020-09-23
GG-30754 [IGNITE-13401] Java thin: Fix unmarshalling failure when cache configuration message or binary type message start with byte 103

ptupitsyn:


isapego (approved):
Looks good.

PR #1443, opened 2020-09-24
GG-30704 Show connection status in the result of management.sh --status

ingvard:


nva (commented):
All changes looks like debug and doesn't related with parent issue

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupDescriptor.java
```diff
@@ -268,6 +268,9 @@ public int groupId() {
      * @param cacheCfg Cache config.
      */
     public void config(CacheConfiguration cacheCfg) {
+        if (cacheCfg == null)
+            throw new NullPointerException("Cache configuration can't be null.");
```
MaxLosevskoy, line 272: Is `NPE` better than some `IllegalStateException` or such?

@ingvard @nva

PR #1456, opened 2020-09-29
GG-30804

sergeyuttsel:


modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
```diff
@@ -619,6 +619,8 @@ private void onDiscovery0(DiscoveryNotification notification) {
                     boolean incMinorTopVer;
 
                     if (customMsg instanceof ChangeGlobalStateMessage) {
+                        ctx.state().onStateMessage();
```
sergey-chugunov-1985, line 622: We don't need this new method in ClusterStateProcessor: it's okay to add this piece of logic right into existing onStateChangeMessage.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -626,13 +626,17 @@ private void onDiscoveryEvent(DiscoveryEvent evt, DiscoCache cache) {
                     else {
                         DiscoveryDataClusterState state = cctx.kernalContext().state().clusterState();
 
+                        assert state.transition() : state + " evt: " + evt + " msg: " + customMsg;
```
sergey-chugunov-1985, line 629: Why do we need this assertion? What does it protect us from?

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -176,6 +179,12 @@
     /** */
     private final ConcurrentMap<UUID, GridFutureAdapter<Void>> transitionFuts = new ConcurrentHashMap<>();
 
+    /**
+     * This latch is used to ensure that on client nodes the {@link ChangeGlobalStateMessage} is fully processed
+     * in the {@link GridCachePartitionExchangeManager} before the {@link ChangeGlobalStateFinishMessage} is processed.
+     */
+    private volatile CountDownLatch changeStateInProgress;
```
sergey-chugunov-1985, line 186: I reviewed how StateChange messages are created and sent, it seems possible to get two consequent StateChange messages before receiving StateChangeFinished for the first one.

So single latch isn't enough here, we should use ConcurrentHashMap and map latches by request IDs available both in StateChange and StateChangeFinished messages.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -622,6 +631,24 @@ private boolean isBaselineSatisfied(BaselineTopology blt, List<ClusterNode> srvN
     @Override public void onStateFinishMessage(ChangeGlobalStateFinishMessage msg) {
         DiscoveryDataClusterState discoClusterState = globalState;
 
+        if (ctx.clientNode() && !ctx.isDaemon() && changeStateInProgress != null) {
```
sergey-chugunov-1985, line 634: I don't like this style when we inline a very specific logic as plain blocks of code into existing methods.

Why not to create a new class called for instance ClientStateChangeSynchronizer and encapsulate this logic there (including aforementioned ConcurrentHashMap)?

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -662,6 +689,23 @@ private boolean isBaselineSatisfied(BaselineTopology blt, List<ClusterNode> srvN
...
     }
 
+    /** */
+    public void onStateMessage() {
+        if (ctx.clientNode() && !ctx.isDaemon())
+            changeStateInProgress = new CountDownLatch(1);
+    }
+
+    /** */
+    public void onChangeStateMessageWasProcessed() {
```
sergey-chugunov-1985, line 699: I suggest to rename this method to `onStateMessageProcessingFinished`.

sergey-chugunov-1985 (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
sergeyuttsel: This line was removed in https://github.com/gridgain/gridgain/commit/c932f681e53c1b5d5fe5c637d5e387ded8c6443c#diff-896382ae47330cb2c3aa1505290c99e18c93c338b9e584106d2e9e14e00a4b92 and a ticket GG-30804 was created. So now I revert it.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
sergeyuttsel: `onStateChangeMessage` invokes from other code. But we need to invoke `onStateMessage()` only for this case.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
sergeyuttsel: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
sergeyuttsel: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
sergeyuttsel: Done

PR #1461, opened 2020-09-29
GG-28222 CPP Thin: Use random address order

isapego:


modules/platforms/cpp/network/os/win/src/network/tcp_socket_client.cpp
```diff
@@ -28,6 +30,23 @@ namespace ignite
...
+        {
+            std::vector<addrinfo*> res;
+
+            for (addrinfo *it = addrsIn; it != NULL; it = it->ai_next)
+                res.push_back(it);
+
+            std::random_shuffle(res.begin(), res.end());
+
+            return res;
+        }
```
ptupitsyn, line 48: This looks exactly like the code above for Linux, can we deduplicate?

ptupitsyn (approved):
Looks good to me in general, please see one comment.

modules/platforms/cpp/network/os/win/src/network/tcp_socket_client.cpp
isapego: Well in theory, we can, using templates. The hard part here is that despite of the same name `addrinfo`s in Windows and Linux are two different classes, so we can not just make a simple function. I'll think if I can improve this part without over-complicating the code though.

isapego: Fixed. Check out new version.

ptupitsyn: Looks good, thanks!

PR #1468, opened 2020-10-01
GG-30952 Java thin: Fix ClientSizeCacheCreationDestructionTest expected exception type

ptupitsyn:


isapego (commented):
Changing exception type seems like change in visible behavior for user.

ptupitsyn:
@isapego you are right, but `ClientServerError` is internal and should not be thrown to the user. This behavior was changed in #1448 to be consistent across sync and async APIs.

PR #1469, opened 2020-10-01
GG-29771Extend information for IgniteDataIntegrityViolationException that can be thrown while reading WAL in FileInput.Crc32CheckingFileInput#close

vmalin:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/wal/WalCorruptionTest.java
```diff
@@ -0,0 +1,215 @@
...
+                IgniteDataIntegrityViolationException ex = cause(e, IgniteDataIntegrityViolationException.class);
+
+                if (ex != null) {
+                    assertTrue(ex.getMessage().matches("val: [-0-9]{1,11}, writtenCrc: [-0-9]{1,11}, " +
+                        "crcStartPos: " + secondRecordPos +
+                        ", crcEndPos: " + (secondRecordPos + REC_TYPE_SIZE + walPtrSerializedSize + INIT_NEW_PAGE_RECORD_PLAIN_SIZE)));
+
+                    return;
+                }
+                else {
```
vmalin, line 140: Looks like braces are redundant here

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/wal/WalCorruptionTest.java
```diff
@@ -0,0 +1,215 @@
...
+
+        /** {@inheritDoc} */
+        @Override protected IgniteCheckedException handleRecordException(
+            @NotNull Exception e,
+            @Nullable FileWALPointer ptr
+        ) {
+            return e instanceof IgniteCheckedException ? (IgniteCheckedException)e : null;
+        }
+    }
+}
```
vmalin, line 215: I don't know if our code style obliges every file to have newline in the end but all others do. I think it should be added

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/io/FileInput.java
```diff
@@ -102,7 +111,7 @@ public Crc32CheckingFileInput(FileInput delegate, boolean skipCheck) {
                 ensure(5);
 
                 throw new IgniteDataIntegrityViolationException(
-                    "val: " + val + " writtenCrc: " + writtenCrc
+                    "val: " + val + ", writtenCrc: " + writtenCrc + ", crcStartPos: " + calcStartPos + ", crcEndPos: " + calcEndPos
```
vmalin, line 114: In the ticket description three values were mentioned: WAL record size, CRC start offset, CRC end offset. But only two were added. Is this planned?

denis-chudov: It's not that simple right here, because abstraction level is lower than where WAL records are. The point is to see the range on which CRC is calculated - crcStartPos and crcStartPos should let us know.

PR #1482, opened 2020-10-07
GG-28316: Partial LEFT JOIN support from REPLICATED to PARTITIONED

slukyano:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/sql/SplitterUtils.java
```diff
@@ -164,6 +164,35 @@ public static boolean hasLeftJoin(GridSqlAst from) {
...
+    /**
+     * Checks whether the expression has an OUTER JOIN from replicated to partitioned.
+     *
+     * This is used to infer the `treatReplicatedAsPartitioned` flag
+     * to eventually pass it to {@link org.apache.ignite.spi.indexing.IndexingQueryFilterImpl}.
+     *
+     * @param from FROM expression.
+     * @return {@code true} if the expression has an OUTER JOIN from replicated to partitioned.
+     */
+    public static boolean hasOuterJoinReplicatedPartitioned(GridSqlAst from) {
```
korlov42, line 176: looks like this is not used anywhere

korlov42: or it just problem with my IDEA...

korlov42 (approved):

PR #1485, opened 2020-10-08
GG-29269 fix incorrect cache.size() with ttl enabled

alievmirza:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
```diff
@@ -2558,10 +2558,19 @@ private int cleanup0(GridCacheContext cctx, @Nullable List<MvccLinkAwareSearchRo
...
-            if (oldRow == null)
-                incrementSize(cctx.cacheId());
+            finishUpdate(cctx, newRow, oldRow, false);
+        }
 
-            KeyCacheObject key = newRow.key();
+        /**
+         * @param cctx Cache context.
+         * @param newRow New row.
+         * @param oldRow Old row if available.
```
AMashenkov, line 2567:
```suggestion
         * @param oldRow Old row if available.
         * @param oldRowExpired Old row expiration flag
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
```diff
@@ -3102,6 +3111,10 @@ public void restoreState(long size, long updCntr, @Nullable Map<Integer, Long> c
                 return oldRow;
             }
 
+            @Override public boolean oldRowExpiredFlag() {
```
AMashenkov, line 3114:
```suggestion
            /**
            * Flag that indicates if oldRow was expired during invoke.
            * @return {@code true} if old row was expired, {@code false} otherwise.
            */
            @Override public boolean oldRowExpiredFlag() {
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManager.java
```diff
@@ -578,6 +578,11 @@ public long cacheEntriesCount(int cacheId, boolean primary, boolean backup, Affi
          * @return Old row.
          */
         @Nullable public CacheDataRow oldRow();
+
+        /**
+         * Flag that indicates if oldRow was expired during invoke
+         */
```
AMashenkov, line 584:
```suggestion
         * Flag that indicates if oldRow was expired during invoke.
         * @return {@code true} if old row was expired, {@code false} otherwise.
         */
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/ttl/CacheSizeTtlTest.java
```diff
@@ -0,0 +1,132 @@
...
+ */
+public class CacheSizeTtlTest extends GridCommonAbstractTest {
+    /** Cache name. */
+    private static final String CACHE_NAME = "TestCache";
+
+    /** Entry expiry duration. */
+    private static final Duration ENTRY_EXPIRY_DURATION = new Duration(SECONDS, 1);
+
+    /** Wait condition duration. */
+    private static final long WAIT_COND_DURATION = 60_000L;
```
AMashenkov, line 53: Let's use getTestTimeout() instead.

AMashenkov (changes requested):

AMashenkov (approved):

PR #1496, opened 2020-10-13
GG-31113 [IGNITE-10837] Add ips of baseline nodes to control.sh --baseline

kgusakov:


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/AbstractCommand.java
```diff
@@ -0,0 +1,34 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
sk0x50, line 2: 2020

kgusakov:
Merged

PR #1514, opened 2020-10-23
GG-30867 Hide redundant Spring logs to the lower(DEBUG) level.

nva:


ingvard (approved):

dmekhanikov (changes requested):
Please try assembling GridGain with the fix and run it. The log messages are still there.

dmekhanikov:
It seems that it only works when JUL is configured. When you change it to Log4j2 for example, the messages are back again. They don't go away even if you apply the same configuration change in the Log4j2 config.

PR #1515, opened 2020-10-26
GG-31171: Fix ODBC Date metadata

isapego:


modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/odbc/OdbcColumnMeta.java
```diff
@@ -137,6 +138,18 @@ public void write(BinaryRawWriter writer) {
         }
     }
 
+    /**
+     * Get ODBC type ID for the type.
+     * @param dataType Data type class.
+     * @return Type ID.
+     */
+    private static byte getTypeId(Class<?> dataType) {
+        if (dataType.equals(java.sql.Date.class))
```
ptupitsyn, line 147: Can we move this check to `BinaryUtils.typeByClass`? It does not seem to be ODBC-specific.

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/odbc/OdbcColumnMeta.java
isapego: I'm not sure about that, this change can affect a lot of code and break a lot of things. Maybe we need an opinion of someone from core here. @agoncharuk, what do you think?

ptupitsyn (approved):

PR #1519, opened 2020-10-28
GG-31021 Fixed uninformative message in console about using deprecated active()

aries-3:


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2142,9 +2142,9 @@ private void ackStart(RuntimeMXBean rtBean) {
             log.info(str);
         }
 
-        if (!ctx.state().clusterState().active()) {
-            U.quietAndInfo(log, ">>> Ignite cluster is not active (limited functionality available). " +
-                "Use control.(sh|bat) script or IgniteCluster interface to activate.");
+        if (!ClusterState.active(ctx.state().clusterState().state())) {
+            U.quietAndInfo(log, ">>> Ignite cluster is in INACTIVE state (limited functionality available). " +
```
sanpwc, line 2146: Let's use `ClusterState.INACTIVE` here instead of hardcoded enum instance name.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2142,9 +2142,9 @@ private void ackStart(RuntimeMXBean rtBean) {
...
+                "Use control.(sh|bat) script or IgniteCluster.state(ClusterState.ACTIVE) to change the state.");
```
sanpwc, line 2147: There's also ACTIVE_READ_ONLY state. Don't you think that we should also mention it here?

PR #1520, opened 2020-10-28
GG-29756 Add REST tests for the new cluster state change API, additional --force param for ChangeState

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/client/GridClientClusterState.java
```diff
@@ -47,7 +47,18 @@
      * Changes cluster state to {@code newState}.
      *
      * @param newState New cluster state.
+     * @param forceDeactivation If {@code true}, cluster deactivation will be forced.
      * @throws GridClientException If the request to change the cluster state failed.
+     * @see ClusterState#INACTIVE
+     */
+    public void state(ClusterState newState, boolean forceDeactivation) throws GridClientException;
```
sk0x50, line 54: This API looks odd to me. Let's assume I call the method with the following parameters: `state(ClasterStae.ACTIVE, true);`
Does it mean that cluster will be forcibly deactivated in any way? Maybe `forceDeactivation` will be just ignored in this case. 
Long story short, it would be nice to update javadoc, at least it makes sense to mention that `forceDeactivation` is used only for cases when new cluster state is `INACTIVE`.

modules/core/src/main/java/org/apache/ignite/internal/client/impl/connection/GridClientNioTcpConnection.java
```diff
@@ -503,6 +504,7 @@ else if (now - lastPingSndTime > pingInterval && lastPingRcvTime != Long.MAX_VAL
                     sndFut.get();
                 }
                 catch (Exception e) {
+                    System.err.println("!!! ses.send3 " + e.toString());
```
sk0x50, line 507: Please remove debug output.

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -1455,6 +1496,7 @@ else if (joiningNodeBlt.id() < clusterBlt.id()) {
         U.log(
             log,
             "Sending " + prettyString(state) +
+                ", forceDeactivation=" + forceDeactivation +
```
sk0x50, line 1499: It seems to me, this line should be added after ` request from node [id=" + ctx.localNodeId() +`.
`Sending activation request from node[id=..., topVer=..., froceDeactivation=..., client=...]`

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/client/message/GridClientClusterStateRequestV2.java
```diff
@@ -0,0 +1,69 @@
...
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import org.apache.ignite.cluster.ClusterState;
+
+/**
+ * Enhanced version of {@link GridClientClusterStateRequest}.
+ * Introduced to support forced version of the change state command and keep backward compatibility
+ * with nodes of old version that may occur in cluster at the rolling updates.
+ */
+public class GridClientClusterStateRequestV2 extends GridClientClusterStateRequest {
```
sk0x50, line 29: Hmm, I don't think it is a good idea to extend deprecated class )))

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/protocols/tcp/GridTcpRestNioListener.java
```diff
@@ -496,6 +497,9 @@ else if (msg instanceof GridClientStateRequest) {
         else if (msg instanceof GridClientClusterStateRequest) {
             GridClientClusterStateRequest req = (GridClientClusterStateRequest)msg;
 
+            boolean forceDeactivation = !(msg instanceof GridClientClusterStateRequestV2) ||
```
sk0x50, line 500: this field can be moved to the corresponding branch.
```
if (req.isReqCurrentState()) {
    ...
}
else {
    ...
    boolean forceDeactivation = ...;
    restChangeReq.forceDeactivation(forceDeactivation);
}
```

modules/core/src/main/java/org/apache/ignite/internal/client/GridClientClusterState.java
zstan: but, existing java doc already highlight this? i hope : _If {@code true}, cluster deactivation will be forced._

zstan: i extend a bit existing java doc

sk0x50: I think that this "extended" javadoc should be applied to all places.
For example, `GridClientConnection#changeState(ClusterState state, UUID destNodeId, boolean forceDeactivation)`

modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/GridClusterStateProcessor.java
```diff
@@ -2124,6 +2174,16 @@ public static ClusterState clusterStateWithLessFeatures(ClusterState state1, Clu
         return S.toString(GridClusterStateProcessor.class, this);
     }
 
+    /**
+     * @return Lists in-memory user defined caches.
+     */
+    private List<String> listInMemoryUserCaches() {
+        return ctx.cache().cacheDescriptors().values().stream()
+            .filter(desc -> !isPersistentCache(desc.cacheConfiguration(), ctx.config().getDataStorageConfiguration())
+                && (!desc.cacheConfiguration().isWriteBehindEnabled() || !desc.cacheConfiguration().isReadThrough())
```
sk0x50, line 2183: Should it be `isWriteThrough()` instead fo `isWriteBehindEnabled()`?

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/client/message/GridClientClusterStateRequest.java
```diff
@@ -33,7 +34,7 @@
     private boolean reqCurrentState;
 
     /** New cluster state. */
-    private ClusterState state;
+    protected ClusterState state;
```
sk0x50, line 37: This change is not needed for now.

PR #1526, opened 2020-10-28
GG-31148 Thread Pools incomprehensible metric description

vmalin:


modules/core/src/main/java/org/apache/ignite/mxbean/ThreadPoolMXBean.java
```diff
@@ -143,7 +143,7 @@
      *
      * @return Class name of current rejection handler.
      */
-    @MXBeanDescription("Class name of current rejection handler.")
+    @MXBeanDescription("Handler class name for rejection tasks.")
```
vmalin, line 146: Do we need to change the description? RejectedExecutionHandler is actually a java SE class. If you think the change is needed, then
Handler class name for reject_ed_ tasks.

modules/core/src/main/java/org/apache/ignite/mxbean/ThreadPoolMXBean.java
```diff
@@ -143,7 +143,7 @@
      *
      * @return Class name of current rejection handler.
      */
-    @MXBeanDescription("Class name of current rejection handler.")
+    @MXBeanDescription("Handler class name for rejected tasks.")
```
sk0x50, line 146: Maybe, it would be good to copy&paste javadoc for See `java.util.concurrent.RejectedExecutionHandler`,
i.e. `The class name of a handler for tasks that cannot be executed by a thread pool.`
What do you think?

a-polyakov: Please check

modules/core/src/main/java/org/apache/ignite/mxbean/ThreadPoolMXBean.java
```diff
@@ -139,11 +139,13 @@
     public boolean isTerminating();
 
     /**
-     * Returns the class name of current rejection handler.
+     * The class name of the handler for invoked when execute cannot accept a task.
```
sk0x50, line 142: Returns the class name of the handler for tasks that cannot be executed.

PR #1527, opened 2020-10-28
GG-31159 ManagementEnabled incomprehensible metric description

vmalin:


modules/core/src/main/java/org/apache/ignite/mxbean/CacheMetricsMXBean.java
```diff
@@ -307,7 +307,7 @@
     public boolean isStatisticsEnabled();
 
     /** {@inheritDoc} */
-    @Override @MXBeanDescription("True if management is enabled.")
+    @Override @MXBeanDescription("Checks whether cache management via MBean is Enabled.")
```
vmalin, line 310: _True if_ cache management via MBean is _e_nabled.

modules/core/src/main/java/org/apache/ignite/cache/CacheMetrics.java
```diff
@@ -657,7 +657,7 @@
     public boolean isStatisticsEnabled();
 
     /**
-     * Checks whether management is enabled on this cache.
+     * Checks whether cache management via MBean is enabled.
```
sk0x50, line 660: To be honest, both versions look good to me. The original javadoc comes from JSR 107 - 12.2. MXBean Definitions.
`/**
* Checks whether management is enabled on this cache.
* <p>
* The default value is <code>false</code>.
*
* @return true if management is enabled
*/
boolean isManagementEnabled();`

So, it is up to you.

PR #1579, opened 2020-11-17
GG-24620

ascherbakoff:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsSingleMessage.java
```diff
@@ -356,6 +394,9 @@ public void exchangeStartTime(long exchangeStartTime) {
             if (partHistCntrs != null && partHistCntrsBytes == null)
                 partHistCntrsBytes0 = U.marshal(ctx, partHistCntrs);
 
+            if (partClearCntrs != null && partClearCntrsBytes == null)
+                partClearCntrsBytes0 = U.marshal(ctx, partHistCntrs);
```
agoncharuk, line 398: Copy-paste, should be U.marshal(ctx, partClearCntrs);
Need to check why this has not been detected in tests: essentially, partsHistCntr was always used instead of partClearCntrs.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/tree/AbstractPendingEntryInnerIO.java
```diff
@@ -42,7 +42,12 @@
         assert row.link != 0;
         assert row.expireTime != 0;
 
-        PageUtils.putLong(pageAddr, off, row.expireTime);
+        long expireTime = row.expireTime;
+
+        if (row.tombstone)
+            expireTime |= 0x8000000000000000L;
+
+        PageUtils.putLong(pageAddr, off, expireTime);
```
agoncharuk, line 50: storeByOffset & getLookupRow are identical for AbstractPendingEntryInnerIO and AbstractPendingEntryLeafIO. Should we move this code somewhere up the class hierarchy?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionsSingleMessage.java
ascherbakoff: Fixed and added a reproducer `testConsistencyOnCounterTriggeredRebalanceClearTombstonesNotCrd`.
Existing tests doesn't reveal the problem because coordinator is a supplier.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/tree/AbstractPendingEntryInnerIO.java
ascherbakoff: This looks like a refactoring job. Let's fix it in another ticket.

PR #1585, opened 2020-11-20
GG-31179 remove/replace log4j from the project.

slukyano:


modules/yardstick/pom.xml
```diff
@@ -58,12 +58,6 @@
             <version>${project.version}</version>
         </dependency>
 
-        <dependency>
-            <groupId>org.gridgain</groupId>
-            <artifactId>ignite-log4j</artifactId>
-            <version>${project.version}</version>
```
slukyano, line 64: Let's replace it with log4j2 instead. I'm afraid we'll lose decent logging in benchmarks if we just remove this module.

pom.xml
```diff
@@ -482,7 +482,7 @@
                                     <exportAntProperties>true</exportAntProperties>
                                 </configuration>
                             </execution>
-                            <execution>
+                            <!--<execution>
```
slukyano, line 485: I understand that this is a leftover from testing and this will go away

modules/hibernate-4.2/src/test/java/org/apache/ignite/cache/hibernate/HibernateL2CacheMultiJvmTest.java
```diff
@@ -51,6 +53,13 @@
     /** */
     private static final String TIMESTAMP_CACHE = "org.hibernate.cache.spi.UpdateTimestampsCache";
 
+    /** */
+    public static final String ROUTER_LOG_CFG = "modules/hibernate-4.2/config/ignite-log4j.xml";
+
+    @Override protected long getTestTimeout() {
+        return 1_000_000;
```
slukyano, line 60: Need to remove this

modules/hibernate-4.2/src/test/java/org/apache/ignite/cache/hibernate/HibernateL2CacheMultiJvmTest.java
```diff
@@ -229,6 +242,8 @@ public void testL2Cache() throws Exception {
         SessionFactory startHibernate(String igniteInstanceName) {
             log.info("Start hibernate on node: " + igniteInstanceName);
 
+            Logger.getRootLogger().setLevel(org.apache.log4j.Level.INFO);
```
slukyano, line 245: Please add a comment on why this is needed

modules/hibernate-4.2/src/test/java/org/apache/ignite/cache/hibernate/HibernateL2CacheMultiJvmTest.java
antkr: Removed.

modules/hibernate-4.2/src/test/java/org/apache/ignite/cache/hibernate/HibernateL2CacheMultiJvmTest.java
antkr: Added comment.

PR #1589, opened 2020-11-23
GG-30867 Hide redundant Spring logs to the lower(DEBUG) level

nva:


modules/log4j2/pom.xml
```diff
@@ -66,6 +66,12 @@
             <artifactId>log4j-core</artifactId>
             <version>${log4j.version}</version>
         </dependency>
+
+        <dependency>
+            <groupId>org.apache.logging.log4j</groupId>
+            <artifactId>log4j-jcl</artifactId>
```
dmekhanikov, line 72: Why do we add a bridge for JUL in the SLF4J, but not here?

nva: Done

modules/slf4j/src/main/java/org/apache/ignite/logger/slf4j/Slf4jLogger.java
```diff
@@ -65,7 +66,12 @@ public Slf4jLogger(Logger impl) {
 
         this.impl = impl;
 
-        quiet = Boolean.valueOf(System.getProperty(IGNITE_QUIET, "true"));
+        quiet = Boolean.parseBoolean(System.getProperty(IGNITE_QUIET, "true"));
+
+        // Optionally remove existing handlers attached to j.u.l root logger (since SLF4J 1.6.5)
+        SLF4JBridgeHandler.removeHandlersForRootLogger();
```
dmekhanikov, line 72: What does this thing do? Is it required?
The comment looks confusing. It says "Optionally", meaning that this step is not really required. We decided to perform it, but kept the "Optionally" word.

modules/slf4j/src/test/java/org/apache/ignite/logger/slf4j/Slf4jLoggerSelfTest.java
```diff
@@ -0,0 +1,181 @@
...
+        Logger log1 = Logger.getLogger(this.getClass().getName());
+        log1.info("Text that should be presence in logs");
+
+        Logger log2 = Logger.getLogger("org.springframework.context.ApplicationContext");
+        log2.info("INFO logs from org.springframework package should be excluded from logs");
+
+        String logContent = U.readFileToString(logFile.getAbsolutePath(), "UTF-8");
+
+        assertTrue("Logs from JUL logger should be present in log file",
+            logContent.contains("[INFO ][main][Slf4jLoggerSelfTest] Text that should be presence in logs"));
```
dmekhanikov, line 91:
```suggestion
            logContent.contains("[INFO ][main][Slf4jLoggerSelfTest] Text that should be present in logs"));
```

modules/slf4j/src/test/java/org/apache/ignite/logger/slf4j/Slf4jLoggerSelfTest.java
```diff
@@ -0,0 +1,181 @@
...
+    }
+
+    /**
+     * Check that JUL is redirected to Slf4j .
+     *
+     * Start the local node and check presence of log file.
+     * Check that this is really a log of a started node.
+     * Log something using JUL logging.
+     * Check that logs present in log file.
+     * Log something in INFO level from package that is blocked in log4j2 configuration using JUL logging.
```
dmekhanikov, line 73: Why do we say about log4j2 here? Is SLF backed by log4j? Let's clarify it somewhere in the comments.

dmekhanikov (changes requested):

modules/slf4j/src/test/java/org/apache/ignite/logger/slf4j/Slf4jLoggerSelfTest.java
nva: Fixed to slf4j. I don't think such details should be in the test description.

modules/slf4j/src/main/java/org/apache/ignite/logger/slf4j/Slf4jLogger.java
nva: Remove all other handlers that can catch logs from JUL. I think it required.
Remove "Optionally" word.

dmekhanikov: Let's remove the "(since SLF4J 1.6.5)" part as well then.

dmekhanikov (commented):
Please fix all entries of "should be presence" in tests.
Let's also move test log messages into a String variable, so that the strings are not duplicated in code.

modules/log4j2/src/test/java/org/apache/ignite/logger/log4j2/Log4j2LoggerSelfTest.java
```diff
@@ -152,14 +157,79 @@ public void testLogFilesTwoNodes() throws Exception {
...
+     * @throws Exception If error occurs.
+     */
+    @Test
+    public void testJULIsRedirectedToLog4j2() throws Exception {
+        assertEquals("org.apache.logging.log4j.jul.LogManager", LogManager.getLogManager().getClass().getName());
+        
+        File logFile = checkOneNode(0);
+
+        Logger log1 = Logger.getLogger(this.getClass().getName());
+        log1.info("Text that should be presence in logs");
```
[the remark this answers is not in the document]
nva: done

config/java.util.logging.properties
```diff
@@ -54,6 +54,9 @@ handlers=java.util.logging.ConsoleHandler, org.apache.ignite.logger.java.JavaLog
 #
 #org.apache.ignite.CourtesyConfigNotice.level=OFF
 
+org.springframework.level=WARNING
```
aealeksandrov, line 57: Why are these properties at the WARNING level? I think it would be better to use ERROR. This makes it easier to debug problems.

dmekhanikov:
The following messages appear now:
```
Jan 20, 2021 5:13:28 PM org.glassfish.grizzly.nio.NIOTransport start
INFO: Adjusting kernel thread pool to max size 5 to handle configured number of SelectorRunners
```

Maybe it's worth decreasing the log level for glassfish as well?

modules/log4j2/src/main/java/org/apache/ignite/logger/log4j2/Log4J2Logger.java
```diff
@@ -131,6 +131,11 @@ public Log4J2Logger() {
 
         quiet = quiet0;
         cfg = null;
+
+        // Remove existing handlers attached to j.u.l root logger.
+        Log4jBridgeHandler.removeHandlersForRootLogger();
```
aealeksandrov, line 136: As far as I can see, you removed all handlers attached to the j.u.l root registrar. Can you explain why? As I can see from the examples, you can simply add your handler via Log4jBridgeHandler.install().

If I understood this correctly after calling this method, all other handlers (logs appenders) configured by the client, for example, will be removed. Please correct me because I am not close to these methods, but it looks dangerous.

config/java.util.logging.properties
nva: This was copied from `config/ignite-log4j2.xml`, `config/ignite-log4j.xml` Should we change this too?

modules/log4j2/src/main/java/org/apache/ignite/logger/log4j2/Log4J2Logger.java
nva: This was an optional step from the examples. Let's not install the same handler twice. This can happen if you start/stop a node in the same jvm

config/java.util.logging.properties
dmekhanikov: ERROR will show fewer issues than WARN. Will it really make debugging easier?

nva:
> djusting kernel thread pool to max size 5 to handle configured number of SelectorRunners
Done

config/java.util.logging.properties
aealeksandrov: I guess that ERROR shows warns but WARN doesn't show errors. Please correct me if I am wrong.

nva: No, this level work as threshold. So WARN will be include ERROR and WARN, ERROR will include only ERROR

dmekhanikov (approved):

aealeksandrov (approved):

PR #1600, opened 2020-11-24
GG-31385

antkr:


slukyano (approved):
Let's merge this but I believe another, deeper change may also be warranted here. I'll leave a comment in Jira.

PR #1615, opened 2020-12-01
Gg 31878

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PagePartitionMetaIO.java
```diff
@@ -252,11 +253,49 @@ public long getUpdateTreeRoot(long pageAddr) {
...
-    public void setUpdateTreeRoot(long pageAddr, long listRoot) {
+    public boolean setUpdateTreeRoot(long pageAddr, long listRoot) {
         throw new UnsupportedOperationException("Partition update log is not supported by " +
                 "this PagePartitionMetaIO version: ver=" + getVersion());
     }
 
+    /**
+     * @param pageAddr Page address.
+     */
+    public int getEncryptedPageIndex(long pageAddr) {
```
vldpyatkov, line 264: What is the reason to pull all methods to parent and make them unsupported?

ascherbakoff: This is to avoid casts to specific PagePartitionMetaIO versions. Check gg-26420 saveStoreMetadata to see how it works.

PR #1616, opened 2020-12-02
GG-30897 Fix baseline attributes handling on node join

vmalin:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -1020,11 +1022,13 @@ private void finishRecovery() throws IgniteCheckedException {
 
             notifyMetastorageReadyForReadWrite();
 
+            kernalCtx.state().addAffinityAttrsToCurrBlt(kernalCtx.grid().localNode().consistentId(), kernalCtx.affinityNodeAttributes());
```
zstan, line 1025: trim it

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/baseline/IgniteNodeAttributesInBaselineTest.java
```diff
@@ -0,0 +1,171 @@
...
+        return cfg;
+    }
+
+    /** {@inheritDoc} */
+    @Override protected void beforeTest() throws Exception {
+        super.beforeTest();
+
+        cleanPersistenceDir();
+
+        IgniteEx ignite = startGrids(GRIDS_NUM);
```
zstan, line 89: can we use startGridsMultiThreaded ver here ?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/baseline/IgniteNodeAttributesInBaselineTest.java
```diff
@@ -0,0 +1,171 @@
...
+
+            assertNotNull(cause);
+
+            IgniteSpiException rootCause = (IgniteSpiException)cause.getCause();
+
+            assertNotNull(rootCause);
+
+            String msg = rootCause.getMessage();
+
+            String expMsg = "Some affinity attributes in cluster baseline topology are missing on joining node. " +
```
zstan, line 161: i suppose that this string need to be statically coded into appropriate class, additionally seems we need to reformat _Attr name:_ and _Attr val:_ strings - make it more readable.

modules/platforms/dotnet/Apache.Ignite.Core/IgniteConfiguration.cs
```diff
@@ -1322,6 +1342,12 @@ public bool IsDaemon
         [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
         public IDictionary<string, object> UserAttributes { get; set; }
 
+        /// <summary>
+        /// Gets or sets the affinity attributes for this node.
```
ptupitsyn, line 1346: Let's add a more detailed description here like we have on the Java side.

modules/platforms/dotnet/Apache.Ignite.Core/IgniteConfiguration.cs
```diff
@@ -1322,6 +1342,12 @@ public bool IsDaemon
...
+        /// </summary>
+        [SuppressMessage("Microsoft.Usage", "CA2227:CollectionPropertiesShouldBeReadOnly")]
+        public IDictionary<string, string> AffinityAttributes { get; set; }
```
ptupitsyn, line 1349: Thanks for adding the property on .NET side! Can you please also update `BaselineNode.cs` and add some checks to `PersistenceTest.TestBaselineTopology` to verify attribute propagation?

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/cluster/BaselineNode.java
```diff
@@ -67,4 +70,23 @@
...
     public Map<String, Object> attributes();
+
+    /**
+     * Gets affinity attribute by name.
+     *
+     * @param name of the attribute.
+     * @return Affinity attribute.
+     */
+    public default String affinityAttribute(String name) {
+        return (String)attributes().get(AFFINITY_ATTR_PREFIX + '.' + name);
```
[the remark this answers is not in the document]
vmalin: If we wand to avoid creating new entity we'll have to tell user to use specific prefix for affinity attributes. It might be easier for us but not for users

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
vmalin: more or less trimmed

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/baseline/IgniteNodeAttributesInBaselineTest.java
vmalin: Fixed readability a bit. I deliberately avoided usual "attrName=attrVal" notation because here main peace of information is attrName and attrVal is  optional

PR #1625, opened 2020-12-03
GG-31806 TDE - Phase-2. CLI process management

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/visor/encryption/VisorChangeMasterKeyTask.java
```diff
@@ -0,0 +1,62 @@
...
+         * @param masterKeyName Master key name.
+         * @param debug Flag indicating whether debug information should be printed into node log.
+         */
+        protected VisorChangeMasterKeyJob(String masterKeyName, boolean debug) {
+            super(masterKeyName, debug);
+        }
+
+        /** {@inheritDoc} */
+        @Override protected String run(String masterKeyName) throws IgniteException {
+            ignite.encryption().changeMasterKey(masterKeyName).get();
```
sk0x50, line 57: I think it makes sense to use a job continuation:
```
        if (fut == null) {
            fut = starting async operation;

            if (!fut.isDone()) {
                jobCtx.holdcc();

                fut.listen((IgniteInClosure<IgniteFuture<ResultT>>)f -> jobCtx.callcc());

                return null;
            }
        }

        return fut.get();
```

PR #1641, opened 2020-12-09
GG-31933

denis-chudov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/DurableBackgroundCleanupIndexTreeTask.java
```diff
@@ -192,4 +207,118 @@ public DurableBackgroundCleanupIndexTreeTask(
     @Override public String toString() {
         return S.toString(DurableBackgroundCleanupIndexTreeTask.class, this);
     }
+
+    /**
+     * H2Tree that can be used only to destroy it.
+     */
+    public static class H2TreeToDestroy extends H2Tree {
```
akalash, line 214: I didn't understand what this class is for? As I see H2Tree contains the same implementation of methods(temporaryReleaseLock, maxLockHoldTime)

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/LongDestroyDurableBackgroundTaskTest.java
```diff
@@ -641,6 +641,34 @@ public void testDestroyTaskLifecycle() throws Exception {
         assertTrue(taskLifecycleListener.check());
     }
 
+    /**
+     * Check that index deletion tasks are executed successfully when corresponding cache group is deleted.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testIrrelevantTasksAreCleared() throws Exception {
```
akalash, line 650: How does this test ensure that stopping node happens during index removing? What if the index drop happened before the node stopped and this test doesn't check anything.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/DurableBackgroundCleanupIndexTreeTask.java
denis-chudov: This class uses cache group context instead of cache context in these methods' implementation.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/LongDestroyDurableBackgroundTaskTest.java
denis-chudov: As in other tests, this is achieved by significant delay in org.apache.ignite.internal.processors.cache.persistence.db.LongDestroyDurableBackgroundTaskTest.H2TreeTest#destroyDownPages and size of the index guarantees that this method is called multiple times in one thread. But I can add assertion to ensure that task hadn't completed when grid is stopped.

PR #1643, opened 2020-12-09
GG-32007 Fixed erroneous synchronization for accessing transaction configuration

sk0x50:


modules/opencensus/src/test/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusDiscoveryTracingTest.java
```diff
@@ -116,7 +116,8 @@ public void testClientNodeCacheCreateTracing() throws Exception {
                 catch (Exception e) {
                     fail("Unable to create cache.");
                 }
-            }
+            },
+            GRID_CNT
```
sanpwc, line 120: As far as I understand it was made in order to overcome the problem of configuration unavailability on client node, cause of "async" nature of metastore data propagation to the client nodes.
Despite the fact that it's valid to use same node for both setting configuration and trigger operation to be traced, I'd rather use wait for condition approach within some utility method. What's more important we should ensure that any of given approaches is used everywhere but not only within given test. For example OpenCensusTxTracingTest also should be modified.
In other words I suggest to add some sort of `ensureTracingConfigurationWasSet(expectedConfig, cluster) ` that will be triggered after setting configuration and starting client nodes.

modules/opencensus/src/test/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusDiscoveryTracingTest.java
```diff
@@ -286,8 +290,25 @@ public void testClientNodeStopTracing() throws Exception {
...
+    ) throws Exception {
+        performClientBasedOperation(preparationOperations, operationToCheck, 0);
+    }
+
+    /**
+     * Perform preparation operations and operation to trigger span creation.
+     *
+     * @param preparationOperations Operations to run without tracing.
+     * @param operationToCheck Operations to run with tracing.
+     * @param nodeIdx Node id that that is used for changing tracing configuration.
```
sanpwc, line 303: If you'll continue with same-node-solution please add more meaningful comment that will describe why it's needed.

sk0x50: Ok, will do.

modules/opencensus/src/test/java/org/apache/ignite/internal/processors/monitoring/opencensus/OpenCensusDiscoveryTracingTest.java
sk0x50: Ok, it makes sense to me. I will try to modify tests in that way.

PR #1645, opened 2020-12-09
GG-31996 Choose specific deployment for equal classes loaded with dif…

zstan:
…ferent classLoaders


modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -231,23 +234,33 @@
...
      */
-    @Nullable private GridDeployment deployment(String alias) {
+    @Nullable private GridDeployment deployment(String alias, ClassLoader clsLdr) {
         Deque<GridDeployment> deps = cache.get(alias);
 
         if (deps != null) {
-            GridDeployment dep = deps.peekFirst();
+            for (GridDeployment dep : deps) {
+                if (dep.classLoader() == clsLdr) {
+                    if (log.isDebugEnabled())
```
vldpyatkov, line 251: If you used a trace, you should use the corresponding condition (`log.isTraceEnabled()`).

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -231,23 +234,33 @@
...
 
-            if (dep != null && !dep.undeployed())
-                return dep;
+                    return dep;
+                }
+            }
         }
 
+        if (log.isDebugEnabled())
+            log.trace("Deployment was not found for class with specific class loader [alias=" + alias +
```
vldpyatkov, line 261: Return debug here.

modules/core/src/test/java/org/apache/ignite/p2p/GridP2PMissedResourceCacheSizeSelfTest.java
```diff
@@ -289,4 +299,84 @@ public void testSize2ContinuousMode() throws Exception {
...
+     *
+     * @throws Exception if error occur.
+     */
+    @Test
+    public void testCheckTaskClassloaderCacheContinuousMode() throws Exception {
+        testCheckTaskClassloaderCache(DeploymentMode.CONTINUOUS);
+    }
+
+    /** */
+    public void testCheckTaskClassloaderCache(DeploymentMode depMode) throws Exception {
```
vldpyatkov, line 344: You should declare the parameter even if it is a test method.

PR #1659, opened 2020-12-16
Gg 32079

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -3511,6 +3513,30 @@ private CacheConfiguration getOrCreateConfigFromTemplate(String cacheName) throw
         }
     }
 
+    /**
+     * Prepares encryption key for the cache or return {@code null}
+     * if it will use a key with stored on metasatorage.
```
alievmirza, line 3518: key with stored on metasatorage. -> key which is stored in the metasatorage.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -3511,6 +3513,30 @@ private CacheConfiguration getOrCreateConfigFromTemplate(String cacheName) throw
...
+     * @param proposedKey Proposed encryption key.
+     * @return Encryption key.
+     */
+    private byte[] prepareEncryptionGroupKey(String cacheName, @Nullable CacheConfiguration ccfg, byte[] proposedKey) {
+        if (proposedKey != null) {
+            int grpId = CU.cacheGroupId(cacheName, ccfg == null ? null : ccfg.getGroupName());
+
+            if (ctx.encryption().groupKey(grpId) != null) {
+                U.warn(log, "Encryption key for this cache has already existed," +
+                    " the new key was ignored [gepId=" + grpId + ", cache=" + cacheName + ']');
```
alievmirza, line 3531: gepId -> grpId

alievmirza (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
vldpyatkov: Done.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
vldpyatkov: Done.

PR #1693, opened 2020-12-30
GG-32131 [IGNITE-13865] Support DateTime as a key or value in .NET and Java

ptupitsyn:
* Add BinaryConfiguration.ForceTimestamp as an easier way to enable Timestamp mode globally
* Fix BinarySystemHandlers to allow DateTime in Timestamp format as cache key and value


isapego (approved):
Looks good.

PR #1699, opened 2021-01-04
GG-32290 - Move cluster tag API from IgniteClusterEx to IgniteCluster

aries-3:


modules/core/src/main/java/org/apache/ignite/internal/cluster/IgniteClusterEx.java
```diff
@@ -40,27 +34,4 @@
      * @return Unique cluster ID.
      */
     public UUID id();
```
sergey-chugunov-1985, line 36: Cluster ID is part of the same functionality as Cluster Tag and should be republished back to IgniteCluster interface as well.

After that we should delete IgniteClusterEx interface as unnecessary.

sergey-chugunov-1985 (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/cluster/IgniteClusterEx.java
aries-3: done

PR #1704, opened 2021-01-06
GG-32292 CQ SystemView shows full data for each cq

aries-3:


modules/core/src/main/java/org/apache/ignite/spi/systemview/view/ContinuousQueryView.java
```diff
@@ -44,6 +44,9 @@
     /** Routine id. */
     private final UUID routineId;
 
+    /** Class name of the local listener */
+    private String localListener;
```
slukyano, line 48: Let's also add a consistent ID of the node that has registered the CQ (the one that is running the local listener)

aries-3: Ok, done!

modules/core/src/main/java/org/apache/ignite/internal/processors/continuous/ContinuousQueriesCollection.java
```diff
@@ -0,0 +1,88 @@
...
+    /** */
+    public ContinuousQueriesCollection(String locIgniteName, Supplier<Integer> sizeFunc) {
+        this.sizeFunc = sizeFunc;
+        this.locIgniteName = locIgniteName;
+    }
+
+    /** {@inheritDoc} */
+    @NotNull
+    @Override public Iterator<ContinuousQueryView> iterator() {
+        Ignite ignite = Ignition.ignite(locIgniteName);
```
slukyano, line 49: Not a fan of calling `Ignition.ignite` in internal code. Can't we pass `Ignite` instance instead of the instance name to the collection constructor?

PR #1709, opened 2021-01-11
GG-32308 Broken sql test after GG-18613 ticket

ingvard:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/SqlQueryHistorySelfTest.java
```diff
@@ -519,7 +522,7 @@ private void checkQueryBooleanParams(boolean exp, boolean loc) {
         assertEquals(exp, hist.distributedJoins());
         assertEquals(exp, hist.enforceJoinOrder());
         assertEquals(loc, hist.local());
-        assertEquals(exp, hist.lazy());
+        assertEquals(activateLazyByDflt || exp, hist.lazy());
```
ygerzhedovich, line 525: it looks strange. Let's use extra parameters for the ''checkQueryBooleanParams'' method

PR #1714, opened 2021-01-12
GG-31193

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -194,7 +194,10 @@
     CACHE_GROUP_KEY_CHANGE(50),
 
     /** Possibility to safe deactivation, take into account pure in memory caches with possible data loss.*/
-    SAFE_CLUSTER_DEACTIVATION(51);
+    SAFE_CLUSTER_DEACTIVATION(51),
+
+    /** Custom snapshot operations. */
+    CUSTOM_SNAPSHOT_OPERATIONS(52);
```
tkalkirill, line 200: ID 52 is already taken.
https://ggsystems.atlassian.net/wiki/spaces/GG/pages/1192198276/Community+edition+features+list

PR #1720, opened 2021-01-15
GG-32092 IgniteCompute tasks with same name, running from one node an…

vldpyatkov:
…d different ClassLoaders can lead to OOM


modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -247,26 +245,11 @@
...
 
-        if (getBoolean(IGNITE_DEPLOYMENT_PRESERVE_LOCAL)) {
-            if (deps != null) {
-                for (GridDeployment dep : deps) {
-                    if (dep.classLoader() == clsLdr) {
-                        if (log.isTraceEnabled())
-                            log.trace("Deployment was found for class with specific class loader [alias=" + alias +
-                                ", clsLdr=" + clsLdr + "]");
+        if (deps != null) {
+            GridDeployment dep = deps.peekFirst();
```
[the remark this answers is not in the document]
zstan: This is strange place as for me, i prefer to replace clsLdr argument = revert old logic and trigger the ticket with this research, what do you think ? I can assign new ticket to myself.

modules/core/src/main/java/org/apache/ignite/internal/managers/deployment/GridDeploymentLocalStore.java
```diff
@@ -325,7 +308,13 @@ private GridDeployment deploy(DeploymentMode depMode, ClassLoader ldr, Class<?>
...
                 }
 
-                IgniteUuid ldrId = IgniteUuid.fromUuid(ctx.localNodeId());
+                IgniteUuid ldrId;
+
+                // IsolatedDeployment wait for continiously grow of ldrId
+                if (depMode == DeploymentMode.ISOLATED || depMode == DeploymentMode.PRIVATE)
+                    ldrId = IgniteUuid.fromUuid(ctx.localNodeId());
+                else
+                    ldrId = new IgniteUuid(ctx.localNodeId(), ldr.hashCode());
```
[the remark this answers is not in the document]
zstan: Try to explain, without this trick (i use testCheckTaskClassloaderCache for explanation and mention shared dep mode)
1. now we have deterministic ID here, so repeatedly call  _client.compute().execute(_ will gwnwrate the same ID.
2. From server side : org.apache.ignite.internal.managers.deployment.GridDeploymentManager#getGlobalDeployment
we will find already deployed meta here:
_GridDeployment dep = verStore.searchDeploymentCache(meta);_ and return, thus no additional GridDeploymentPerVersionStore.SharedDeployment instance will be produced.

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
```diff
@@ -110,7 +105,16 @@
...
+     * Finds class loader for the given class.
+     *
+     * @param rsrcName Class name or class alias to find class loader for.
+     * @return Deployed class loader, or {@code null} if not deployed.
+     *
+     * @deprecated It is recommended to use {@link DeploymentSpi#findResource(String, ClassLoader)} instead.
+     * This method can return incorrect resource.
+     */
+    @Deprecated
     @Nullable @Override public DeploymentResource findResource(String rsrcName) {
```
vldpyatkov, line 118: Call findResource(rsrcName, null) here.

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
```diff
@@ -146,6 +150,68 @@
...
+
+            try {
+                Class<?> cls = U.forName(clsName, clsLdr);
+
+                assert cls != null;
+
+                // Return resource.
+                return new DeploymentResourceAdapter(rsrcName, cls, clsLdr);
+            }
+            catch (ClassNotFoundException ignored) {
```
vldpyatkov, line 207: Print a warning, otherwise we never find a problem.

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
```diff
@@ -171,38 +237,33 @@ private String getResourceName(String clsName, Map<String, String> rsrcs) {
...
 
-            if (rmvClsLdrs != null) {
-                for (ClassLoader cldLdr : rmvClsLdrs)
-                    onClassLoaderReleased(cldLdr);
-            }
+            newRsrcs = addResource(ldr, clsLdrRsrcs, rsrc);
         }
 
-        return !F.isEmpty(newRsrcs);
+        return !F.isEmpty(newRsrcs) && !move;
```
vldpyatkov, line 256: This is a suspecting expression.
I think you should stay it as before (!F.isEmpty(newRsrcs)).

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
```diff
@@ -171,38 +237,33 @@ private String getResourceName(String clsName, Map<String, String> rsrcs) {
...
+                ldrRsrcs.remove(ldr);
 
-        if (!getBoolean(IGNITE_DEPLOYMENT_PRESERVE_LOCAL)) {
-            Collection<ClassLoader> rmvClsLdrs = null;
+                move = true;
+            }
 
-            if (!F.isEmpty(newRsrcs)) {
-                rmvClsLdrs = new LinkedList<>();
+            ConcurrentMap<String, String> old = ldrRsrcs.putIfAbsent(ldr,
```
vldpyatkov, line 247: Try to use a ldrRsrcs.compute method here.
It may get rid of synchronization and change order in the map in that places where it requires.

zstan: don`t understand how .compute will help to move forward here.

zstan: ok, i fix it.

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
zstan: done

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
zstan: done

modules/core/src/main/java/org/apache/ignite/spi/deployment/local/LocalDeploymentSpi.java
zstan: done

PR #1727, opened 2021-01-21
GG-32437 Create a JMX metric to show realtime sql engine memory utilization

akorensh:
…zation


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -325,6 +326,15 @@ public long memoryLimit() {
...
     }
 
+    /**
+     *
+     * @return memory in bytes currently left available for the queries on this node.
+     */
+    public long getFreeMem() {
+        return ((LongMetric)ctx.metric().registry(SqlMemoryStatisticsHolder.SQL_QUOTAS_REG_NAME).findMetric("freeMem")).value();
+    }
+
```
AMashenkov, line 336: Why SqlMemoryManager refers to MetricRegistry for a metric?
SqlMemoryManager is a metric supplier, and JMX bean should refers to a registry, then the registry should ask memory manager for a calculated "freemem".

AMashenkov (changes requested):

AMashenkov (approved):

PR #1744, opened 2021-01-29
GG-32433

tkalkirill:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -796,131 +797,148 @@ private void mvccPreloadEntries(
...
                 }
-            }
-            finally {
-                ctx.database().checkpointReadUnlock();
+                finally {
+                    ctx.database().checkpointReadUnlock();
+                }
             }
         }
+        finally {
```
vldpyatkov, line 874: Do not use try-finaly block here.
If there is any exception has happened just miss an update in the metric.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -796,131 +797,148 @@ private void mvccPreloadEntries(
...
                 }
+                finally {
+                    ctx.database().checkpointReadUnlock();
+                }
             }
-            finally {
-                ctx.database().checkpointReadUnlock();
-            }
+        }
+        finally {
```
vldpyatkov, line 940: Get rid of try-finaly here as I told above.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -781,11 +781,12 @@ private void ownPartition(
     /**
      * Adds mvcc entries with theirs history to partition p.
      *
+     * @param topVer Topology version.
      * @param node Node which sent entry.
      * @param p Partition id.
      * @param infos Entries info for preload.
-     * @param topVer Topology version.
      * @throws IgniteInterruptedCheckedException If interrupted.
+     * @throws IgniteCheckedException If failed.
```
vldpyatkov, line 789: It was a definitely worth comment.
Please, write useful comment if you want to add this javadoc.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -796,9 +796,12 @@ private void mvccPreloadEntries(
         if (!infos.hasNext())
             return;
 
+        // Received keys by caches, for statistics.
+        Map<Integer, AtomicLong> receivedKeys = new HashMap<>();
```
vldpyatkov, line 800: Atomic is excessive here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -940,13 +952,11 @@ private boolean preloadEntry(
         int p,
         GridCacheEntryInfo entry,
         AffinityTopologyVersion topVer,
-        GridCacheContext cctx
+        @Nullable GridCacheContext<?, ?> cctx
     ) throws IgniteCheckedException {
         assert !grp.mvccEnabled();
```
aimiebell, line 957: I detect that this code is problematic. According to the [Bad practice (BAD_PRACTICE)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#bad-practice-bad-practice), [Nm: Use of identifier that is a keyword in later versions of Java (NM_FUTURE_KEYWORD_USED_AS_IDENTIFIER)](https://spotbugs.readthedocs.io/en/stable/bugDescriptions.html#nm-use-of-identifier-that-is-a-keyword-in-later-versions-of-java-nm-future-keyword-used-as-identifier).
The identifier is a word that is reserved as a keyword in later versions of Java, and your code will need to be changed in order to compile it in later versions of Java.

modules/core/src/test/java/org/apache/ignite/internal/util/GridMutableLongTest.java
```diff
@@ -0,0 +1,60 @@
...
+ * Class for testing {@link GridMutableLong}.
+ */
+public class GridMutableLongTest extends GridCommonAbstractTest {
+    /**
+     * Checking the correctness of object initialization.
+     */
+    @Test
+    public void testInitialization() {
+        assertEquals(0, new GridMutableLong().get());
+        assertEquals(5, new GridMutableLong(5).get());
```
sk0x50, line 32: To be honest, I don't see the difference between 5 and 10. It seems to me, one assertion can be skipped.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/rebalancing/RebalanceMetricsTest.java
```diff
@@ -0,0 +1,111 @@
...
+        }
+
+        IgniteEx n1 = startGrid(1);
+        awaitPartitionMapExchange();
+
+        for (CacheConfiguration<?, ?> cacheCfg : cacheConfigs) {
+            GridCacheContext<?, ?> cacheCtx0 = n0.context().cache().cache(cacheCfg.getName()).context();
+            GridCacheContext<?, ?> cacheCtx1 = n1.context().cache().cache(cacheCfg.getName()).context();
+
+            assertEquals(cacheCfg.getName(), cacheCfg.isStatisticsEnabled(), cacheCtx0.statisticsEnabled());
```
sk0x50, line 93: I would like a human-readable message here and below.
This one does not seem good enough:
```
java.lang.AssertionError: default0 
Expected :false
Actual   :true
```

modules/core/src/test/java/org/apache/ignite/internal/util/GridMutableLongTest.java
tkalkirill: I think that an additional check will not hurt.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/distributed/rebalancing/RebalanceMetricsTest.java
tkalkirill: I don't think it will help much.

PR #1745, opened 2021-02-01
GG-32544 PK becomes corrupted after node restart

korlov42:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -640,7 +640,7 @@ public static void processBinaryMeta(GridKernalContext ctx, QueryEntity qryEntit
             d.addProperty(prop, false);
         }
 
-        if (!isKeyClsSqlType)
+        if (!isKeyClsSqlType && qryEntity instanceof QueryEntityEx && ((QueryEntityEx)qryEntity).isPreserveKeysOrder())
```
ygerzhedovich, line 643: Which cases qryEntity couldn't be an instance of  QueryEntytyEx?

korlov42: In case a table created from Java API. Actually it COULD be, but it SHOULDN'T be. But for non-sql tables this is not a problem anyway since the unwrap exists only for tables created via SQL API

ygerzhedovich: As of now yes, in the future seems to unwrap could be extended to all cases. However, it will be an issue for another ticket.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -1075,9 +1077,15 @@ private static QueryEntity toQueryEntity(GridSqlCreateTable createTbl) {
 
             res.setKeyFieldName(pkCol.columnName());
         }
-        else
+        else {
             res.setKeyFields(createTbl.primaryKeyColumns());
 
+            if (IgniteFeatures.allNodesSupports(ctx, F.view(ctx.discovery().allNodes(),
+                IgniteDiscoverySpi.SRV_NODES), IgniteFeatures.SPECIFIED_SEQ_PK_KEYS)
+            )
```
AMashenkov, line 1085: Should we check if we are in the middle of rolling upgrade?

korlov42: IgniteFeatures.allNodesSupports already handles this case

PR #1747, opened 2021-02-01
GG-32247

makedonskaya:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounterErrorWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+package org.apache.ignite.internal.processors.cache;
+
+import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.IgniteLogger;
+import org.apache.ignite.internal.util.GridLongList;
+import org.apache.ignite.internal.util.typedef.internal.SB;
+import org.apache.ignite.internal.util.typedef.internal.U;
+import org.jetbrains.annotations.Nullable;
+
+import java.util.Iterator;
```
sanpwc, line 26: Incorrect import order and unnecessary blank line.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounterErrorWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+    @Override public long reserve(long delta) {
+        String before = toString();
+
+        try {
+            return delegate.reserve(delta);
+        } catch (AssertionError e) {
+            SB sb = new SB();
+
+            sb.a("[op=reserve" +
+                ", grpId=" + grp.groupId() +
```
sanpwc, line 65: Up to you: what about using sb.append instead of concatenation here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounterTrackingImpl.java
```diff
@@ -454,8 +455,28 @@ public boolean within(long cntr) {
...
     @Override public String toString() {
-        return "Counter [lwm=" + get() + ", holes=" + queue +
-            ", maxApplied=" + highestAppliedCounter() + ", hwm=" + reserveCntr.get() + ", clearCntr=" + tombstoneClearCounter() + ']';
+        String quequeStr;
+        long hwm;
+
+        synchronized (this) {
+            quequeStr = queue.toString();
+
+            hwm = reserveCntr.get();
```
sanpwc, line 464: Should we also get lwm within synchronized block?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounterErrorWrapper.java
```diff
@@ -0,0 +1,188 @@
...
+    public PartitionUpdateCounterErrorWrapper(int partId, PartitionUpdateCounter delegate) {
+        this.partId = partId;
+        this.grp = delegate.context();
+        this.log = grp.shared().logger(getClass());
+        this.delegate = delegate;
+    }
+
+    /** {@inheritDoc} */
+    @Override public long reserve(long delta) {
+        String before = toString();
```
sanpwc, line 57: Why we need before? It seems confusing to prepare string interpretation on every reserve call.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/PartitionUpdateCounterErrorWrapper.java
```diff
@@ -0,0 +1,185 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public long get() {
+        return delegate.get();
+    }
+
+    /** {@inheritDoc} */
+    @Override public boolean equals(Object o) {
+        if (o instanceof PartitionUpdateCounterErrorWrapper) {
```
sk0x50, line 136: Please remove unnecessary brackets.

PR #1757, opened 2021-02-04
GG-32381 [IGNITE-14003] Improve heap consumption on reserving rebalance iterator.

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
```diff
@@ -1186,20 +1184,30 @@ private long allocateForTree() throws IgniteCheckedException {
             }
 
             @Override protected boolean onHasNext() throws IgniteCheckedException {
+                if (cur == null)
+                    cur = data.cursor(CacheDataRowAdapter.RowData.FULL_WITH_HINTS);
```
sk0x50, line 1188: I think it should be as follows
```
        final GridCursor<? extends CacheDataRow> cur = grp.mvccEnabled() ? data.cursor(FULL_WITH_HINTS) :	
            data.cursor(IgniteCacheOffheapManager.DATA_AND_TOMBSONES);
```

PR #1760, opened 2021-02-05
GG-32342 Add time of first page fetching.

vsisko:


nva (commented):
Can we add a tests for this?

wirtsleg (changes requested):
68 line:
> long start = U.currentTimeMillis();

Let's change it to `System.currentTimeMillis()` too.

And VisorQueryHolder#duration is calculated with `U.currentTImeMillis()`.

vsisko:
Changing getting of current time from IgniteUtils to System. Added test.

nva (approved):

wirtsleg (approved):

PR #1767, opened 2021-02-11
GG-31913

denis-chudov:


modules/core/src/test/java/org/apache/ignite/internal/IgniteClientFailuresTest.java
```diff
@@ -115,28 +122,40 @@ public void testNoMessagesFromFailureProcessor() throws Exception {
...
+        boolean waitRes = GridTestUtils.waitForCondition(() -> (cl.topology(cl.topologyVersion()).size() == 4),
             20_000);
 
-        checkCacheOperations(client01.cache("cache0"));
-
         assertTrue(waitRes);
+
+        checkCacheOperations(client02.cache("cache0"));
+
+        assertEquals( 4, srv0.context().discovery().allNodes().size());
```
denis-chudov, line 153: space not needed

modules/core/src/test/java/org/apache/ignite/internal/IgniteClientFailuresTest.java
```diff
@@ -239,4 +258,27 @@ private void breakClient(IgniteEx client) {
...
+            "Kernal",
+            clazz.getSimpleName());
+
+        MBeanServer mbeanSrv = ManagementFactory.getPlatformMBeanServer();
+
+        if (!mbeanSrv.isRegistered(mbeanName))
+            fail("MBean is not registered: " + mbeanName.getCanonicalName());
+
+        return MBeanServerInvocationHandler.newProxyInstance(mbeanSrv, mbeanName, ClusterMetricsMXBean.class, true);
+    }
```
denis-chudov, line 283: you could use GridCommonAbstractTest#getMxBean instead.

PR #1774, opened 2021-02-14
GG-32415: Fix timezone for thin client SQL

isapego:


modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -310,4 +314,49 @@ public static String translateSqlWildcardsToRegex(String sqlPtrn) {
...
+     */
+    public static java.util.Date convertWithTimeZone(java.util.Date date, TimeZone tzFrom, TimeZone tzTo) {
+        if (tzTo == null || tzFrom == null || tzTo.getID().equals(tzFrom.getID()))
+            return date;
+
+        return new Time(convertWithTimeZone(tzFrom, tzTo, date.getTime()));
+    }
+
+    /**
+     * Change timezone from one to another timezone.
```
ptupitsyn, line 346:
```suggestion
     * Change timezone from one to another.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -310,4 +314,49 @@ public static String translateSqlWildcardsToRegex(String sqlPtrn) {
...
+     *
+     * @param tzFrom Timezone from which to convert.
+     * @param tzTo Timezone to convert to/
```
ptupitsyn, line 349:
```suggestion
     * @param tzTo Timezone to convert to.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
```diff
@@ -73,20 +75,22 @@
...
     public static final ClientListenerProtocolVersion VER_1_7_0 = ClientListenerProtocolVersion.create(1, 7, 0);
 
-    /**
-     * Version 1.7.1. User Attributes introduced. New version is needed because user attributes passed in handshake.
-     */
+    /** Version 1.7.1. User Attributes introduced. New version is needed because user attributes passed in handshake. */
     public static final ClientListenerProtocolVersion VER_1_7_1 = ClientListenerProtocolVersion.create(1, 7, 1);
 
+    /** Version 1.8.0. Timezone fix introduced. New version is needed because timezone passed in handshake. */
+    public static final ClientListenerProtocolVersion VER_1_8_0 = ClientListenerProtocolVersion.create(1, 8, 0);
```
ptupitsyn, line 82: I would prefer to avoid adding a new protocol version - future maintenance costs are high, especially because this won't be merged to Apache Ignite.

Can we pass the timezone as a User Attribute? I think it'll better fit the nature of this whole fix.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientSqlUtils.java
```diff
@@ -0,0 +1,118 @@
...
+    @Nullable
+    public static Object readSqlField(BinaryRawReaderEx reader, ClientProtocolContext protoCtx)
+            throws BinaryObjectException {
+        BinaryReaderExImpl reader0 = (BinaryReaderExImpl) reader;
+        TimeZone serverTz = TimeZone.getDefault();
+        TimeZone clientTz = protoCtx.clientTimeZone();
+
+        byte type = reader0.readByte();
+        switch (type) {
+            case GridBinaryMarshaller.DATE: {
```
ptupitsyn, line 73: Redundant `{` / `}` as per code style, here and below

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
isapego: This is a good suggestion. Will do.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientSqlUtils.java
isapego: Actually, according to guideline the only case when curly braces are not needed is when the following expression is one-liner (there we have 2 lines of code in every code block)

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
```diff
@@ -310,4 +314,49 @@ public static String translateSqlWildcardsToRegex(String sqlPtrn) {
 
         return toRegex.substring(1);
     }
-}
\ No newline at end of file
+
+    /**
+     */
+    public static Timestamp convertWithTimeZone(Timestamp ts, TimeZone tzFrom, TimeZone tzTo) {
```
ptupitsyn, line 320: Is it possible for `ts` to be null here?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
```diff
@@ -205,6 +206,12 @@ public ClientProtocolContext currentProtocolContext() {
         if (currentProtocolContext.isFeatureSupported(USER_ATTRIBUTES))
             userAttrs = reader.readMap();
 
+        if (userAttrs != null) {
+            String clientTzId = userAttrs.get(USER_ATTR_TIMEZONE);
+
+            currentProtocolContext.clientTimeZone(TimeZone.getTimeZone(clientTzId));
```
ptupitsyn, line 212: NPE when `clientTzId` is `null`

modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/SqlListenerUtils.java
isapego: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/ClientConnectionContext.java
isapego: Good catch

ptupitsyn (approved):

PR #1778, opened 2021-02-16
GG-32708 Fix flaky CacheGroupsMetricsRebalanceTest#testRebalance

tkalkirill:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheGroupsMetricsRebalanceTest.java
```diff
@@ -192,24 +194,22 @@ public void testRebalance() throws Exception {
...
+                throw new AssertionError();
             }
+
+            return false;
         }, EventType.EVT_CACHE_REBALANCE_STOPPED);
 
         assertTrue(l1.await(5, TimeUnit.SECONDS));
 
-        ignite = ignite(4);
+        ignite = ignite(1);
```
vldpyatkov, line 212: You already have  initialized Ignite before.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheGroupsMetricsRebalanceTest.java
```diff
@@ -173,11 +173,13 @@
     }
 
     /**
+     * Checking the correctness of {@link CacheMetrics#getRebalancingKeysRate}.
```
vldpyatkov, line 176: We try to start all method description from a verb with "s" in the tail everywhere.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheGroupsMetricsRebalanceTest.java
```diff
@@ -192,25 +194,23 @@ public void testRebalance() throws Exception {
...
-                l1.countDown();
+        ignite = startGrid(1);
 
-                try {
-                    assertTrue(l2.await(5, TimeUnit.SECONDS));
-                }
-                catch (InterruptedException e) {
-                    throw new AssertionError();
-                }
+        ignite.events().localListen((IgnitePredicate<Event>)evt -> {
```
sk0x50, line 199: This listener must be specified via `IgniteConfiguration` otherwise there is no guarantee that rebalancing is not already finished.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheGroupsMetricsRebalanceTest.java
```diff
@@ -192,25 +194,23 @@ public void testRebalance() throws Exception {
         final CountDownLatch l1 = new CountDownLatch(1);
```
sk0x50, line 194: I would prefer meaningful names: `startRebalanceLatch` and `blockRebalanceLatch` for instance.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheGroupsMetricsRebalanceTest.java
```diff
@@ -192,25 +194,23 @@ public void testRebalance() throws Exception {
...
-                }
-                catch (InterruptedException e) {
-                    throw new AssertionError();
-                }
+        ignite.events().localListen((IgnitePredicate<Event>)evt -> {
+            l1.countDown();
 
-                return false;
+            try {
+                assertTrue(l2.await(5, TimeUnit.SECONDS));
```
sk0x50, line 203: It seems to me, five seconds is small enough and so it can lead to false-positive results. Anyway, it is up to you.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheGroupsMetricsRebalanceTest.java
```diff
@@ -192,25 +194,23 @@ public void testRebalance() throws Exception {
...
+            }
+            catch (InterruptedException e) {
+                throw new AssertionError();
```
sk0x50, line 206: Well, this assertion will be thrown in rebalance thread which does not affect a test-runner itself. Therefore it can be ignored by this test.

PR #1780, opened 2021-02-16
GG-30596

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -210,6 +210,9 @@
      * is improper for PITR*/
     IMPROPER_SCHEDULE_FOR_PITR_WARNING(56),
 
+    /** Chains of snapshot operations. */
+    SNAPSHOT_OPERATIONS_CHAINING(57),
```
vmalin, line 214: New features should be added to this list to book the number: https://ggsystems.atlassian.net/wiki/spaces/GG/pages/1192198276/Community+edition+features+list

PR #1808, opened 2021-03-05
GG-32559

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1398,7 +1397,7 @@ public void requestPartitions() {
 
             if (isDone()) {
                 assert !result() : "Rebalance future was done, but partitions never requested [grp="
-                    + grp.cacheOrGroupName() + ", topVer=" + topologyVersion() + "]";
+                    + grp.cacheOrGroupName() + ", topVer=" + topologyVersion() + ", rebalanceId=" + rebalanceId + "]";
```
sk0x50, line 1400: please change "]" to ']' here and below

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1556,6 +1531,9 @@ public void ownPartitionsAndFinishFuture(AffinityTopologyVersion topVer) {
                 return;
             }
 
+            if (this.rebalanceId != rebalanceId)
```
sk0x50, line 1534: Please add debug message here.

PR #1810, opened 2021-03-09
GG-32759 Fixed baseline auto adjustment triggered by merged exchanges.

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cluster/baseline/autoadjust/BaselineAutoAdjustScheduler.java
```diff
@@ -55,29 +55,59 @@ public BaselineAutoAdjustScheduler(GridTimeoutProcessor timeoutProcessor,
...
             baselineTimeoutObj = new BaselineMultiplyUseTimeoutObject(
                 baselineAutoAdjustData,
                 delay, baselineAutoAdjustExecutor,
                 timeoutProcessor,
                 log
             )
         );
+
+        if (log.isDebugEnabled()) {
+            log.info("Scheduled new baseline timeout object [data=" +
```
alievmirza, line 100: Seems that this log might be wrong if `added` == `false`

sk0x50: Thanks! Fixed

PR #1818, opened 2021-03-12
GG-24525 Suppress redundant WAL archive warning message in non-persistent mode

sashapolo:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -622,28 +621,37 @@ private void validateConfiguration(DataStorageConfiguration memCfg) throws Ignit
...
+                    "non-persistent mode, so this parameter will be ignored");
+            }
+            return;
+        }
+
+        if (memCfg.isWalHistorySizeParameterUsed())
+            LT.warn(log,
+                "DataRegionConfiguration.walHistorySize property is deprecated and is no longer supported. " +
+                    "It will be ignored and DataRegionConfiguration.maxWalArchiveSize property will be used " +
+                    "instead to control eviction of archived WAL files"
```
sergey-chugunov-1985, line 644: Lets use work 'removing' or similar here instead of eviction. Eviction term is already used in our configuration with different meanings.

Otherwise patch looks good to me.

sergey-chugunov-1985 (changes requested):

PR #1822, opened 2021-03-17
GG-32412 JDBC client 8.7.2-p15 fails to connect to 8.7.33

korlov42:


modules/core/src/main/java/org/apache/ignite/internal/processors/odbc/jdbc/JdbcConnectionContext.java
```diff
@@ -185,34 +186,40 @@ public JdbcConnectionContext(GridKernalContext ctx, GridNioSession ses, GridSpin
...
 
                 updateBatchSize = JdbcUtils.readNullableInteger(reader);
-
-                if (ver.compareTo(VER_2_8_1) >= 0) {
-                    if (reader.readBoolean())
-                        maxMemory = reader.readLong();
-                }
             }
+            catch (Exception ex) {
+                if (ver.compareTo(VER_2_8_0) != 0)
```
AMashenkov, line 200: What's wrong with 2.8.0?
Why doesn't exeption thrown for all versions <2.8.0 (or >2.8.0)?

korlov42: this version is broken, because some time ago a cancellation support was back ported to 8.7. This brought a new protocol version (2.8.0), but other 2.8.0-related features were not back ported.

We already tried to solve the problem with connecting valid_client -> invalid_server, added the limitedV2_8_0Enabled connection option. Now it's time to fix connection for invalid_client->valid_server

PR #1828, opened 2021-03-18
GG-30557 Include Control Center Agent with shaded dependencies into the build.

dmekhanikov:
Related PR in the agent repo: https://github.com/ggprivate/control-center-agent/pull/189


assembly/dependencies-apache-ignite.xml
```diff
@@ -205,6 +205,9 @@
         </fileSet>
         <fileSet>
             <directory>${basedir}/../control-center-agent/modules/control-center-agent/target/libs</directory>
+            <includes>
+                <include>commons-logging*.jar</include>
+            </includes>
```
nva, line 210: I don't think that we should filter agent 's libs in CE repository. Let's add exclusion to `copy-libs-dev` target in CCA repository. Like we do for org.apache.ignite,org.gridgain artifacts

config/ignite-log4j.xml
```diff
@@ -101,6 +101,16 @@
...
 
+    <category name="org.gridgain.control.springframework">
+        <level value="WARN"/>
+    </category>
+    <category name="org.gridgain.control.glassfish.grizzly">
+        <level value="WARN"/>
+    </category>
+    <category name="org.gridgain.control.jackson">
+        <level value="ERROR"/>
+    </category>
```
nva, line 112: Why are you changing the log level for Jackson? We haven't done that before.

vsisko (approved):

config/ignite-log4j.xml
dmekhanikov: The following code starts acting up because of the relocation:
 ```java
try {
    Class<?> cls = Class.forName("com.fasterxml.jackson.databind.ext.Java7HandlersImpl");
    impl = (Java7Handlers) ClassUtil.createInstance(cls, false);
} catch (Throwable t) {
    // 09-Sep-2019, tatu: Could choose not to log this, but since this is less likely
    //    to miss (than annotations), do it
    java.util.logging.Logger.getLogger(Java7Handlers.class.getName())
        .warning("Unable to load JDK7 types (java.nio.file.Path): no Java7 type support added");
}
```
It affects the ability to serialize nio classes which we don't use.

assembly/dependencies-apache-ignite.xml
dmekhanikov: We could make `copy-libs-dev` copy only the `commons-logging` jar into the `libs` directory, but `copy-libs-dev` is executed at the test-compile stage. At that point the shading hasn't taken place yet. 
Am I missing something?

nva: yes, `copy-libs-dev` is executed at the test-compile stage. In `libs` directory you should copy deps that should be placed in module dir in release.

config/ignite-log4j.xml
nva: It looks like a road to nowhere. How many reflection problems have you not found? You just mute error and broke serialization for some classes. We can catch this only on release build, but we doesn't have such tests.

assembly/dependencies-apache-ignite.xml
nva: Removed shading of commons-logging

config/ignite-log4j.xml
nva: exclude `com.fasterxml.jackson.databind.ext` package from renaming

nva (approved):

PR #1829, opened 2021-03-21
GG-32272 Fixed race condition and ClassNotFoundException in SFTP snapshots

SammyVimes:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -217,7 +217,9 @@
     DISTRIBUTED_ROLLING_UPGRADE_MODE_V2(58),
 
     /** Chains of snapshot operations. */
-    SNAPSHOT_OPERATIONS_CHAINING(59);
+    SNAPSHOT_OPERATIONS_CHAINING(59),
+
+    SNAPSHOT_SFTP_UPLOAD_V2(60);
```
ibessonov, line 222: I see no comment here.

ibessonov (approved):

PR #1843, opened 2021-03-25
GG-32193 Fixed an issue where it's possible to put a compressed binary object into the foreign cache that leads to corrupt B+Tree

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
```diff
@@ -1900,10 +1900,12 @@ private boolean isIncrementalDrEnabled(GridCacheContext cctx) {
...
                 CacheObjectContext coCtx = cctx.cacheObjectContext();
 
-                key.prepareForCache(coCtx, coCtx.compressKeys());
+                key = key.prepareForCache(coCtx, coCtx.compressKeys());
 
                 if (val != null)
-                    val.prepareForCache(coCtx, true);
+                    val = val.prepareForCache(coCtx, true);
+
+                dataRow = makeDataRow(key, val, ver, expireTime, cacheId);
```
sk0x50, line 1908: Well, it looks like we always create a new `dataRow` even though `key` and `val` are not changed after `prepareForCache`. Is it OK?

modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryObjectImpl.java
```diff
@@ -230,39 +230,43 @@ public BinaryObjectImpl(BinaryContext ctx, byte[] arr, int start) {
...
+            assert F.eq(this.compressionStrategy, entryCompressionStrategy) :
+                "Re-prepare compressed binary object form for a different cache";
 
-        if (entryCompressionCfg == null) {
-            compressedArr = UNCOMPRESSED;
+            return this;
+        }
 
-            return;
+        if (entryCompressionStrategy == null) {
```
sk0x50, line 249: Please remove unnecessary brackets

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
alamar: Why are they not changed? We reassign both key and val, and prepareForCache() may return a new object.

alamar: canUpdateOldRow() guarantees that compression is off.

sk0x50: Well, do we really need to create a new data row in case of entry compression is disabled?

PR #1865, opened 2021-04-03
GG-32853 Cluster prints Coordinator changed messages with current coordinator as daemon when the last one is oldest

makedonskaya:


modules/core/src/test/java/org/apache/ignite/internal/GridDiscoveryManagerChangeCoordinatorTest.java
```diff
@@ -67,23 +66,35 @@
...
+        lsnr.reset();
+
+        LogListener lsnr2 = LogListener.matches(CRD_CHANGE_MSG).build();
 
         listeningLog.registerListener(lsnr);
 
         stopGrid("client");
 
-        assertFalse(lsnr.check());
+        assertFalse(lsnr2.check());
```
denis-chudov, line 97: lsnr2 isn't registered anywhere, do you really need it? 
You can reuse lsnr, as it was reset.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
```diff
@@ -1598,10 +1598,10 @@ private void topologySnapshotMessage(
 
         clo.apply(summary.toString());
 
-        ClusterNode currCrd = discoCache.coordinator();
+        ClusterNode currCrd = discoCache.oldestServerNode();
```
sergey-chugunov-1985, line 1601: Should we use `oldestAliveServerNode` instead of `oldestServerNode` here? Not sure, just asking.

sergey-chugunov-1985 (commented):
I left a minor question, if everything OK with it, patch looks good to me.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/GridDiscoveryManager.java
makedonskaya: Fixed.

PR #1885, opened 2021-04-16
GG-33087 fix NPE in control.sh if there is illegal parameter after --cache indexes_force_rebuild

aries-3:


aealeksandrov (approved):
LGTM

PR #1887, opened 2021-04-16
GG-32845

vldpyatkov:


sergey-chugunov-1985 (changes requested):
I left several minor comments, please review them.

PR #1894, opened 2021-04-20
GG-33112 Implement metrics for in-memory index pages 

sashapolo:
https://ggsystems.atlassian.net/browse/GG-33112


modules/indexing/src/test/java/org/apache/ignite/internal/metric/IndexPagesMetricsPageDisplacementTest.java
```diff
@@ -0,0 +1,189 @@
...
+
+            idxPagesOnDisk = getIdxPagesOnDisk(grpId).size();
+            idxPagesInMemory = idxPageCounter.countIdxPagesInMemory(grpId);
+
+            assertThat(pageMetrics.indexPages().value(), is(idxPagesInMemory));
+
+        } while (idxPagesOnDisk <= idxPagesInMemory);
+
+        // load pages back into memory and check that the metric value has increased
+        touchIdxPages(grpId);
```
ibessonov, line 136: Is this ok that you don't release these pages?

modules/indexing/src/test/java/org/apache/ignite/internal/metric/IndexPagesMetricsPageDisplacementTest.java
```diff
@@ -0,0 +1,189 @@
...
+
+        // Page Store contains a one-page header
+        long numPages = pageStore.size() / pageStore.getPageSize() - 1;
+
+        for (int i = 0; i < numPages; i++) {
+            long pageId = PageIdUtils.pageId(PageIdAllocator.INDEX_PARTITION, (byte)0, i);
+
+            try {
+                pageStore.read(pageId, buf, false);
+            } catch (IgniteDataIntegrityViolationException ignored) {
```
ibessonov, line 178: Is this the exception that you don't have locally?

sashapolo: yes

sashapolo: I forgot to add a comment here, will do

modules/indexing/src/test/java/org/apache/ignite/internal/metric/IndexPagesMetricsPageDisplacementTest.java
sashapolo: I have a comment inside the `touchIdxPages` method: `// acquire, but not release all pages, so that they don't get displaced back to the storage`. Or do you think my reasoning is wrong?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
```diff
@@ -425,23 +519,73 @@ public void resetDirtyPages() {
...
+    /**
+     * Returns memory page metrics associated with the given cache group.
+     */
+    public PageMetrics cacheGrpPageMetrics(int cacheGrpId) {
+        PageMetrics pageMetrics = cacheGrpMetrics.get(cacheGrpId);
+
+        if (pageMetrics != null)
+            return pageMetrics;
+
+        synchronized (cacheGrpMetricsLock) {
```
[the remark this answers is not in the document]
sashapolo: Do you think it is necessary here? I thought that the write performance is not critical here, while using a synchronized block instead of CAS will result in less complicated code.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
```diff
@@ -187,16 +249,40 @@ public DataRegionMetricsImpl(DataRegionConfiguration memPlcCfg,
...
+                @Override public void increment() {
+                    add(1);
+                }
+
+                @Override public void add(long x) {
+                    if (metricsEnabled && x > 0)
+                        allocRate.add(x);
+                }
+
+                @Override public void decrement() {
```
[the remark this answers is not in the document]
sashapolo: These callbacks are for the page allocation rate metrics, so I think it doesn't make a lot of sense to take de-allocation into account. Also, this correlates with the previous implementation, which also discarded the negative values (see `updateAllocRate` method)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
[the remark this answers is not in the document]
ibessonov: Ok, makes sense.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/DataRegionMetricsImpl.java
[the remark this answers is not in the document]
ibessonov: Thanks for the explanation.

ibessonov (approved):

PR #1896, opened 2021-04-21
GG-33106

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/internal/binary/builder/BinaryBuilderReader.java
```diff
@@ -516,15 +516,26 @@ public Object parseValue() {
...
-                return res;
+                try {
+                    Object res = parseValue();
+
+                    if (res instanceof BinaryObjectBuilderImpl)
+                        objMap.put(objStart, (BinaryObjectBuilderImpl)res);
+
+                    return res;
+                }
+                finally {
```
ygerzhedovich, line 536: I'm not sure that we should save 'pos' in Exception case, maybe enough restore 'pos' before return statement?

modules/core/src/test/java/org/apache/ignite/internal/binary/BinaryMarshallerSelfTest.java
```diff
@@ -1031,6 +1035,56 @@ public void testWriteReplaceInheritable() throws Exception {
...
+                    colHnd = bob.getField(flds.get2());
+                }
+                else {
+                    colHnd = bob.getField(flds.get2());
+                    col = bob.getField(flds.get1());
+                }
+
+                // Must be assertSame but now BinaryObjectBuilder doesn't support handle to collection.
+                // Now we check only that BinaryObjectBuilder#getField doesn't crash and returns valid collection.
+                assertEquals("Check: " + flds, col, colHnd);
```
korlov42, line 1081: worth to print out iteration number in case of error

korlov42 (approved):

PR #1898, opened 2021-04-21
GG-30691 Fix access right issues in computation of system metrics.

dmekhanikov:


modules/core/src/test/java/org/apache/ignite/internal/metric/SystemMetricsTest.java
```diff
@@ -0,0 +1,65 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
alievmirza, line 2: 2021

alievmirza (changes requested):

PR #1903, opened 2021-04-26
GG-29004 Utils IgniteWalConverter add ability to skip reading errors

a-polyakov:


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/walconverter/StandaloneWalRecordsIteratorIgnoreError.java
```diff
@@ -0,0 +1,115 @@
...
+            FileWALPointer actualFilePtr = new FileWALPointer(hnd.idx(), (int)hnd.in().position(), 0);
+
+            try {
+                WALRecord rec = hnd.ser().readRecord(hnd.in(), actualFilePtr);
+
+                actualFilePtr.length(rec.size());
+
+                result = new IgniteBiTuple<>(actualFilePtr, postProcessRecord(rec));
+            }
+            catch (SegmentEofException eof) {
```
vldpyatkov, line 81: Looks like you can to catch both:
SegmentEofException | EOFException err

vldpyatkov: Also, need to add a message:
log.error("Critical exception has happened during WAL was scanned", err)

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/walconverter/StandaloneWalRecordsIteratorIgnoreError.java
```diff
@@ -0,0 +1,115 @@
...
+
+                    in.seek(actualFilePtr.fileOffset());
+
+                    final int recordType = in.readUnsignedByte() - 1;
+
+                    final long idx = in.readLong();
+
+                    final int offset = in.readInt();
+
+                    final int len = in.readInt();
```
vldpyatkov, line 101: Do not use final everywhere.

PR #1911, opened 2021-05-05
GG-28042 Cluster metrics BusyTimePercentage, IdleTimePercentage, UpTime was calculated incorrectly

a-polyakov:
…me was calculated incorrectly

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
+++ b/modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
@@ -347,7 +347,7 @@ public interface ClusterMetrics {
      * Gets percentage of time this node is busy executing jobs vs. idling.
      *
      * @return Percentage of time this node is busy (value is less than
-     *      or equal to {@code 1} and greater than or equal to {@code 0})
+     *      or equal to {@code 100} and greater than or equal to {@code 0})
      */
     public float getBusyTimePercentage();
 
@@ -355,7 +355,7 @@ public interface ClusterMetrics {
      * Gets percentage of time this node is idling vs. executing jobs.
      *
      * @return Percentage of time this node is idle (value is less than
-     *      or equal to {@code 1} and greater than or equal to {@code 0})
+     *      or equal to {@code 100} and greater than or equal to {@code 0})
      */
     public float getIdleTimePercentage();
 
--- a/modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
@@ -112,7 +112,7 @@ public class ClusterMetricsMXBeanImpl implements ClusterMetricsMXBean {
 
     /** {@inheritDoc} */
     @Override public float getBusyTimePercentage() {
-        return metrics().getBusyTimePercentage() * 100;
+        return metrics().getBusyTimePercentage();
     }
 
     /** {@inheritDoc} */
@@ -187,7 +187,7 @@ public class ClusterMetricsMXBeanImpl implements ClusterMetricsMXBean {
 
     /** {@inheritDoc} */
     @Override public float getIdleTimePercentage() {
-        return metrics().getIdleTimePercentage() * 100;
+        return metrics().getIdleTimePercentage();
     }
 
     /** {@inheritDoc} */
--- a/modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
@@ -398,7 +398,7 @@ public class ClusterMetricsSnapshot implements ClusterMetrics {
 
             nonHeapInit += m.getNonHeapMemoryInitialized();
 
-            upTime = max(upTime, m.getUpTime());
+            upTime += m.getUpTime();
 
             lastDataVer = max(lastDataVer, m.getLastDataVersion());
 
@@ -823,12 +823,12 @@ public class ClusterMetricsSnapshot implements ClusterMetrics {
 
     /** {@inheritDoc} */
     @Override public float getBusyTimePercentage() {
-        return 1 - getIdleTimePercentage();
+        return Math.max(100 - getIdleTimePercentage(), 0);
     }
 
     /** {@inheritDoc} */
     @Override public float getIdleTimePercentage() {
-        return getTotalIdleTime() / (float)getUpTime();
+        return Math.min((float)100. * getTotalIdleTime() / getUpTime(), (float)100.);
     }
 
     /** {@inheritDoc} */
--- a/modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
@@ -379,12 +379,12 @@ public class ClusterMetricsImpl implements ClusterMetrics {
 
     /** {@inheritDoc} */
     @Override public float getBusyTimePercentage() {
-        return 1 - getIdleTimePercentage();
+        return Math.max(100 - getIdleTimePercentage(), 0);
     }
 
     /** {@inheritDoc} */
     @Override public float getIdleTimePercentage() {
-        return getTotalIdleTime() / (float)getUpTime();
+        return Math.min((float)100 * getTotalIdleTime() / getUpTime(), (float)100.);
     }
 
     /** {@inheritDoc} */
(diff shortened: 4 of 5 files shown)


modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
```diff
@@ -347,15 +347,15 @@
      * Gets percentage of time this node is busy executing jobs vs. idling.
      *
      * @return Percentage of time this node is busy (value is less than
-     *      or equal to {@code 1} and greater than or equal to {@code 0})
+     *      or equal to {@code 100} and greater than or equal to {@code 0})
```
denis-chudov, line 350: You change the public contract with these changes. Let's leave this as is.

modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
```diff
@@ -347,15 +347,15 @@
...
      */
     public float getBusyTimePercentage();
 
     /**
      * Gets percentage of time this node is idling vs. executing jobs.
      *
      * @return Percentage of time this node is idle (value is less than
-     *      or equal to {@code 1} and greater than or equal to {@code 0})
+     *      or equal to {@code 100} and greater than or equal to {@code 0})
```
denis-chudov, line 358: You change the public contract with these changes. Let's leave this as is.

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
```diff
@@ -398,7 +398,7 @@ public ClusterMetricsSnapshot(ClusterGroup p) {
 
             nonHeapInit += m.getNonHeapMemoryInitialized();
 
-            upTime = max(upTime, m.getUpTime());
+            upTime += m.getUpTime();
```
denis-chudov, line 401: What about `max(0, m.getUpTime())` ?

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
```diff
@@ -823,12 +823,12 @@ public void setCurrentIdleTime(long curIdleTime) {
 
     /** {@inheritDoc} */
     @Override public float getBusyTimePercentage() {
-        return 1 - getIdleTimePercentage();
+        return Math.max(100 - getIdleTimePercentage(), 0);
```
denis-chudov, line 826: Let's replace 100 with 1 here to leave public contract unchanged.

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
```diff
@@ -823,12 +823,12 @@ public void setCurrentIdleTime(long curIdleTime) {
...
     }
 
     /** {@inheritDoc} */
     @Override public float getIdleTimePercentage() {
-        return getTotalIdleTime() / (float)getUpTime();
+        return Math.min((float)100. * getTotalIdleTime() / getUpTime(), (float)100.);
```
denis-chudov, line 831: Let's remove a multiplier here to leave public contract unchanged.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
```diff
@@ -379,12 +379,12 @@ public ClusterMetricsImpl(GridKernalContext ctx, long nodeStartTime) {
 
     /** {@inheritDoc} */
     @Override public float getBusyTimePercentage() {
-        return 1 - getIdleTimePercentage();
+        return Math.max(100 - getIdleTimePercentage(), 0);
```
denis-chudov, line 382: Let's replace 100 with 1 here to leave public contract unchanged.

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
```diff
@@ -379,12 +379,12 @@ public ClusterMetricsImpl(GridKernalContext ctx, long nodeStartTime) {
...
     }
 
     /** {@inheritDoc} */
     @Override public float getIdleTimePercentage() {
-        return getTotalIdleTime() / (float)getUpTime();
+        return Math.min((float)100 * getTotalIdleTime() / getUpTime(), (float)100.);
```
denis-chudov, line 387: Let's remove a multiplier here to leave public contract unchanged.

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
```diff
@@ -112,7 +112,7 @@ private ClusterMetrics metrics() {
 
     /** {@inheritDoc} */
     @Override public float getBusyTimePercentage() {
-        return metrics().getBusyTimePercentage() * 100;
+        return metrics().getBusyTimePercentage();
     }
 
     /** {@inheritDoc} */
```
denis-chudov, line 118: Did you think about including boundaries of this value in mx bean description in `ClusterMetricsMXBean`?

modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
a-polyakov: It is not logical. Method name percentage of use, but to output the share

modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
a-polyakov: It is not logical. Method name percentage of use, but to output the share

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
a-polyakov: The logic of cluster metrics is considered as the sum of the nodes, and in this case it was the maximum among the nodes, which breaks all further calculation. actually, this is the correction for the ticket, everything else is added to protect and edit inconsistencies in the name and content.

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
a-polyakov: Again, the method is called "Percentage"

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
a-polyakov: Again, the method is called " Percentage"

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
a-polyakov: Again, the method is called " Percentage"

modules/core/src/main/java/org/apache/ignite/internal/managers/discovery/ClusterMetricsImpl.java
a-polyakov: Again, the method is called " Percentage"

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsMXBeanImpl.java
a-polyakov: based on the name of the method, yes

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
denis-chudov: It seems that m.getUpTime() can be `-1` in some cases (if not initialized) but if you sure it can't - let's leave it as is.

modules/core/src/main/java/org/apache/ignite/cluster/ClusterMetrics.java
denis-chudov: I agree, but some users can already have monitoring system configured for value range [0;1], and we can't just break this in minor fix which even doesn't touch this directly. In such cases we need to continue living with it - and create a ticket to fix this in Ignite 3.

a-polyakov: Nobody used the metrics, and when they tried they noticed an error, in fact, this is why a ticket was opened

modules/core/src/main/java/org/apache/ignite/internal/ClusterMetricsSnapshot.java
a-polyakov: UpTime solve in ClusterMetricsImpl 
MetricRegistry mreg = ctx.metric().registry(SYS_METRICS);
upTime = mreg.findMetric(UP_TIME);
... 
in GridMetricManager
sysreg.register(UP_TIME, rt::getUptime, null);
...
java.lang.management.RuntimeMXBean#getUptime
sun.management.VMManagement#getUptime
...
Finally
Returns the uptime of the Java virtual machine in milliseconds.

PR #1915, opened 2021-05-07
GG-33148 Fix local discovery and port binding on z/OS.

slukyano:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2223,6 +2226,16 @@ public static synchronized boolean isLocalHostChanged() throws IOException {
...
+                try {
+                    InetAddress localHost = InetAddress.getByName(LOCALHOST);
+
+                    if (!locAddrs.contains(localHost))
+                        locAddrs.add(localHost);
+                }
+                catch (Exception ex) {
+                    // Ignore.
+                }
+
```
slukyano, line 2238: What's the purpose of this code? What is the scenario which doesn't work without it?

config/apache-ignite-zos/default-config.xml
```diff
@@ -25,17 +25,13 @@
...
         <property name="discoverySpi">
             <bean class="org.apache.ignite.spi.discovery.tcp.TcpDiscoverySpi">
-                <property name="localAddress" value="127.0.0.1"/>
-
                 <property name="ipFinder">
                     <bean class="org.apache.ignite.spi.discovery.tcp.ipfinder.vm.TcpDiscoveryVmIpFinder">
                         <property name="addresses">
                             <list>
-                                <value>127.0.0.1:47500..47510</value>
+                                <value>localhost:47500..47510</value>
```
slukyano, line 34: if you want to bind to all hosts, shouldn't this be 0.0.0.0?

alamar: This is not binding address. It should point to valid host/port of a node to discover (including itself).

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
alamar: Let me add a comment about that.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -2223,6 +2226,19 @@ public static synchronized boolean isLocalHostChanged() throws IOException {
                     }
                 }
 
+                /* On z/OS NetworkInterface will expose IPv6 ::1 and IPv4 external address, but not 127.0.0.1
+                    "localhost" resolves to 127.0.0.1 so we need to add it to binding addresses list so
+                    that discovery and clients can connect to "localhost" ports. */
```
slukyano, line 2231: I'd reword the comment like this based on our conversation:

```suggestion
                /* On z/OS NetworkInterface::getNetworkInterfaces returns external interfaces 
                    and IPv6 loopback but not IPv4 loopback.
                    Trying to additionally resolve "localhost" to find the IPv4 loopback address. */
```

PR #1930, opened 2021-05-18
GG-33217 Fixed deadlock in GridCacheDataStore initialization.

ibessonov:


sashapolo (approved):
Please leave a comment in Jira explaining what is happening here

PR #1940, opened 2021-05-25
gg-33259: calculate total table row count from overridden values.

Berkof:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/stat/IgniteStatisticsHelper.java
```diff
@@ -149,6 +149,18 @@ public static ObjectStatisticsImpl aggregateLocalStatistics(
             colStats.put(col.getName(), stat);
         }
 
+        long overridedRowCnt = -1;
+
+        for (StatisticsColumnConfiguration ccfg : cfg.columns().values()) {
+            if (ccfg.overrides() != null && ccfg.overrides().total() != null) {
+                Long colRowCnt = ccfg.overrides().total();
+
+                overridedRowCnt = (overridedRowCnt > colRowCnt) ? overridedRowCnt : colRowCnt;
```
ygerzhedovich, line 158: Why not Math.max(overridedRowCnt, cloRowCnt)?

Berkof: To allow user override this value to lower it.

ygerzhedovich: I didn't catch your comment. I mean the construction could be amended to Math.max.

Berkof: Ok, if I did it - how user can lower rowCount to make it less than actual rows count? Or you trying to say that we'll never need it?

PR #1963, opened 2021-06-10
GG-32178

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -751,6 +772,11 @@ public int emptyDataPages() {
         return initRecycledPage0(pageId, flag, initIO);
     }
 
+    /** {@inheritDoc} */
+    @Override protected int bucketsCnt() {
```
sergey-chugunov-1985, line 776: We don't need this method for two reasons: first, it is used in `pageIds` method which I believe should be eliminated and second, there is already a field `buckets` in `PagesList` class with exactly this semantics.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/PagesList.java
```diff
@@ -1995,6 +2002,138 @@ public long metaPageId() {
         return metaPageId;
     }
 
+    /**
+     * Collects the pages that list consists of.
+     *
+     * @param pageIds Initial list of pages ids, that will be added to result collection.
+     * @return Collection of pages that list consists of.
+     */
+    protected Collection<Long> pageIds(long... pageIds) {
```
sergey-chugunov-1985, line 2011: This method seems to bring very little value as it dumps pages that are fine (not corrupted) creating a lot of noise for further investigation of the issue but not making it easier.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/PagesPossibleCorruptionDiagnosticTest.java
```diff
@@ -100,9 +112,20 @@
         super.afterTest();
     }
 
+    private FilePageStore filePageStore(IgniteEx ignite, int partId) throws IgniteCheckedException {
```
sergey-chugunov-1985, line 115: No javadoc on this method, please fix.

sergey-chugunov-1985 (changes requested):

PR #1975, opened 2021-06-18
GG-32638

denis-chudov:


modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -1877,6 +1883,18 @@ private boolean inverseTcpConnectionFeatureIsSupported(ClusterNode node) {
         return new GridIoMessage(plc, topic, topicOrd, msg, ordered, timeout, skipOnTimeout);
     }
 
+    /**
+     * @return Version of security processor feature supported by cluster.
+     */
+    private IgniteFeatures currentSecurityProcSupport() {
+        if (allNodesSupports(ctx, ctx.discovery().allNodes(), IGNITE_SECURITY_PROCESSOR_V2))
```
sergey-chugunov-1985, line 1890: Lets eliminate double call of `ctx.discovery().allNodes()` method.

sergey-chugunov-1985 (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
denis-chudov: fixed

PR #1978, opened 2021-06-21
GG-33153 Introduce "-keep=N" switch for snapshot utility delete operation.

SammyVimes:


modules/core/src/main/java/org/apache/ignite/internal/visor/VisorDataTransferObject.java
```diff
@@ -55,6 +55,9 @@
     /** Version 5. */
     protected static final byte V5 = 5;
 
+    /** Version 5. */
```
tkalkirill, line 58: Version 6.

PR #1987, opened 2021-06-28
GG-32940 [IGNITE-14424] Fix cache group re-encryption after cluster secondary activation. (#9036)

alievmirza:


modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -1110,6 +1110,18 @@ private void releaseWalKeys(long segmentIdx) {
...
     }
 
+    /** {@inheritDoc} */
+    @Override public void onDoneAfterTopologyUnlock(GridDhtPartitionsExchangeFuture fut) {
+        if (!fut.isFailed() && (fut.activateCluster() || fut.localJoinExchange())) {
+            try {
+                startReencryption(reencryptGroups.keySet());
+            }
+            catch (IgniteCheckedException e) {
+                log.error("Unable to start reencryption", e);
```
sk0x50, line 1120: To be honest, it seems odd to me, just logging an exception. I think it would be nice to propagate this exception to the FailureHandler (except NodeStoppingException, perhaps).

alievmirza: If we look at the method `startReencryption` we can see the following code
```
            fut.listen(f -> {
                if (f.isCancelled() || f.error() != null) {
                    log.warning("Reencryption " +
                        (f.isCancelled() ? "cancelled" : "failed") + " [grp=" + grpId + "]", f.error());

                    return;
                }
```
As far as I can understand, the contract of reencryption allows to fail the process and we don't see any propagation to FailureHandler here, so I would say that this is unnecessary to call failure handler above, WDYT?

sk0x50: In my understanding, cache re-encryption can be canceled at any time via control-utility. So, it is ok to check the re-encryption future `f.isCancelled()`, but I still do not think that logging a warning message for a failed future is enough.

PR #2000, opened 2021-07-08
GG-33456 Reset connection may result a one second pause on processing messages by the NIO server

vmalin:


modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -2510,20 +2511,10 @@ private void processSelectedKeysOptimized(SelectionKey[] keys) throws ClosedByIn
                     throw e;
                 }
                 catch (Exception | Error e) { // TODO IGNITE-2659.
```
vldpyatkov, line 2513: I think, if you would catch Throwable, it looks more pretty.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -2577,21 +2568,33 @@ private void processSelectedKeys(Set<SelectionKey> keys) throws ClosedByInterrup
...
-                    else if (log.isDebugEnabled())
-                        log.debug("Failed to process selector key [ses=" + ses + ", err=" + e + ']');
+        /**
+         * Processes errors occured while processing selector keys.
+         *
+         * @param e Exception or error which occured.
+         * @param attach GridNioKeyAttachment.
+         */
+        private void procesKeysSelectionError(Throwable e, GridNioKeyAttachment attach) {
+            if (!X.hasCause(e, Exception.class)) {
```
vldpyatkov, line 2583: If we think here only about Assertion or other Errors, change the condition to _X.hasCause(e, Error.class)_.
But it only my view.

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
vmalin: I agree it's more correct to change to Throwable. However, catching Error explicitly shows that this code contains another error to be fixed. I would like to stay it as is. Also, IGNITE-2659 is added to our team's backlog as [IGN-4019](https://ggsystems.atlassian.net/browse/IGN-4019)

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -2510,20 +2511,10 @@ private void processSelectedKeysOptimized(SelectionKey[] keys) throws ClosedByIn
                     throw e;
                 }
                 catch (Exception | Error e) { // TODO IGNITE-2659.
-                    try {
-                        U.sleep(1000);
-                    }
-                    catch (IgniteInterruptedCheckedException ignore) {
-                        // No-op.
-                    }
+                    procesKeysSelectionError(e, attach);
```
ibessonov, line 2514: Typo in the word "process"

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -2577,21 +2568,33 @@ private void processSelectedKeys(Set<SelectionKey> keys) throws ClosedByInterrup
...
-                    else if (log.isDebugEnabled())
-                        log.debug("Failed to process selector key [ses=" + ses + ", err=" + e + ']');
+        /**
+         * Processes errors occured while processing selector keys.
+         *
+         * @param e Exception or error which occured.
+         * @param attach GridNioKeyAttachment.
+         */
+        private void procesKeysSelectionError(Throwable e, GridNioKeyAttachment attach) {
+            if (X.hasCause(e, Error.class)) {
```
ibessonov, line 2583: Can you please add a comment about this condition and sleep? I don't understand it and issues descriptions tells me nothing as well

vmalin: As I understand, this sleep was added by Alexey Goncharuk as a workaround for IGN-4019. But it shouldn't affect exceptions: holding the thread for 1 second broke other connections. All this should go away when IGNITE-2659 is fixed.

ibessonov (approved):

PR #2008, opened 2021-07-16
GG-33321 catch all reading-related exceptions in org.apache.ignite.internal.pr…

denis-chudov:
…ocessors.cache.persistence.checkpoint.CheckpointMarkersStorage#readPointer

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointMarkersStorage.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointMarkersStorage.java
@@ -304,7 +304,7 @@ public class CheckpointMarkersStorage {
 
             return new FileWALPointer(buf.getLong(), buf.getInt(), buf.getInt());
         }
-        catch (IOException e) {
+        catch (Exception e) {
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointMarkersStorage.java
```diff
@@ -304,7 +304,7 @@ private WALPointer readPointer(File cpMarkerFile, ByteBuffer buf) throws IgniteC
 
             return new FileWALPointer(buf.getLong(), buf.getInt(), buf.getInt());
         }
-        catch (IOException e) {
+        catch (Exception e) {
```
sergey-chugunov-1985, line 307: What about writing a test for the issue?

sergey-chugunov-1985 (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointMarkersStorage.java
denis-chudov: Done.

PR #2019, opened 2021-07-23
GG-33520 [8.7-next]-[GG-33519] Assertion error on join with empty LFS.

vmalin:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheAffinitySharedManager.java
```diff
@@ -2927,7 +2927,7 @@ private CacheGroupNoAffOrFilteredHolder createHolder(
      * @param log Logger to print to.
      */
     public void printWaitInfo(IgniteLogger log) {
-        if (waitInfo == null || waitInfo.empty() || !log.isInfoEnabled())
+        if (waitInfo == null || waitInfo.assignments.isEmpty() || !log.isInfoEnabled())
```
sk0x50, line 2930: `waitInfo` is not a volatile variable, so you need to read this variable under the `mux`, I think.
Moreover, it can be released by `CacheAffinitySharedManager#onCacheGroupStopped`, for example.
So, I would suggest the following fix:
```
WaitRebalanceInfo info;

synchronized (mux) {
    info = waitInfo;
}

if (info == null || info.assignments.isEmpty() || !log.isInfoEnabled())
    return;

...
```

What do you think?

vmalin: Done

PR #2034, opened 2021-07-30
GG-33154 Set cluster ID from system property

wirtsleg:
https://ggsystems.atlassian.net/browse/GG-33154

Decided to keep this changed in the ticket branch.


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1731,6 +1737,34 @@ public static double getDouble(String name, double dflt) {
         return res;
     }
 
+    /**
+     * Gets either system property or environment variable with given name.
+     * The result is transformed to {@link UUID} using {@link UUID#fromString} method.
+     *
+     * @param name Name of the system property or environment variable.
+     * @param dflt Default value.
+     * @return Integer value of the system property or environment variable.
```
nva, line 1746:
```suggestion
     * @return UUID value of the system property or environment variable.
```

nva (approved):

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1731,6 +1737,34 @@ public static double getDouble(String name, double dflt) {
...
+        if (s == null)
+            return dflt;
+
+        UUID res;
+
+        try {
+            res = UUID.fromString(s);
+        }
+        catch (IllegalArgumentException ignore) {
+            res = dflt;
```
slukyano, line 1762: I'd avoid swalloing the exception here. At least a warning would be nice

wirtsleg: It is default behavioral for the whole class (`getInteger`, `getLong`, etc.). I don't think we should change it specifically for this method.

slukyano (approved):

PR #2035, opened 2021-08-02
GG-30606 Run checkpoint on a cluster

odiszapc:
Feature allows run checkpoint manually on a whole cluster (on all the servr nodes). API:

**CLI:**
control.sh --checkpoint

**Java:**
ignite.cluster().checkpoint();

**JMX:**
org.apache.ignite.mxbean.IgniteClusterMXBean#checkpoint()


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/checkpointing/subcommands/CheckpointForceCommand.java
```diff
@@ -0,0 +1,55 @@
...
+import org.apache.ignite.internal.commandline.checkpointing.CheckpointAbstractSubCommand;
+import org.apache.ignite.internal.commandline.checkpointing.CheckpointingForceResult;
+import org.apache.ignite.internal.commandline.checkpointing.CheckpointingForceTask;
+import org.apache.ignite.internal.commandline.checkpointing.CheckpointingSubCommandsList;
+import org.apache.ignite.internal.commandline.meta.subcommands.VoidDto;
+
+/**
+ *
+ */
+public class CheckpointForceCommand extends CheckpointAbstractSubCommand<VoidDto, CheckpointingForceResult> {
```
sk0x50, line 30: Please add meaningful Javadoc here and below.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
+/*
+ * Copyright 2021 GridGain Systems, Inc. and Contributors.
```
sk0x50, line 2: 2021 -> 2022

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
...
+import org.apache.ignite.lang.IgniteInClosure;
+import org.apache.ignite.resources.JobContextResource;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * Task for checkpointing operation
+ */
+@GridInternal
+public class CheckpointingForceTask extends VisorMultiNodeTask<VoidDto, CheckpointingForceResult, NodeCheckpointingResult> {
+
```
sk0x50, line 42: Please remove the empty line.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
...
+    /** */
+    private static final long serialVersionUID = 0L;
+
+    @Override protected VisorJob<VoidDto, NodeCheckpointingResult> job(VoidDto arg) {
```
sk0x50, line 46: Please add meaningful JavaDoc here and below.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
...
+
+                startTime = U.currentTimeMillis();
+                CheckpointProgress progress = ignite.context().cache().context().database().forceCheckpoint("force");
+                if (progress == null) {
+                    res.durationMillis(U.currentTimeMillis() - startTime);
+                    return res;
+                }
+
+                fut = progress.futureFor(CheckpointState.FINISHED);
+                if (fut.isDone())
```
sk0x50, line 100: In case of the `fut` is already done here, this implementation always returns `null` instead of the correct result.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
...
+
+                else {
+                    jobCtx.holdcc();
+
+                    fut.listen(new IgniteInClosure<IgniteInternalFuture>() {
+                        @Override public void apply(IgniteInternalFuture f) {
+                            res.durationMillis(U.currentTimeMillis() - startTime);
+
+                            if (f.isCancelled())
+                                res.error(f.error());
```
sk0x50, line 111: This exception will be overridden by `getFuture()` call.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
...
+
+                return null;
+            }
+            getFuture(fut, res);
+
+            return res;
+        }
+    }
+
+    private static Object getFuture(GridFutureAdapter fut, NodeCheckpointingResult res) {
```
sk0x50, line 126: The name of this method is not clear to me. Its name does not correlate with what this method actually does.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
```diff
@@ -0,0 +1,136 @@
...
+        try {
+            return fut.get();
+        }
+        catch (IgniteCheckedException e) {
+            res.error(e);
+        }
+
+        return null;
```
sk0x50, line 134: Do we need returned value at all?

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceResult.java
```diff
@@ -0,0 +1,82 @@
...
+import java.util.Map;
+import java.util.UUID;
+import org.apache.ignite.internal.dto.IgniteDataTransferObject;
+import org.apache.ignite.internal.processors.task.GridInternal;
+import org.apache.ignite.internal.util.typedef.internal.U;
+
+/**
+ *
+ */
+@GridInternal
```
sk0x50, line 32: This annotation makes sense for tasks only.

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointingForceTask.java
sk0x50: It would be nice to get an answer :)

modules/core/src/main/java/org/apache/ignite/internal/commandline/checkpointing/NodeCheckpointingResult.java
```diff
@@ -0,0 +1,114 @@
...
+import java.io.ObjectInput;
+import java.io.ObjectOutput;
+import java.util.UUID;
+import org.apache.ignite.internal.dto.IgniteDataTransferObject;
+import org.apache.ignite.internal.processors.task.GridInternal;
+
+/**
+ *
+ */
+@GridInternal
```
sk0x50, line 29: This annotation makes sense for tasks only.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointAbstractSubCommand.java
```diff
@@ -0,0 +1,97 @@
...
+ * Base cli command class for {@link CheckpointForceCommand }
+ *
+ * Launch checkpointing process on a cluster and return results
+ */
+public abstract class CheckpointAbstractSubCommand<
+    MetadataArgsDto extends IgniteDataTransferObject,
+    MetadataResultDto extends IgniteDataTransferObject>
+    extends AbstractCommand<MetadataArgsDto> {
+
+    /**
```
sk0x50, line 45: Please, add meaningful Javadoc for methods and fields in accordance with the code style:
```
Every method, field or initializer public, private or protected in top-level, inner or anonymous type should have at least minimal Javadoc comments including description and description of parameters using @param, @return and @throws Javadoc tags, where applicable.
```

And use one-line comments where applicable:
```
    /** Command arguments. */
    private MetadataArgs args;
```

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/checkpointing/CheckpointCommand.java
```diff
@@ -0,0 +1,73 @@
...
+        );
+    }
+
+    /** {@inheritDoc} */
+    @Override public String name() {
+        return CHECKPOINT.toCommandName();
+    }
+
+    @Override public void parseArguments(CommandArgIterator argIter) {
+        if (!argIter.hasNextArg()) {
```
sk0x50, line 48: It seems that this code is not quite correct. In case of using auto-confirmation syntax: 'control.sh --checkpoint --yes`  this code throws an exception(Expected checkpointing action).

Moreover, I don't think that we need `CheckpointForceCommand` at all.

modules/core/src/test/java/org/apache/ignite/internal/GridCheckpointTest.java
```diff
@@ -0,0 +1,158 @@
...
+                    .setMaxSize(100 * 1024 * 1024)
+            );
+
+        cfg.setDataStorageConfiguration(dbCfg);
+
+        return cfg;
+    }
+
+    /** */
+    protected CacheConfiguration cacheConfiguration() {
```
sk0x50, line 94: Looks like this method is not used.

modules/core/src/test/java/org/apache/ignite/internal/GridCheckpointTest.java
```diff
@@ -0,0 +1,158 @@
...
+
+        final int numOfCheckpointsAfter = checkpointHist.checkpoints().size();
+        final CheckpointEntry lastCheckpointAfter = checkpointHist.lastCheckpoint();
+
+        assertTrue(numOfCheckpointsAfter > numOfCheckpointsBefore);
+        assertNotSame(lastCheckpointBefore, lastCheckpointAfter);
+    }
+
+    @Test
+    public void testForceCheckpointingJmx() throws Exception {
```
sk0x50, line 128: Do we need an additional test for a case when cluster does not contains any server node?

modules/control-utility/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerParsingTest.java
```diff
@@ -1141,6 +1141,7 @@ private boolean requireArgs(@Nullable CommandList cmd) {
             cmd == CommandList.WARM_UP ||
             cmd == CommandList.PROPERTY ||
             cmd == CommandList.METRIC ||
-            cmd == CommandList.DEFRAGMENTATION;
+            cmd == CommandList.DEFRAGMENTATION ||
+            cmd == CommandList.CHECKPOINT;
```
sk0x50, line 1145: Why is this needed? Checkpoint command does not require additional arguments.

PR #2038, opened 2021-08-03
GG-33584 Return the IGNITE_THRESHOLD_WAL_ARCHIVE_SIZE_PERCENTAGE and deprecate it

tkalkirill:


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1008,11 +1018,9 @@
 
     /**
      * Whenever read load balancing is enabled, that means 'get' requests will be distributed between primary and backup
-     * nodes if it is possible and {@link CacheConfiguration#readFromBackup} is {@code true}.
+     * nodes if it is possible and {@code CacheConfiguration#readFromBackup} is {@code true}.
```
ibessonov, line 1021: I suppose that this method doesn't exist, right?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
```diff
@@ -650,14 +652,31 @@ static void checkWalArchiveSizeConfiguration(
...
+                        "DataRegionConfiguration.minWalArchiveSize must be less than or equal to " +
+                            "DataRegionConfiguration.maxWalArchiveSize or equal to %d " +
+                            "(to be half of maxWalArchiveSize), current settings:" + U.nl() +
+                            "DataRegionConfiguration.minWalArchiveSize: %d bytes" + U.nl() +
+                            "DataRegionConfiguration.maxWalArchiveSize: %d bytes",
+                        HALF_MAX_WAL_ARCHIVE_SIZE, min, max));
+                }
+            }
+            else if (percentage != -1) {
+                log.warning(String.format(
```
ibessonov, line 669: Shouldn't we also validate that "percentage" is positive?

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
tkalkirill: This is a link to the field.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/IgniteCacheDatabaseSharedManager.java
tkalkirill: In this case, 0 and a negative value does not break the logic, since we will always try to delete segments until this value.

ibessonov (approved):

PR #2046, opened 2021-08-05
GG-33596 fix tests which were broken by GG-32372

RodionSmolnikovGG:


modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -1933,6 +1935,15 @@ public static void retryAssert(@Nullable IgniteLogger log, int retries, long ret
         return bytes;
     }
 
+    /**
+     * Clear file without deletion
+     *
+     * @param path to file
```
denis-chudov, line 1941: pls add dots to the ends of lines

PR #2049, opened 2021-08-12
GG-33285 [IGNITE-14805] Avoid unconditional Lucene index creation in …

ygerzhedovich:
…case of a String value type for a table


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/DistributedSqlConfiguration.java
```diff
@@ -227,4 +242,16 @@ public int defaultQueryTimeout() {
...
+
+    /** */
+    public boolean isDisableCreateLuceneIndexForStringValueType() {
+        Boolean ret = disableCreateLuceneIndexForStringValueType.get();
+
+        return ret != null && ret;
+    }
+
+    /** */
+    public GridFutureAdapter<?> createLuceneIndexForStringValueType(boolean create) throws IgniteCheckedException {
```
vladErmakov07, line 254: Looks misleading. I expect that we will enable the functionality (when calling with true value), but in fact we are disabling it.

AMashenkov (approved):

PR #2053, opened 2021-08-15
GG-33188 added timeout for transactions related to security contexts on system cache.

sk0x50:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
```diff
@@ -3225,6 +3225,18 @@ public CacheMetricsImpl metrics0() {
         return txs.txStartEx(ctx, concurrency, isolation, timeout, txSize).proxy();
     }
 
+    /** {@inheritDoc} */
+    @Override public GridNearTxLocal txStartEx(
```
ascherbakoff, line 3229: Introducing these methods doesn't make sense to me.
You can unwrap local tx using the following code: 

`GridNearTxLocal tx0 = ((TransactionProxyImpl)tx).tx();`

modules/control-utility/src/test/java/org/apache/ignite/util/GridCommandHandlerTest.java
```diff
@@ -1730,127 +1730,132 @@ public void testActiveTransactions() throws Exception {
         CommandHandler h = new CommandHandler();
 
         final VisorTxInfo[] toKill = {null};
+        IgniteInternalFuture<?> startFut = null;
```
ascherbakoff, line 1733: Why you have changed the test ? It looks correct to me and doesn't need changing.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheAdapter.java
sk0x50: In my opinion, this code does not look good enough. Moreover, it requires creating a new `TransactionProxyImpl` instance which is redundant I think.

modules/control-utility/src/test/java/org/apache/ignite/util/GridCommandHandlerTest.java
sk0x50: Unfortunately, the failure of this test can lead to the fact that the whole test suite will hang.

Let's consider the following scenario:
The test starts new transactions and one of them is "blocked" by the `unlockLatch`
     `IgniteInternalFuture<?> fut = startTransactions("testActiveTransactions", lockLatch, unlockLatch, true);`
This method uses the wrong method in order to wait for that latch `U.awaitQuiet(unlockLatch);`
So, if the test fails before this latch is released, then a thread from the pool `GridTestSafeThreadFactory#startedThreads` (which is used by `runMultiThreadedAsync`) has no chance to be completed and this is the cause of hanging the whole test suite.

 It seems to me, that test is absolutely correct and I should only change `U.awaitQuiet(unlockLatch);` to `unlockLatch.await()`.

PR #2060, opened 2021-08-20
GG-33679 [IGNITE-15078] Remove SimpleDateTimeFormat usages

sashapolo:
https://ggsystems.atlassian.net/browse/GG-33679

### DIFF
--- a/modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/ProgressPrinter.java
+++ b/modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/ProgressPrinter.java
@@ -16,9 +16,7 @@
 package org.apache.ignite.internal.commandline;
 
 import java.io.PrintStream;
-import java.text.SimpleDateFormat;
-import java.util.Date;
-import java.util.TimeZone;
+import java.util.concurrent.TimeUnit;
 import org.apache.ignite.internal.util.GridStringBuilder;
 
 /**
@@ -33,7 +31,7 @@ public class ProgressPrinter {
     private static final int MIN_PROGRESS_BAR_LENGTH = 20;
 
     /** */
-    private static final int MAX_CAPTION_LENTH = 40;
+    private static final int MAX_CAPTION_LENGTH = 40;
 
     /** */
     private final long total;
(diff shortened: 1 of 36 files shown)


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/ProgressPrinter.java
```diff
@@ -147,4 +140,15 @@ private void printProgress0(long curr, double currRatio) {
 
         printStream.print(progressBar);
     }
+
+    /**
+     * Formats the given duration in milliseconds in "HH:mm:ss" format.
+     */
+    private static String formatDuration(long durationMillis) {
```
ibessonov, line 147: I think you could have used IgniteUtils.SHORT_DATE_FMT, right? What was the reason

ibessonov (commented):
I believe some imports are in wrong order here, can you please check it? Otherwise, everything looks good.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/ProgressPrinter.java
sashapolo: `DateTimeFormatter` is not suitable for formatting durations, because it creates a string representation of a point in time. Same applies for the `SimpleDateFormat`, the original code did a strange things like:
```
long timeRunning = System.currentTimeMillis() - timeStarted;
...
timeFormat.format(new Date(timeRunning));
```
Or am I missing something?

ibessonov: Yes, it did exactly what you mentioned, but duration is indistinguishable from any other time instant. As far as I know, timestamps start 00:00:00 so old code was valid in that regard, it didn't use any dates.

sashapolo: yes, but strictly speaking, that's a hack. However, I may be too much of a purist here, and we can convert it to an `Instant`. What do you think?

ibessonov: I think you can leave it as it is right now, not a big deal. From the purist standpoint, your solution is better.

PR #2075, opened 2021-09-02
Gg 33616

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -114,14 +124,46 @@ public IgniteSecurityProcessor(GridKernalContext ctx, GridSecurityProcessor secP
...
-        SecurityContext res = node != null ? secCtxs.computeIfAbsent(subjId,
-            uuid -> nodeSecurityContext(marsh, U.resolveClassLoader(ctx.config()), node))
-            : secPrc.securityContext(subjId);
+        SecurityContext res = null;
+
+        if (!senderNodeId.equals(subjId)) {
+            ClusterNode senderNode = Optional.ofNullable(ctx.discovery().node(senderNodeId))
+                .orElseGet(() -> ctx.discovery().historicalNode(senderNodeId));
+
+            if (checkSenderNodeSubject && senderNode.isClient()) {
```
sk0x50, line 142: `senderNode` can be `null`

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -114,14 +124,46 @@ public IgniteSecurityProcessor(GridKernalContext ctx, GridSecurityProcessor secP
...
+                SecuritySubjectType type = node != null ? SecuritySubjectType.REMOTE_NODE : SecuritySubjectType.REMOTE_CLIENT;
+
+                log.warning("Switched to the 'deny all' policy because a client node tries to execute a request on behalf of another node " +
```
sk0x50, line 145: I think this warning should be logged if `forcibleNodeKillEnabled` is `false`

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
```diff
@@ -114,14 +124,46 @@ public IgniteSecurityProcessor(GridKernalContext ctx, GridSecurityProcessor secP
...
+                    "[senderNodeId=" + senderNodeId + ", subjId=" + subjId + ", type=" + type + ']');
+
+                if (forcibleNodeKillEnabled && senderNode != null && ctx.config().getCommunicationSpi() instanceof TcpCommunicationSpi) {
+                    TcpCommunicationSpi tcpCommSpi = (TcpCommunicationSpi)ctx.config().getCommunicationSpi();
+
+                    SecurityException ex = new SecurityException("Client node tried to execute a request on behalf of another node.");
+
+                    CommunicationTcpUtils.failNode(senderNode, tcpCommSpi.getSpiContext(), ex, log);
+
+                    throw ex;
```
sk0x50, line 155: It seems to me, this exception will trigger FailureHandler and that is not correct, I think

modules/core/src/test/java/org/apache/ignite/internal/TestRecordingCommunicationSpi.java
```diff
@@ -72,6 +72,27 @@
...
+     */
+    public TestRecordingCommunicationSpi() {
+        this(true);
+    }
+
+    /**
+     * @param unwrapOriginMessage True when a message will be unwrapped before provide to a predicate,
+     * false the message will be provided as an Io message.
+     */
+    public TestRecordingCommunicationSpi(boolean unwrapOriginMessage) {
```
sk0x50, line 92: I think that we do not need to modify `TestRecordingCommunicationSpi` itself. Using a custom `Spi` is more straightforward I think. Something like this:
`
    public static class MyCustomSpi extends TestRecordingCommunicationSpi {
        /** Security subject id. */
        private volatile UUID subjectId;

        /** {@inheritDoc} */
        @Override public void sendMessage(
            ClusterNode node,
            Message msg,
            IgniteInClosure<IgniteException> ackC
        ) throws IgniteSpiException {
            if (msg instanceof GridIoSecurityAwareMessage && subjectId != null) {
                GridIoSecurityAwareMessage ioSecMsg = (GridIoSecurityAwareMessage)msg;

                log.info("Msg is sending to node [msg=" + ioSecMsg.message().getClass().getSimpleName() +
                    ", node=" + node.consistentId() + ']');

                GridTestUtils.setFieldValue(msg, GridIoSecurityAwareMessage.class, "secSubjId", subjectId);
            }

            super.sendMessage(node, msg, ackC);
        }

        public void securitySubjectId(UUID subjectId) {
            this.subjectId = subjectId;
        }
    }
`

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
vldpyatkov: Fixed.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
vldpyatkov: I stay it as it was, because the logic is changed.
Now it always tries to execute a suspecting request with deny all policy, but when the forcible node kill property is enabled, it kills the node simultaneously.

modules/core/src/main/java/org/apache/ignite/internal/processors/security/IgniteSecurityProcessor.java
vldpyatkov: Fxed.

modules/core/src/test/java/org/apache/ignite/internal/TestRecordingCommunicationSpi.java
vldpyatkov: Fixed.

modules/core/src/test/java/org/apache/ignite/internal/TestRecordingCommunicationSpi.java
```diff
@@ -72,6 +72,27 @@
     /** */
     private volatile IgniteBiInClosure<ClusterNode, Message> c;
 
+    /**
+     * When the filed is true an message will be unwrapped before it will be passed to a predicate.
+     */
+    private final boolean unwrapOriginMessage;
```
sk0x50, line 78: Looks like, this change is not needed anymore.

PR #2076, opened 2021-09-02
GG-33731 Make GridGainFeatures deprecated.

Erixonich:


modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -44,6 +44,9 @@
 
 /**
  * Defines supported features and check its on other nodes.
+ *
+ * Note: for creating new feature and assigning ID, please update
```
vmalin, line 48: More articles to the God of articles. Note: for creating **a** new feature and assigning **an** ID, please update **the** page

Erixonich: done.

PR #2077, opened 2021-09-02
GG-33437 Add fast compression options to snapshot-utility.

Erixonich:


modules/core/src/test/java/org/apache/ignite/testframework/GridTestUtils.java
```diff
@@ -140,7 +140,7 @@
     public static final long DFLT_BUSYWAIT_SLEEP_INTERVAL = 200;
 
     /** */
-    public static final long DFLT_TEST_TIMEOUT = 5 * 60 * 1000;
+    public static final long DFLT_TEST_TIMEOUT = 5000 * 60 * 1000;
```
zstan, line 143: redefine timeout in certain test

Erixonich: It wrong code. Sorry, I committed it by mistake.

modules/core/src/main/java/org/apache/ignite/internal/IgniteFeatures.java
```diff
@@ -238,7 +238,10 @@
      * The feature allows authenticating a node remotely.
      * It is need when the one node which does not have an authentication required to authorize itself or another one.
      */
-    REMOTE_AUTH(61);
+    REMOTE_AUTH(61),
+
+    /** Snapshot operations with ZSTD, LZ4, SNAPPY algorithms. */
+    SNAPSHOT_COMPRESSION_MODERN_OPTION(63);
```
sergey-chugunov-1985, line 244: I don't like naming here. Lets replace `MODERN` with `EXTENDED`, and when we need to add even more options, we'll add the same feature but with prefix V2 - see how it was done for SFTP_UPLOAD feature.

sergey-chugunov-1985 (changes requested):

PR #2079, opened 2021-09-06
GG-33617 Added additional check on handshake phase

vldpyatkov:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/InboundConnectionHandler.java
```diff
@@ -77,6 +79,14 @@
      */
     private static final IgniteProductVersion VERSION_SINCE_CLIENT_COULD_WAIT_TO_CONNECT = IgniteProductVersion.fromString("2.1.4");
 
+    /**
+     * Enables additional check that sender of communication handshake message corresponds to the node id
+     * included in the message. In other words, the remote address of sender is equal to the already known address
+     * which corresponds to received node id in the message.
+     */
+    private final boolean checkCoomHandshakeSender =
```
ibessonov, line 87: Please fix the typo in the name, it should be "checkCommHandshakeSender"

ibessonov (approved):

PR #2089, opened 2021-09-11
GG-33831: fix parent.relativePath

slukyano:


vveider (approved):
Looks good! We already use the same at AI-3 / GG-9

PR #2092, opened 2021-09-15
GG-33827 Fix snapshot creation for statically configured caches

SammyVimes:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupContext.java
```diff
@@ -1093,9 +1093,10 @@ public boolean hasContinuousQueryCaches() {
     }
 
     /**
+     * @param cacheDescriptor Cache descriptor.
      * @throws IgniteCheckedException If failed.
      */
-    public void start() throws IgniteCheckedException {
+    public void start(DynamicCacheDescriptor cacheDescriptor) throws IgniteCheckedException {
```
ibessonov, line 1099: Why does cache group suddenly depends on single cache descriptor? What's happening here?

ibessonov: nevermind

ibessonov (approved):

ibessonov (approved):

PR #2098, opened 2021-09-23
GG-33835 Fix StackOverflowError in case if exception suppressed with itself.

Erixonich:


modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -464,37 +469,22 @@ private static void cloneField(Map<Integer, Integer> identityIdxs, List<Object>
...
+    public static boolean hasCause(
+        @Nullable Throwable t,
+        @Nullable String msg,
+        @Nullable Class<?>... cls
+    ) {
         if (t == null || F.isEmpty(cls))
             return false;
 
-        assert cls != null;
+        Predicate<Throwable> pred = thr -> Arrays.stream(cls)
```
tkalkirill, line 480: This method can be used in hot methods, it should be converted to methods with loops.

tkalkirill: Use indexed loops whenever possible.

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -464,37 +469,26 @@ private static void cloneField(Map<Integer, Integer> identityIdxs, List<Object>
...
+        @Nullable String msg,
+        @Nullable Class<?>... cls
+    ) {
         if (t == null || F.isEmpty(cls))
             return false;
 
-        assert cls != null;
-
-        for (Throwable th = t; th != null; th = th.getCause()) {
+        Predicate<Throwable> pred = thr -> {
```
tkalkirill, line 480: Use method instead predicate, for use in hot methods.

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
Erixonich: changed to raw loop. Thank you.

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
Erixonich: Are you trying to reduce `dynamic dispatch` byte-code instruction in this case?
We have to use a different `traverse` function in each case(functions cause() and hasCause() demand different predicates) - that yields copy-paste in algorithm.
I am not sure that a good thing without benchmarking because the creation throwable itself is a very expensive process (mostly capturing stack trace) and `dynamic dispatch` overhead is not so visible against it.

Erixonich: Alternatively, we can traverse the whole 'tree' of throwables and memorize all elements in some list, but we will lose short-circuit`ness of operation and more garbage will be produced.

Erixonich: I've fixed this possible issue by inlining the filter into the traversal function. I don't think we need to expand filters in future - so it looks ok in possibly hot code.

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+    /**
+     * Traverses `tree` of {@link Throwable} to find first cause/suppressed which meets test condition of predicate.
+     * Function is aware of possible circular references through tracking of tested objects.
+     *
+     * @param t Throwable - `root of the tree`.
+     * @param msg Possible throwable message.
+     * @param types Candidate types.
+     * @return First throwable meets test condition or {@code null} if none has matched.
+     */
+    @Nullable private static Throwable searchForCause(Throwable t, @Nullable String msg, Class<?>... types) {
```
tkalkirill, line 544: This method is not needed, you can directly call the` searchForCause(java.lang.Throwable, java.util.Set<java.lang.Throwable>, java.lang.String, java.lang.Class<?>...)`

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+                return found;
         }
 
         return null;
     }
 
+    /**
+     * Checks whether given throwable is assignable from any of candidate types and contains given non-null message.
+     *
+     * @param thr Object.
```
tkalkirill, line 595: Not object

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+     * @param msg Possible throwable message.
+     * @param candidateTypes Types to check.
```
tkalkirill, line 597: What types?

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+     * @return {@code True} if such type in found, else {@code False}.
```
tkalkirill, line 598: {@code true}
{@code false}

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+    @Nullable private static Throwable searchForCause(
+        Throwable t,
+        Set<Throwable> dejaVu,
+        @Nullable String msg,
+        Class<?>... types
+    ) {
+        if (isThrowableValid(t, msg, types))
+            return t;
+
+        if (dejaVu.contains(t))
```
tkalkirill, line 568: U can use `java.util.Set#add`

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
-            if (th.getCause() == th)
-                break;
+        return searchForCause(t, dejaVu, msg, types);
+    }
+
+    /**
+     * Traverses `tree` of {@link Throwable} to find first cause/suppressed is assignable from any of given types.
+     * Function is aware of possible circular references through tracking of tested objects.
+     *
+     * @param t Throwable - `root of the tree`.
```
tkalkirill, line 554: I guess this is not the root

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+     * @param dejaVu Set of throwable for tracking already tested objects.
+     * @param types Candidate types.
```
tkalkirill, line 556: What types?

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
```diff
@@ -53,6 +54,42 @@ public void testHasCause() {
...
+    public void testCauseDoesNotRaiseStackOverflowWhenCircularReferenceInSuppressed() {
+        NullPointerException npe = new NullPointerException();
+
+        IOException ioExc = new IOException(npe);
+
+        IgniteException exc = new IgniteException(ioExc);
+
+        ioExc.addSuppressed(exc);
+
+        X.cause(exc, NullPointerException.class);
```
tkalkirill, line 71: Use `assertEquals` or something like that

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
```diff
@@ -53,6 +54,42 @@ public void testHasCause() {
...
+        X.hasCause(exc, NullPointerException.class);
```
tkalkirill, line 72: Use `assertTrue` or something like that

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
```diff
@@ -53,6 +54,42 @@ public void testHasCause() {
...
+    public void testCauseDoesNotRaiseStackOverflowWhenCircularReferenceInCaused() {
+        NullPointerException npe = new NullPointerException();
+
+        IOException ioExc = new IOException(npe);
+
+        IgniteException exc = new IgniteException(ioExc);
+
+        npe.initCause(exc);
+
+        X.cause(exc, NullPointerException.class);
```
tkalkirill, line 89: Use `assertEquals` or something like that

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
```diff
@@ -53,6 +54,42 @@ public void testHasCause() {
...
+        X.hasCause(exc, NullPointerException.class);
```
tkalkirill, line 90: Use `assertTrue` or something like that

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
```diff
@@ -53,6 +54,42 @@ public void testHasCause() {
         assert gridEx.getCause(NumberFormatException.class) == null;
     }
 
+    /**
+     * Checks that using function with Throwable that contains circular reference in {@code getSuppressed()}
+     *     does not cause {@link StackOverflowError}.
+     */
+    @Test
+    public void testCauseDoesNotRaiseStackOverflowWhenCircularReferenceInSuppressed() {
```
tkalkirill, line 62: Add a test that did not find an exception

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +526,87 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
+     * Traverses `tree` of {@link Throwable} to find first cause/suppressed which meets test condition of predicate.
+     * Function is aware of possible circular references through tracking of tested objects.
+     *
+     * @param t Throwable - `root of the tree`.
+     * @param msg Possible throwable message.
+     * @param types Candidate types.
+     * @return First throwable meets test condition or {@code null} if none has matched.
+     */
+    @Nullable private static Throwable searchForCause(Throwable t, @Nullable String msg, Class<?>... types) {
+        Set<Throwable> dejaVu = Collections.newSetFromMap(new IdentityHashMap<>());
```
tkalkirill, line 545: Maybe use ThreadLocal WDYT? @ibessonov

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
Erixonich: ok. Inlined.

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
Erixonich: added asserts to existing tests for this case.

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
Erixonich: I don't think it is a good Idea. Almost the same gc pressure (we need to clean set before function start), but we get one level of indirection when using ThreadLocal. 

P.S. Same mechanism is used in jdk for printing throwable stacktrace. see java.lang.Throwable#printStackTrace(java.lang.Throwable.PrintStreamOrWriter)

modules/core/src/main/java/org/apache/ignite/internal/util/typedef/X.java
```diff
@@ -547,24 +528,73 @@ public static boolean hasSuppressed(@Nullable Throwable t, @Nullable Class<? ext
...
 
-            if (th.getCause() == th)
-                break;
+    /**
+     * Traverses `tree` of {@link Throwable} to find first cause/suppressed is assignable from any of given types.
+     * Function is aware of possible circular references through tracking of tested objects.
+     *
+     * @param t Throwable.
+     * @param dejaVu Set of throwable for tracking already tested objects.
+     * @param types Candidate types.
```
tkalkirill, line 545: Give more descriptions about the candidates.

modules/core/src/test/java/org/apache/ignite/lang/GridXSelfTest.java
```diff
@@ -53,6 +54,54 @@ public void testHasCause() {
...
+
+        assertEquals(npe, X.cause(exc, NullPointerException.class));
+
+        assertTrue(X.hasCause(exc, NullPointerException.class));
+
+        assertNull(X.cause(exc, ArithmeticException.class));
+
+        assertFalse(X.hasCause(exc, ArithmeticException.class));
+
+        npe.printStackTrace();
```
tkalkirill, line 102: It's not needed.

PR #2101, opened 2021-09-28
GG-33665 [IGNITE-15256] Java Thin: Fix ClassNotFoundException on service call after failover

ptupitsyn:
Clear thin client caches when cluster connection is lost.

The root cause of the issue is that client believes that server nodes preserve registered user types after reconnect, however, it could be the case that server nodes lose this information (e.g. full cluster restart with clean up of working dirs). This leads to ClassNotFoundException.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/client/PersonExternalizable.java
+++ b/modules/core/src/test/java/org/apache/ignite/client/PersonExternalizable.java
@@ -1,12 +1,11 @@
 /*
- * Licensed to the Apache Software Foundation (ASF) under one or more
- * contributor license agreements.  See the NOTICE file distributed with
- * this work for additional information regarding copyright ownership.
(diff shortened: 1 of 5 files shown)


modules/core/src/test/java/org/apache/ignite/client/PersonExternalizable.java
```diff
@@ -0,0 +1,57 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
isapego, line 2: 2021

isapego (approved):
Looks good to me, only 1 minor comment

modules/core/src/test/java/org/apache/ignite/client/PersonExternalizable.java
ptupitsyn: Fixed

PR #2116, opened 2021-10-11
GG-33862 Fixed spontaneous SocketTimeoutException in server socket accept (JDK-8247750).

ibessonov:
https://ggsystems.atlassian.net/browse/GG-33862

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
@@ -64,8 +64,10 @@ import java.net.InetAddress;
 import java.net.InetSocketAddress;
 import java.net.MalformedURLException;
 import java.net.NetworkInterface;
+import java.net.ServerSocket;
 import java.net.Socket;
 import java.net.SocketException;
+import java.net.SocketTimeoutException;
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -12557,4 +12559,24 @@ public static int hashToIndex(int hash, int size) {
...
+     * @throws IOException If an I/O error occurs when waiting for a connection.
+     * @see ServerSocket#accept()
+     */
+    public static Socket acceptServerSocket(ServerSocket srvrSock) throws IOException {
+        while (true) {
+            try {
+                return srvrSock.accept();
+            }
+            catch (SocketTimeoutException e) {
+                if (srvrSock.getSoTimeout() != 0)
```
SammyVimes, line 12577: AFAIK timeout can be "-1" (according to this https://hg.openjdk.java.net/jdk/jdk/rev/6e36b8dda5ab#l2.26), maybe `> 0` will be more suitable here?

SammyVimes (changes requested):

SammyVimes (approved):
LGTM!

PR #2120, opened 2021-10-13
GG-34005 Assertion error on SQL query splitting while dropping table

Berkof:
Replace assertion error with Exceptions, check nulls. Not fix all problems with exchange, just keep nodes from failing by assertion error.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
@@ -1223,6 +1223,9 @@ public class GridCachePartitionExchangeManager<K, V> extends GridCacheSharedMana
      * @return Last topology version before the provided one when affinity was modified.
      */
     public AffinityTopologyVersion lastAffinityChangedTopologyVersion(AffinityTopologyVersion topVer) {
+        if (topVer == null)
+            return null;
+
         if (topVer.topologyVersion() <= 0)
             return topVer;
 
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -1223,6 +1223,9 @@ public boolean hasPendingServerExchange() {
      * @return Last topology version before the provided one when affinity was modified.
      */
     public AffinityTopologyVersion lastAffinityChangedTopologyVersion(AffinityTopologyVersion topVer) {
+        if (topVer == null)
```
ygerzhedovich, line 1226: I'm not sure that is a valid check. Possibly we shouldn't be there in such a case at all. Let's ask @sergey-chugunov-1985

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/preloader/GridDhtPartitionDemander.java
```diff
@@ -1603,7 +1603,9 @@ public boolean isInitial() {
          */
         private void tryCancel() {
             if (STATE_UPD.compareAndSet(this, RebalanceFutureState.INIT, RebalanceFutureState.MARK_CANCELLED)) {
-                U.log(log, "Rebalancing marked as cancelled [grp=" + grp.cacheOrGroupName() +
+                String cacheName = (grp == null) ? "" : grp.cacheOrGroupName();
```
ygerzhedovich, line 1606: the same. @sergey-chugunov-1985

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/AbstractQueryTableLockAndConnectionPoolSelfTest.java
```diff
@@ -438,13 +438,19 @@ private void checkTablesLockQueryAndDropTable(final Ignite node) throws Exceptio
...
-                            || e.getMessage().contains("Failed to parse query. Table \"TEST\" not found")
-                            || e.getMessage().contains("Cache not found on local node (was concurrently destroyed?)")
-                            || e.getMessage().contains("Getting affinity for too old topology version that is already out of history")
-                            || e.getMessage().contains("Failed to find partitioned cache")
-                            || e.getMessage().contains("Table \"TEST\" not found")
-                            || e.getMessage().contains("Table PUBLIC.TEST already destroyed")
+                        String msg = e.getMessage();
+
+                        if (msg == null)
+                            System.err.println("msg == null");
```
ygerzhedovich, line 444: pls remove it

PR #2158, opened 2021-10-29
GG-33425 Fixed consistency issue for persistence atomic cache when a cache remove operations were occured while a node was offline and then started with the subsequent stop of the node without checkpoint and start it again.

sergeyuttsel:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionsEvictManager.java
```diff
@@ -718,19 +719,21 @@ public EvictReason reason() {
         /**
          * Submits the task for execution.
          *
-         * @return {@code True} if the task was submitted for execution.
+         * @return {@link Future} if the task was submitted for execution.
          */
-        public boolean start() {
+        public Future<?> start() {
```
sk0x50, line 724: I don't think that you need to change the contract. You always can use `finishFuture` from this task as follows:
```
    if (task.run())
        task.finishFuture().get()
```

sergeyuttsel: Fixed

PR #2162, opened 2021-11-01
GG-32736 [IGNITE-15830] Fix performance suggestion when Ignite is run…

rpuch:
… with a jdk other than 8x

When started on Java 11, current Ignite suggests to switch to the latest Java 8 JVM which seems weird as Ignite is officially tested on Java 8 and 11 ( https://ignite.apache.org/docs/latest/quick-start/java ), and the recommended version for running Ignite is 11.

This commit changes the behavior to 'Suggest Java 11 when run on any Java version other than 11'.

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/suggestions/JvmConfigurationSuggestionsTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/suggestions/JvmConfigurationSuggestionsTest.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ * Copyright 2021 GridGain Systems, Inc. and Contributors.
  *
  * Licensed under the GridGain Community Edition License (the "License");
  * you may not use this file except in compliance with the License.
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/apache/ignite/internal/suggestions/JvmConfigurationSuggestionsTest.java
```diff
@@ -0,0 +1,85 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ibessonov, line 2: I know that most people don't case but technically this should be 2021. I don't know why they put year into a license header, that was a horrible idea

ibessonov (approved):

modules/core/src/test/java/org/apache/ignite/internal/suggestions/JvmConfigurationSuggestionsTest.java
rpuch: Fixed the year

rpuch:
Thanks for the review!

PR #2167, opened 2021-11-03
GG-34172 [IGNITE-13553] Node failure with ClassNotFoundException when indexed field of QueryEntity differs from type used in put operation

vladErmakov07:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryTypeDescriptorImpl.java
```diff
@@ -637,6 +654,77 @@ else if (prop.scale() != -1 &&
...
+                            throw new IgniteSQLException("Type for a column '" + idxField +
+                                "' is not compatible with index definition. Expected '" +
+                                propType.getSimpleName() + "', actual type '" +
+                                propVal.getClass().getSimpleName() + "'");
+                        }
+                    }
+                }
+                else if (U.classForName(((BinaryObject)propVal).type().typeName(), Object.class, true)
+                        != propType) {
+                    if (coCtx.kernalContext().cacheObjects().typeId(propType.getName()) ==
```
tledkov-gridgain, line 704: May be adding this condition to the upper if?

modules/indexing/src/test/java/org/apache/ignite/testsuites/IgniteCacheWithIndexingTestSuite.java
```diff
@@ -119,7 +120,9 @@
 
     WrongIndexedTypesTest.class,
 
-    IndexPagesMetricsInMemoryTest.class
-})
+    IndexPagesMetricsInMemoryTest.class,
+
+    EnumClassImplementingIndexedInterfaceTest.class,
+    })
```
tledkov-gridgain, line 126: Wrong indent

PR #2174, opened 2021-11-10
GG-34208 AssertionError on getting cache size from the mbean on the inactive cluster

odiszapc:
Port of  https://issues.apache.org/jira/browse/IGNITE-13227


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheMetricsImpl.java
```diff
@@ -917,7 +917,7 @@ public void onRead(boolean isHit) {
 
     /**
      * Set callback for tx key collisions detection.
-     *
+     *getEntriesStat
```
sk0x50, line 920: Why this change is needed? :)

odiszapc: removed

PR #2182, opened 2021-11-12
GG-34235 Update netty version up to 4.1.70.Final to avoid CVEs

solveme:
CVEs list: 
- CVE-2021-37136
- CVE-2021-37137


aealeksandrov:
I will close this one because brach was created from 8.7-master instead of master

PR #2185, opened 2021-11-12
GG-34237 Enable old logic for inserting of logically equal compound PKs for old indexes.

vladErmakov07:


AMashenkov (approved):

korlov42 (changes requested):
Please assume the following test:

```
    @Test
    public void testInsertKeyWithNullKeyParts2() {
        Set<String> keyFields = new LinkedHashSet<>();
        keyFields.add("ID1");
        keyFields.add("ID2");

        grid(0).getOrCreateCache(
                new CacheConfiguration<>("test")
                        .setSqlSchema("PUBLIC")
                        .setQueryEntities(F.asList(
                                new QueryEntity()
                                        .setTableName("TEST")
                                        .setKeyType("MY_KEY_TYPE")
                                        .setValueType("MY_VALUE_TYPE")
                                        .addQueryField("ID1", Integer.class.getName(), "ID1")
                                        .addQueryField("ID2", Integer.class.getName(), "ID2")
                                        .addQueryField("VAL", Integer.class.getName(), "VAL")
                                        .setKeyFields(keyFields)
                        ))
        );

        sql("insert into test (id1, id2, val) values (1, null, 1);");

        assertThrows(log,
                () -> sql("insert into test (id1, id2, val) values (1, null, 1);"),
                TransactionDuplicateKeyException.class,
                "Duplicate key during INSERT");

        assertThrows(log,
                () -> sql("insert into test (id1, val) values (1, 1);"),
                TransactionDuplicateKeyException.class,
                "Duplicate key during INSERT");

        assertEquals(sql("SELECT * FROM test").getAll().size(), 1);
    }
```

The fix should work and for tables created though Cache API as well

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/dml/UpdatePlanBuilder.java
```diff
@@ -261,8 +266,16 @@ else if (stmt instanceof GridSqlMerge) {
             else
                 hasValProps = true;
         }
-
-        if (hasKeyProps && !rowKeys.isEmpty()) {
+    
+        rowKeys.removeIf(rowKey -> desc.type().property(rowKey).defaultValue() != null);
+    
+        boolean fillPKsWithDefaultsSysSetting = IgniteSystemProperties.getBoolean(
+                IgniteSystemProperties.IGNITE_SQL_FILL_ABSENT_PK_WITH_DEFAULTS, false);
```
AMashenkov, line 273: Let's cache the default value in class field.
Properties has a HashTable underneath with synchronized methods.
So, using Properties on hot-path is a bad practice.

vladErmakov07: Cached it in another class, thx.

AMashenkov (approved):

korlov42 (approved):

PR #2187, opened 2021-11-14
GG-33675 Speed-based throttling hasn't managed to protect the checkpoint buffer of a node

rpuch:
- speed-based throttling had a bug that prevented it to protect CP Buffer when CP progress was not yet reported; it's fixed here
- speed-based throttling has been heavily refactored to make it easier to understand
- the machinery comprising and surrounding write throttling has been commented

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ * Copyright 2021 GridGain Systems, Inc. and Contributors.
  *
  * Licensed under the GridGain Community Edition License (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
@@ -26,7 +26,7 @@ class CheckpointBufferKeeper {
     private final PageMemoryImpl pageMemory;
 
     /**
-     * Creates new instance.
+     * Creates a new instance.
      *
      * @param pageMemory page memory to use
      */
@@ -40,7 +40,7 @@ class CheckpointBufferKeeper {
      *
      * @return {@code true} iff Checkpoint Buffer is in danger zone
      */
-    public boolean isInDangerZone() {
+    boolean isInDangerZone() {
         int checkpointBufLimit = (int)(pageMemory.checkpointBufferPagesSize() * CP_BUF_FILL_THRESHOLD);
 
         return pageMemory.checkpointBufferPagesCount() > checkpointBufLimit;
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
@@ -1,0 +1,48 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ *
+ * Licensed under the GridGain Community Edition License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.gridgain.com/products/software/community-edition/gridgain-community-edition-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.cache.persistence.pagemem;
+
+import static org.apache.ignite.internal.processors.cache.persistence.pagemem.PagesWriteThrottlePolicy.CP_BUF_FILL_THRESHOLD;
+
+/**
+ * Logic used to determine whether Checkpoint Buffer is in danger zone and writer threads should be throttled.
+ */
+class CheckpointBufferKeeper {
+    /** Page memory. */
+    private final PageMemoryImpl pageMemory;
+
+    /**
+     * Creates new instance.
+     *
+     * @param pageMemory page memory to use
+     */
+    CheckpointBufferKeeper(PageMemoryImpl pageMemory) {
+        this.pageMemory = pageMemory;
+    }
+
+    /**
+     * Returns true if Checkpoint Buffer is in danger zone (filled too much) and, hence, writer threads need
+     * to be throttled.
+     *
+     * @return {@code true} iff Checkpoint Buffer is in danger zone
+     */
+    public boolean isInDangerZone() {
+        int checkpointBufLimit = (int)(pageMemory.checkpointBufferPagesSize() * CP_BUF_FILL_THRESHOLD);
+
+        return pageMemory.checkpointBufferPagesCount() > checkpointBufLimit;
+    }
+}
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
@@ -35,7 +35,8 @@ class CheckpointBufferKeeper {
     }
 
     /**
-     * Returns true if Checkpoint Buffer is in danger zone (filled too much) and, hence, writer threads need
+     * Returns true if Checkpoint Buffer is in danger zone (more than
+     * {@link PagesWriteThrottlePolicy#CP_BUF_FILL_THRESHOLD} of the buffer is filled) and, hence, writer threads need
      * to be throttled.
      *
      * @return {@code true} iff Checkpoint Buffer is in danger zone
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ * Copyright 2021 GridGain Systems, Inc. and Contributors.
  *
  * Licensed under the GridGain Community Edition License (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
@@ -36,7 +36,7 @@ class ExponentialBackoff {
     /**
      * Exponential backoff counter.
      */
-    private final AtomicInteger exponentialBackoffCntr = new AtomicInteger(0);
+    private final AtomicInteger exponentialBackoffCounter = new AtomicInteger(0);
 
     /**
      * Constructs a new instance with the given parameters.
@@ -56,15 +56,18 @@ class ExponentialBackoff {
      * @return next backoff duration in nanoseconds
      */
     public long nextDuration() {
-        int exponent = exponentialBackoffCntr.getAndIncrement();
+        int exponent = exponentialBackoffCounter.getAndIncrement();
         return (long) (startingBackoffNanos * Math.pow(backoffRatio, exponent));
     }
 
     /**
      * Resets the  exponential backoff counter so that next call to {@link #nextDuration()}
      * will return {@link #startingBackoffNanos}.
+     *
+     * @return {@code true} iff this backoff was not already in a reset state
      */
-    public void reset() {
-        exponentialBackoffCntr.set(0);
+    public boolean reset() {
+        int oldValue = exponentialBackoffCounter.getAndSet(0);
+        return oldValue != 0;
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
@@ -19,7 +19,7 @@ package org.apache.ignite.internal.processors.cache.persistence.pagemem;
 import java.util.concurrent.atomic.AtomicInteger;
 
 /**
- * Implements exponential backoff logic. Contains a counter and on each {@link #nextDuration()} increments it.
+ * Implements exponential backoff logic. Contains a counter and increments it on each {@link #nextDuration()}.
  * May be reset using {@link #reset()}.
  */
 class ExponentialBackoff {
@@ -61,7 +61,7 @@ class ExponentialBackoff {
     }
 
     /**
-     * Resets the  exponential backoff counter so that next call to {@link #nextDuration()}
+     * Resets the exponential backoff counter so that next call to {@link #nextDuration()}
      * will return {@link #startingBackoffNanos}.
      *
      * @return {@code true} iff this backoff was not already in a reset state
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
@@ -1,0 +1,70 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ *
+ * Licensed under the GridGain Community Edition License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.gridgain.com/products/software/community-edition/gridgain-community-edition-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
+ * WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.cache.persistence.pagemem;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Implements exponential backoff logic. Contains a counter and on each {@link #nextDuration()} increments it.
+ * May be reset using {@link #reset()}.
+ */
+class ExponentialBackoff {
+    /**
+     * Starting backoff duration.
+     */
+    private final long startingBackoffNanos;
+
+    /**
+     * Backoff ratio. Each next duration will be this times longer.
+     */
+    private final double backoffRatio;
+
+    /**
+     * Exponential backoff counter.
+     */
+    private final AtomicInteger exponentialBackoffCntr = new AtomicInteger(0);
+
+    /**
+     * Constructs a new instance with the given parameters.
+     *
+     * @param startingBackoffNanos duration of first backoff in nanoseconds
+     * @param backoffRatio         each next duration will be this times longer
+     */
+    public ExponentialBackoff(long startingBackoffNanos, double backoffRatio) {
+        this.startingBackoffNanos = startingBackoffNanos;
+        this.backoffRatio = backoffRatio;
+    }
+
+    /**
+     * Returns next backoff duration (in nanoseconds). As a side effect, increments the backoff counter so that
+     * next call will return a longer duration.
+     *
+     * @return next backoff duration in nanoseconds
+     */
+    public long nextDuration() {
+        int exponent = exponentialBackoffCntr.getAndIncrement();
+        return (long) (startingBackoffNanos * Math.pow(backoffRatio, exponent));
+    }
+
+    /**
+     * Resets the  exponential backoff counter so that next call to {@link #nextDuration()}
+     * will return {@link #startingBackoffNanos}.
+     */
+    public void reset() {
+        exponentialBackoffCntr.set(0);
+    }
+}
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 GridGain Systems, Inc. and Contributors.
+ * Copyright 2021 GridGain Systems, Inc. and Contributors.
  *
  * Licensed under the GridGain Community Edition License (the "License");
  * you may not use this file except in compliance with the License.
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
@@ -17,7 +17,7 @@
 package org.apache.ignite.internal.processors.cache.persistence.pagemem;
 
 /**
- *
+ * Logic used to protect memory (mainly, Checkpoint Buffer) from exhaustion using exponential backoff.
  */
 class ExponentialBackoffMemoryProtectionThrottle {
     /**
@@ -40,7 +40,7 @@ class ExponentialBackoffMemoryProtectionThrottle {
      *
      * @return park time in nanos
      */
-    long computeProtectionParkTime() {
+    long protectionParkTime() {
         return backoff.nextDuration();
     }
 
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
@@ -45,7 +45,7 @@ class ExponentialBackoffMemoryProtectionThrottle {
     }
 
     /**
-     * Resets backoff counter. Invoked when no throttling is needed anymore.
+     * Resets the backoff counter. Invoked when no throttling is needed anymore.
      *
      * @return {@code true} iff the backoff was not already in a reset state
      */
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
@@ -19,7 +19,7 @@ package org.apache.ignite.internal.processors.cache.persistence.pagemem;
 /**
  *
  */
-class CheckpointBufferProtectionThrottle {
+class ExponentialBackoffMemoryProtectionThrottle {
     /**
      * Starting throttle time. Limits write speed to 1000 MB/s.
      */
@@ -46,8 +46,10 @@ class CheckpointBufferProtectionThrottle {
 
     /**
      * Resets backoff counter. Invoked when no throttling is needed anymore.
+     *
+     * @return {@code true} iff the backoff was not already in a reset state
      */
-    void resetBackoffCounter() {
-        backoff.reset();
+    boolean resetBackoff() {
+        return backoff.reset();
     }
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
@@ -178,7 +178,7 @@ public interface PageMemoryEx extends PageMemory {
     public FullPageId pullPageFromCpBuffer();
 
     /**
-     * Whether Checkpoint Buffer is currently in the danger zone.
+     * Checks if the Checkpoint Buffer is currently in a danger zone.
      *
      * @return {@code true} if measures like throttling to protect Checkpoint Buffer should be applied,
      * and {@code false} otherwise.
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
@@ -178,12 +178,12 @@ public interface PageMemoryEx extends PageMemory {
     public FullPageId pullPageFromCpBuffer();
 
     /**
-     * Checks if the Checkpoint Buffer is currently in a danger zone.
+     * Checks if the Checkpoint Buffer is currently close to exhaustion.
      *
      * @return {@code true} if measures like throttling to protect Checkpoint Buffer should be applied,
      * and {@code false} otherwise.
      */
-    public boolean isCPBufferInDangerZone();
+    public boolean isCpBufferOverflowThresholdExceeded();
 
     /**
      * Total pages can be placed to memory.
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
@@ -118,7 +118,8 @@ public interface PageMemoryEx extends PageMemory {
      * @throws IgniteException If checkpoint has been already started and was not finished.
      * @param allowToReplace The sign which allows to replace pages from a checkpoint by page replacer.
      */
-    public GridMultiCollectionWrapper<FullPageId> beginCheckpoint(IgniteInternalFuture allowToReplace) throws IgniteException;
+    public GridMultiCollectionWrapper<FullPageId> beginCheckpoint(IgniteInternalFuture allowToReplace)
+            throws IgniteException;
 
     /**
      * Finishes checkpoint operation.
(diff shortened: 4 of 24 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ProgressInstantSpeedCalculation.java
```diff
@@ -0,0 +1,63 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.cache.persistence.pagemem;
+
+/**
+ * Represents instant speed calculation of some progress starting from zero.
+ */
+class ProgressInstantSpeedCalculation {
```
sashapolo, line 22: All this class does is wrapping an `IntervalBasedMeasurement` in order to rename its methods. This may be an overkill

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+import org.apache.ignite.internal.util.GridConcurrentHashSet;
+import org.apache.ignite.lang.IgniteOutClosure;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Speed-based throttle used to protect clean pages from exhaustion.
+ */
+class SpeedBasedCleanPagesProtectionThrottle {
+    /**
+     * Maximum dirty pages in region.
```
sashapolo, line 33: Maximum percentage of dirty pages?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+     */
+    private final PageMemoryImpl pageMemory;
+
+    /**
+     * Checkpoint progress provider.
+     */
+    private final IgniteOutClosure<CheckpointProgress> cpProgress;
+
+    /**
+     * Total pages which is possible to store in page memory.
```
sashapolo, line 53:
```suggestion
     * Total pages possible to store in page memory.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+     */
+    private volatile double targetDirtyRatio;
+
+    /**
+     * Current dirty pages ratio (percent of dirty pages in most used segment), negative value means no cp is running.
+     */
+    private volatile double currDirtyRatio;
+
+    /**
+     * Threads set. Contains identifiers of all threads which were marking pages for current checkpoint.
```
sashapolo, line 74:
```suggestion
     * Threads set. Contains identifiers of all threads which were marking pages for the current checkpoint.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+     */
+    private final Set<Long> threadIds = new GridConcurrentHashSet<>();
+
+    /**
+     * Counter of written pages from checkpoint. Value is saved here for detecting checkpoint start.
```
sashapolo, line 79:
```suggestion
     * Counter of written pages from a checkpoint. Value is saved here for detecting a checkpoint start.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+     */
+    private final AtomicInteger lastObservedWritten = new AtomicInteger(0);
+
+    /**
+     * Dirty pages ratio was observed at checkpoint start (here start is moment when first page was actually saved to
```
sashapolo, line 84:
```suggestion
     * Dirty pages ratio that was observed at a checkpoint start (here the start is a moment when the first page was actually saved to
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+        this.pageMemory = pageMemory;
+        this.cpProgress = cpProgress;
+        this.markSpeedAndAvgParkTime = markSpeedAndAvgParkTime;
+
+        totalPages = pageMemory.totalPages();
+    }
+
+    /**
+     * Computes next duration (in nanos) to throttle a thread.
+     * Might return #NO_THROTTLING_MARKER as a marker that no throttling should be applied.
```
sashapolo, line 115:
```suggestion
     * Might return {@link PagesWriteSpeedBasedThrottle#NO_THROTTLING_MARKER} as a marker that no throttling should be applied.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+        final long donePages = cpDonePagesEstimation(cpWrittenPages);
+
+        final long markDirtySpeed = markSpeedAndAvgParkTime.getSpeedOpsPerSec(curNanoTime);
+        cpWriteSpeed.setProgress(donePages, curNanoTime);
+        final long curCpWriteSpeed = cpWriteSpeed.getOpsPerSecond(curNanoTime);
+
+        final int cpTotalPages = cpTotalPages();
+
+        if (cpTotalPages == 0) {
+            // ??? It would be great to ask dspavlov (original author) or Alexey Goncharuk (reviewer) why we need
```
sashapolo, line 154: I guess you'd better not forget to ask them)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+
+import org.apache.ignite.internal.processors.cache.persistence.checkpoint.CheckpointProgress;
+import org.apache.ignite.internal.util.GridConcurrentHashSet;
+import org.apache.ignite.lang.IgniteOutClosure;
+import org.jetbrains.annotations.NotNull;
+
+/**
+ * Speed-based throttle used to protect clean pages from exhaustion.
+ */
+class SpeedBasedCleanPagesProtectionThrottle {
```
sashapolo, line 31: Can you please add some javadocs about the responsibilities of this class? It has a lot of methods all performing some calculations, which are quite hard to aprehend

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+
+package org.apache.ignite.internal.processors.cache.persistence.pagemem;
+
+import static org.apache.ignite.internal.processors.cache.persistence.pagemem.PagesWriteThrottlePolicy.CP_BUF_FILL_THRESHOLD;
+
+/**
+ * Logic used to determine whether Checkpoint Buffer is in danger zone and writer threads should be throttled.
+ */
+class CheckpointBufferKeeper {
```
sergey-chugunov-1985, line 24: I suggest to rename this class: CheckpointBufferOverflowWatchdog, it better reflects its purpose.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
```diff
@@ -0,0 +1,48 @@
...
+        this.pageMemory = pageMemory;
+    }
+
+    /**
+     * Returns true if Checkpoint Buffer is in danger zone (filled too much) and, hence, writer threads need
+     * to be throttled.
+     *
+     * @return {@code true} iff Checkpoint Buffer is in danger zone
+     */
+    public boolean isInDangerZone() {
```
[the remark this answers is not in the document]
sergey-chugunov-1985: I suggest to rename the method to `isOverflowThresholdExceeded`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
```diff
@@ -0,0 +1,73 @@
...
+
+package org.apache.ignite.internal.processors.cache.persistence.pagemem;
+
+import java.util.concurrent.atomic.AtomicInteger;
+
+/**
+ * Implements exponential backoff logic. Contains a counter and on each {@link #nextDuration()} increments it.
+ * May be reset using {@link #reset()}.
+ */
+class ExponentialBackoff {
```
sergey-chugunov-1985, line 25: Lets rename the class to `ExponentialBackoffCalculator` as the whole purpose of the class is simply calculating throttling intervals lenghts.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
```diff
@@ -0,0 +1,55 @@
...
+ * See the License for the specific language governing permissions and
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.processors.cache.persistence.pagemem;
+
+/**
+ * Logic used to protect memory (mainly, Checkpoint Buffer) from exhaustion using exponential backoff.
+ */
+class ExponentialBackoffMemoryProtectionThrottle {
```
sergey-chugunov-1985, line 22: Lets rename the class to `ExponentialBackoffThrottlingStrategy`.
Reasoning:
1. It is multipurpose and can be used not only to protect free memory.
2. We shouldn't use the same name (Throttle) for entities of two types: that encapsulate logic of park time calculation and that do actual throttling (PagesWriteThrottle, PagesWriteSpeedBasedThrottle).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
```diff
@@ -177,9 +178,12 @@ public void checkpointWritePage(
...
 
     /**
-     * Calculates throttling condition.
+     * Whether Checkpoint Buffer is currently in the danger zone.
+     *
+     * @return {@code true} if measures like throttling to protect Checkpoint Buffer should be applied,
+     * and {@code false} otherwise.
      */
-    public boolean shouldThrottle();
+    public boolean isCPBufferInDangerZone();
```
sergey-chugunov-1985, line 186: This name isn't very informative but is scary. I suggest to change it to more neutral `isCPBufferOverflowThresholdExceeded`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottlePolicy.java
```diff
@@ -56,9 +92,10 @@ default void tryWakeupThrottledThreads() {
...
-     * @return {@code True} if throttling should be enabled, and {@code False} otherwise.
+     * Whether Checkpoint Buffer is currently in the danger zone.
+     *
+     * @return {@code true} if measures like throttling to protect Checkpoint Buffer should be applied,
+     * and {@code false} otherwise.
      */
-    default boolean shouldThrottle() {
-        return false;
-    }
+    boolean isCPBufferInDangerZone();
```
sergey-chugunov-1985, line 100: The same rename suggestion.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ProgressInstantSpeedCalculation.java
sergey-chugunov-1985: I agree with @sashapolo, this class is a very thin wrapper around `IntervalBasedMeasurement`, using the measurement directly shouldn't hurt code readability that much.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
sergey-chugunov-1985: I also suggest to rename the class to match with its counterpart, `ExponentialBackoffThrottlingStrategy`.

A name like `SpeedBasedMemoryConsumptionThrottlingStrategy` should be better.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+     * @param curCpWriteSpeed     pages/second checkpoint write speed. 0 speed means 'no data'.
+     * @param cpTotalPages        total pages in checkpoint.
+     * @return pages/second to mark to mark all clean pages as dirty till the end of checkpoint. 0 speed means 'no
+     * data', or when we are not going to throttle due to the current dirty pages ratio being too high
+     */
+    private long calcSpeedToMarkAllSpaceTillEndOfCp(double dirtyPagesRatio,
+                                                    long donePages,
+                                                    long curCpWriteSpeed,
+                                                    int cpTotalPages) {
+
```
sergey-chugunov-1985, line 343: Curly bracket should occupy this line.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
```diff
@@ -0,0 +1,500 @@
...
+            //for slow cp is completed now, drop previous dirty page percent
+            initDirtyRatioAtCpBegin = newMinRatio;
+        }
+    }
+
+    /**
+     * Resets the throttle to its initial state (for example, in the beginning of a checkpoint).
+     */
+    void initialize() {
```
sergey-chugunov-1985, line 487: Naming here isn't perfect: `initialize` implies to be called once, but this method is called over and over again on the same object.

How about `reset` as a name for this method?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
```diff
@@ -0,0 +1,73 @@
...
+
+    /**
+     * Backoff ratio. Each next duration will be this times longer.
+     */
+    private final double backoffRatio;
+
+    /**
+     * Exponential backoff counter.
+     */
+    private final AtomicInteger exponentialBackoffCounter = new AtomicInteger(0);
```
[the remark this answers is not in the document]
rpuch: Could you please elaborate? I don't seem to understand what change is suggested or what is wrong with the current code at this line.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -29,62 +30,26 @@
 /**
  * Throttles threads that generate dirty pages during ongoing checkpoint.
  * Designed to avoid zero dropdowns that can happen if checkpoint buffer is overflowed.
- * Uses average checkpoint write speed and moment speed of marking pages as dirty.<br>
+ * When a page is in checkpoint and Checkpoint Buffer is filled over 2/3, uses exponentially
+ * growing sleep time to throttle.
+ * Otherwise, uses average checkpoint write speed and instant speed of marking pages as dirty.<br>
  *
  * See also: <a href="https://github.com/apache/ignite/tree/master/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem#speed-based-throttling">Speed-based throttling description</a>.
```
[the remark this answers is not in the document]
rpuch: Applied your suggestion

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -29,62 +30,26 @@
 /**
  * Throttles threads that generate dirty pages during ongoing checkpoint.
  * Designed to avoid zero dropdowns that can happen if checkpoint buffer is overflowed.
- * Uses average checkpoint write speed and moment speed of marking pages as dirty.<br>
+ * When a page is in checkpoint and Checkpoint Buffer is filled over 2/3, uses exponentially
```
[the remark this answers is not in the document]
rpuch: The overall logic looks like this:

```
if (isPageInCheckpoint && cpBufferIsInDangerZone()) {
    throttleExponentiallyToProtectCPBuffer();
} else {
    maybeApplySpeedBasedThrottlingToProtectCleanPages();
}
```

`isPageInCheckpoint` is a parameter of `onMarkDirty()` method.

'When a page is in checkpoint' means that `isPageInCheckpoint` is `true`. `onMarkDirty()` is also called from `PageMemoryImpl#allocate()` where `false` is passed as `isPageInCheckpoint`.

Would 'When the page at question is not included in the current checkpoint' instead of 'When a page is in checkpoint' be sufficient, what do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
```diff
@@ -0,0 +1,48 @@
...
+     * Creates new instance.
+     *
+     * @param pageMemory page memory to use
+     */
+    CheckpointBufferKeeper(PageMemoryImpl pageMemory) {
+        this.pageMemory = pageMemory;
+    }
+
+    /**
+     * Returns true if Checkpoint Buffer is in danger zone (filled too much) and, hence, writer threads need
```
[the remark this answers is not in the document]
rpuch: I've reworded it to include #CP_BUF_FILL_THRESHOLD (I wanted to avoid writing 2/3 explicitly to avoid getting an outdated javadoc if the threshold gets changed). Is this sufficient, or should I improve it?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -127,112 +99,57 @@ public PagesWriteSpeedBasedThrottle(
...
+    private long computeThrottlingParkTime(boolean isPageInCheckpoint, long curNanoTime) {
+        if (shouldThrottleToProtectCPBuffer(isPageInCheckpoint))
+            return computeCPBufferProtectionParkTime();
         else {
-            if (isPageInCheckpoint)
-                exponentialBackoffCntr.set(0);
-
-            if (level == ThrottleMode.NO)
-                throttleParkTimeNs = 0;
+            if (isPageInCheckpoint) {
```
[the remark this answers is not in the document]
rpuch: Oops, my bad :) Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -127,112 +99,57 @@ public PagesWriteSpeedBasedThrottle(
...
         }
+    }
 
-        if (throttleParkTimeNs > 0) {
-            recurrentLogIfNeed();
-
-            doPark(throttleParkTimeNs);
-        }
+    /***/
+    private boolean shouldThrottleToProtectCPBuffer(boolean isPageInCheckpoint) {
```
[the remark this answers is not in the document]
rpuch: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -127,112 +99,57 @@ public PagesWriteSpeedBasedThrottle(
...
+        return isPageInCheckpoint && isCPBufferInDangerZone();
+    }
 
-        pageMemory.metrics().addThrottlingTime(U.nanosToMillis(System.nanoTime() - curNanoTime));
+    /***/
+    private long computeCPBufferProtectionParkTime() {
```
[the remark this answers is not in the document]
rpuch: Done

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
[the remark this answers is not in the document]
rpuch: Added a comment. Would this be ok?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
```diff
@@ -127,112 +99,57 @@ public PagesWriteSpeedBasedThrottle(
...
+        return cpBufferProtector.protectionParkTime();
+    }
 
-        speedMarkAndAvgParkTime.addMeasurementForAverageCalculation(throttleParkTimeNs);
+    /***/
+    private long computeCleanPagesProtectionParkTime(long curNanoTime) {
```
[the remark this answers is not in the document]
rpuch: Done

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffTest.java
```diff
@@ -0,0 +1,64 @@
...
+import static org.hamcrest.MatcherAssert.assertThat;
+import static org.hamcrest.Matchers.is;
+import static org.junit.Assert.assertFalse;
+import static org.junit.Assert.assertTrue;
+
+/**
+ * Tests for {@link ExponentialBackoff}.
+ */
+public class ExponentialBackoffTest {
+    private final ExponentialBackoff backoff = new ExponentialBackoff(1000, 1.1);
```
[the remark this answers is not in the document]
rpuch: Introduced constants

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottlePolicy.java
```diff
@@ -39,11 +76,10 @@
...
 
     /**
-     * Callback to try wakeup throttled threads.
+     * Callback to try wakeup throttled threads. Invoked when the Checkpoint Buffer use drops below a certain
+     * threshold.
      */
-    default void tryWakeupThrottledThreads() {
-        // No-op.
-    }
+    void tryWakeupThrottledThreads();
```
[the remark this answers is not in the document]
rpuch: I'm not the inventor of this name, but I think it is because it wakes the threads up conditionally:

```
if (!isCPBufferInDangerZone()) {
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ProgressInstantSpeedCalculation.java
rpuch: The thing is that `IntervalBasedMeasurement` is pretty non-obvious and involved. To understand what it does, you need to read a good part of it. In this concrete place in the code, only small fraction of its functionality is used: just computation of progress speed, where we provide the explicit progress value ourselves. All the averaging and ability to remember history is not used here, but it's not that obvious.

I introduced this adapter to clarify what it does. It has a clear (hopefully) javadoc both on the class and methods. It is just for computing progress speed, that's all. So, in my opinion, it does add value to the code: it hides complexity.

What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
rpuch: Fixed it as 'Maximum fraction of dirty pages' as percentage should probably be in percents.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
rpuch: It's not just a calculator, it also holds state (backoff counter), so `Calculator` seems non-optimal. So it represents the concrete instance of an exponential backoff process, and just `ExponentialBackoff` seems an adequate name.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
rpuch: Renamed the class

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoffMemoryProtectionThrottle.java
rpuch: Renamed the class

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
rpuch: Renamed the class

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PageMemoryEx.java
rpuch: Renamed the method to `isCpBufferOverflowThresholdExceeded` (with 'Cp' instead of 'CP' because there is a method called `pullPageFromCpBuffer` in this interface.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottlePolicy.java
rpuch: Renamed to `isCpBufferOverflowThresholdExceeded`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
rpuch: I agree. Renamed to `reset()`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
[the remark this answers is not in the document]
sashapolo: > Would 'When the page at question is not included in the current checkpoint' instead of 'When a page is in checkpoint' be sufficient, what do you think?

Yes, that's clear to me now

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottlePolicy.java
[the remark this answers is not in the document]
sashapolo: but then those threads would not be throttled, no?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ProgressInstantSpeedCalculation.java
sashapolo: your reasoning is good, adding more clarity in the existing code is cool. You can leave this class if you want to

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/CheckpointBufferKeeper.java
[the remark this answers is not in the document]
sashapolo: yeah, that's exactly what I meant

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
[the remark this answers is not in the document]
sashapolo: looks good to me

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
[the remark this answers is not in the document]
sashapolo: I think that the current code is absolutely fine. I'm suggesting a minor "improvement": we currently store the exponent, which we then use to compute the backoff value using the `pow` function. Instead we can cache the previous backoff value and simply multiply it by `backoffRatio` on each call. This is just an idea, not a requirement

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteSpeedBasedThrottle.java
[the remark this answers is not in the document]
rpuch: Applied the change as 'When the page in question is not included in the current checkpoint' ('in question', not 'at question').

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ExponentialBackoff.java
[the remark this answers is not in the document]
rpuch: Oh, I see what you mean. Yes, we could try storing the current backoff value in an `AtomicLong` and update it using `getAndUpdate(prev -> (long) (prev * backoffRatio))`. But this would compute slightly different backoff values (because we would accumulate rounding error on each step), which does not see important, but still different. Also, under contention, the new code would have to do some wasted multiplications, while now it only wastes increments (which are cheaper). So it is not that obvious whether it would perform better or not under contention.

Let's leave it as is for now and improve it later when we have time for this specific improvement (including perf tests).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
rpuch: Changed the formatting

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/ProgressInstantSpeedCalculation.java
rpuch: Ok, I'm leaving the class for now, and I tried to improve its (and its methods') names and added  more javadoc

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/SpeedBasedCleanPagesProtectionThrottle.java
rpuch: I think I managed to understand when and why it's needed. I've added some comments and renamed a method to put this knowledge to the code.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottlePolicy.java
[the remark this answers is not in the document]
rpuch: Imagine the following situation:

1. Writes are very fast, so the CP Buffer is over 2/3
2. Writes get throttled, threads are parked
3. Meanwhile, the checkpointer does its job and offloads the CP Buffer so that it gets filled only 1/2
4. This is the moment when `tryWakeupThrottledThreads()` gets called. New writes are not throttled anymore, but there could still be some parked threads that were parked before (when the CP Buffer was still over 2/3).

So it seems reasonable to try and wake them up.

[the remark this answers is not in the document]
sashapolo: yes, I understand that, I'm just confused that this method's name starts with `try` as if it may fail to wake up some threads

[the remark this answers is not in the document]
rpuch: How about just `wakeupThrottledThreads()`? Or, if we want to keep the flavor of conditionality, `wakeupThrottledThreadsIfCpBufferIsNotOverloaded()`?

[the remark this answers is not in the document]
sashapolo: I'm all for `wakeupThrottledThreads`

sashapolo (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/pagemem/PagesWriteThrottlePolicy.java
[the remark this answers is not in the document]
rpuch: Renamed the method as suggested

SammyVimes (approved):
Looks good overall, the only thing I was able to found is a small typo :D

rpuch:
Guys, thanks for your reviews!

sergey-chugunov-1985 (approved):
@rpuch Patch looks good to me as well, I approve its merging. Thank you!

rpuch:
Thanks for your reviews, guys! Merged to master.

PR #2199, opened 2021-11-18
GG-34294 Index de-fragmentation fails on DECIMAL and VARCHAR columns

tledkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/defragmentation/IndexingDefragmentation.java
```diff
@@ -278,23 +287,14 @@ private boolean defragmentTable(
...
-
-                final int columnIndex = inlineIndexColumn.columnIndex();
-
-                values.put(columnIndex, value);
-            }
-        }
+        if (inlineIdxs != null)
+            values = PageUtils.getBytes(pageAddr, off, payloadSize);
+        else
+            values = Utils.EMPTY_BYTES;
```
ibessonov, line 297: I see that in IO you write `assert inlineIdxs != null : "no inline index helpers";`, but here it's a valid `else` branch. Which code do I trust?

SammyVimes: yes, you're right, that check is correct, and the other one is odd

ibessonov (approved):

PR #2204, opened 2021-11-22
IGNITE_WAIT_FOR_BACKUPS_ON_SHUTDOWN system property is deprecated.

sergeyuttsel:


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -1501,6 +1501,11 @@ private long checkPoolStarvation(
                 EventType.EVT_NODE_JOINED, localNode());
 
         startTimer.finishGlobalStage("Await exchange");
+
+        if (IgniteSystemProperties.getString(IgniteSystemProperties.IGNITE_WAIT_FOR_BACKUPS_ON_SHUTDOWN) != null) {
```
sk0x50, line 1505: Cosmetic:
I think it would be nice to extract this check into a separate method something like `ackIPv4StackFlagIsSet` etc
and move it to the beginning of the `start` method.

```
        // Ack various information.
        ackAsciiLogo();
        ....
        ackIPv4StackFlagIsSet();
        ackShutdownPolicy();
```

sergeyuttsel: ОК. That place seems more appropriate.

PR #2205, opened 2021-11-22
GG-34258 Add getting/changing the priority of compute tasks for monitoring

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-34258


modules/core/src/main/java/org/apache/ignite/internal/managers/collision/GridCollisionManager.java
```diff
@@ -130,4 +129,11 @@ public void onCollision(
             });
         }
     }
+
+    /**
+     * @return Collision SPI.
+     */
+    public CollisionSpi collisionSpi() {
```
ibessonov, line 136: This is literally equal to `ctx.config().getCollisionSpi()`, right? Why do you need this method?

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
```diff
@@ -343,6 +357,15 @@ public GridJobProcessor(GridKernalContext ctx) {
 
                 return new ComputeJobView(e.getKey(), e.getValue(), state);
             });
+
+        if (!jobAlwaysActivate && ctx.collision().collisionSpi() instanceof PriorityQueueCollisionSpi) {
```
ibessonov, line 361: Maybe extract `ctx.collision().collisionSpi()` into a variable?

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
```diff
@@ -2282,4 +2311,49 @@ private JobsMap(int initCap, float loadFactor, int concurLvl) {
...
     }
+
+    /**
+     * Callback on changing task attributes.
+     *
+     * @param sesId Session ID.
+     * @param jobId Job ID.
+     * @param attrs Changed attributes.
+     */
+    public void onChangeTaskAttributes(IgniteUuid sesId, IgniteUuid jobId, Map<?, ?> attrs) {
```
ibessonov, line 2322: Why is it public?

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -1064,7 +1062,7 @@ private void sendSessionAttributes(Map<?, ?> attrs, GridTaskSessionImpl ses)
 
                 UUID nodeId = sib.nodeId();
 
-                if (!nodeId.equals(locNodeId) && !sib.isJobDone() && !rcvrs.contains(nodeId))
+                if (!sib.isJobDone())
```
ibessonov, line 1065: So, you add a job to `rcvrs` collection even if it's deployed locally

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
```diff
@@ -1087,14 +1085,20 @@ private void sendSessionAttributes(Map<?, ?> attrs, GridTaskSessionImpl ses)
...
 
             UUID nodeId = sib.nodeId();
 
-            // Pair can be null if job is finished.
-            if (rcvrs.remove(nodeId)) {
+            if (locNodeId.equals(nodeId)) {
+                // Local job notification.
+                ctx.job().onChangeTaskAttributes(ses.getId(), s.getJobId(), attrs);
+            }
+            else if (rcvrs.remove(nodeId)) {
```
ibessonov, line 1100: ... but don't remove it after. Am I missing something?

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
tkalkirill: We need to call from class `GridTaskProcessor#sendSessionAttributes` and `GridJobProcessor#processTaskSessionRequest` and therefore `public`.

modules/core/src/main/java/org/apache/ignite/internal/managers/collision/GridCollisionManager.java
tkalkirill: You're right, I can do that.

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
tkalkirill: You're right, I can do that.

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
tkalkirill: You're right, I fix it.

modules/core/src/main/java/org/apache/ignite/internal/processors/task/GridTaskProcessor.java
tkalkirill: You're right, it's bad.

modules/core/src/main/java/org/apache/ignite/internal/managers/collision/GridCollisionManager.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2021 GridGain Systems, Inc. and Contributors.
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ibessonov: Hmm...

ibessonov (approved):

PR #2208, opened 2021-11-23
GG-34220 Support for JDK17

SammyVimes:


ibessonov (approved):

modules/tools/pom.xml
```diff
@@ -37,13 +37,23 @@
         <profile>
             <id>java-9+</id>
```
rpuch, line 38: Do we still want to call this profile `java-9+` (as it is not going to include 15+ anymore)? Wouldn't something like `java-9-14` more honest? Or do we want to be 'backward-compatible' with profile naming?

parent/pom.xml
```diff
@@ -1203,7 +1203,7 @@
         <profile>
             <id>java-9+</id>
```
rpuch, line 1204: Same question about profile naming as above

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/filehandle/FileWriteHandleImpl.java
```diff
@@ -611,4 +586,137 @@ public String safePosition() {
...
+                force.invoke(mappedMemoryUtils, fd.get(buf), address, isSync, index, len);
+            }
+            catch (IllegalAccessException | InvocationTargetException e) {
+                throw new IgniteCheckedException(e);
+            }
+        }
+    }
+
+    /**
+     * Runs fsync on pre-java15 JVMs that doesn't offer a possibility to fsync mapped byte buffers in an aligned way.
```
rpuch, line 682: Probably, 'that don't offer'?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/filehandle/FileWriteHandleImpl.java
```diff
@@ -611,4 +586,137 @@ public String safePosition() {
...
+        if (javaVersion >= 15)
+            return new JDK15FSyncer();
+
+        return new LegacyFSyncer();
+    }
+
+    /**
+     * Runs fsync operation on JRE15 and higher using MappedMemoryUtils which provides aligned fsync.
+     */
+    private static class JDK15FSyncer implements MMapFSyncer {
```
rpuch, line 623: Does this syncer perform better than the legacy one? Its code does not seem much simpler, but now we have 2 syncers to support.

SammyVimes: The old one doesn't work on JDK15 and newer

rpuch: Oh, the most obvious one :)

rpuch (approved):

PR #2216, opened 2021-11-30
Gg 34149

vmalin:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/PartitionCountersNeighborcastFuture.java
```diff
@@ -111,15 +112,22 @@ public void init() {
...
 
                 miniFut.onDone();
             }
         }
 
+        if (log.isInfoEnabled()) {
+            log.info("Starting delivery partition countres to remote nodes [txId=" + tx.nearXidVersion() +
+                ", futId=" + futId +
+                ", topVer=" + tx.topologyVersion() +
+                ", msgs=" + peerMsgs);
```
vmalin, line 128: I'm a bit worried about this message's length. In the test it's ok, but can it become too big on the client's clusters where there are much more nodes and partitions?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/TxCountersNeighbourcastOnRecoveryLogTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        cleanPersistenceDir();
+
+        super.afterTest();
+    }
+
+    /** */
+    @Test
+    public void test() throws Exception {
+        final int NODES_CNT = 3;
```
vmalin, line 88: If I remember correctly capital letters are used for class-level constants. Local constants are defined the same way as regular variables

denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/PartitionCountersNeighborcastFuture.java
denis-chudov: The length of `peerMsgs` list equals to the size of nodes participating in transaction, and the sizes of this list (i.e. PartitionUpdateCountersMessage ) is limited with partition counters associated with this particular transaction. Transaction details we already log in cases like LRT, so I dont think this change can cause some additional problems.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/TxCountersNeighbourcastOnRecoveryLogTest.java
```diff
@@ -0,0 +1,165 @@
...
+
+        Pattern cntrsDeliveryPattern = Pattern.compile("Starting delivery partition countres to remote nodes.*?msgs=" +
+            ".*PartitionUpdateCountersMessage\\{cacheId=1544803905, size=1, cntrs=\\[part=.*?, initCntr=1, cntr=1\\]\\}");
```
alievmirza, line 91: Let's use here utility method to get cacheId instead of writing direct id

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/PartitionCountersNeighborcastFuture.java
alievmirza, line 128: Should we add here `[` `]` for list of messages?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/TxCountersNeighbourcastOnRecoveryLogTest.java
denis-chudov: fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/PartitionCountersNeighborcastFuture.java
denis-chudov: list is wrapped by brackets by default, but actually I forgot to add the last closing bracket, thanks for making me double check.

PR #2218, opened 2021-12-02
GG-34257 [IGNITE-15921] Thin client: drop connection on invalid handshake without allocating buffer

ptupitsyn:
* Check first 3 bytes of the handshake message without allocating the buffer, drop connection in case of garbage data.
* Add 64 MiB handshake size limit.


isapego (approved):
Looks good

PR #2219, opened 2021-12-06
GG-32983 [IGNITE-14375] Pending messages can be erroneously send

zstan:


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/ServerImpl.java
```diff
@@ -6420,6 +6396,36 @@ private void processCustomMessage(TcpDiscoveryCustomEventMessage msg, boolean wa
...
+         * @param msg Processed message.
+         * @return {@code true} If message was appended to pending queue.
+         */
+        private boolean posponeUndeliveredMessages(final TcpDiscoveryCustomEventMessage msg) {
+            boolean joiningEmpty;
+
+            synchronized (mux) {
+                joiningEmpty = joiningNodes.isEmpty();
+
+                if (log.isDebugEnabled())
```
sergey-chugunov-1985, line 6411: If I understand the code correctly here we will always print to debug logs this message as no check for actual value of `joiningEmpty` variable is made.

zstan: thanks! i fixed this issue.

PR #2220, opened 2021-12-06
GG-34370: Logging globalQueryId for long running queries.

AMashenkov:
https://ggsystems.atlassian.net/browse/GG-34370


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -630,7 +630,7 @@ private GridQueryFieldsResult executeSelectLocal(
 
                         H2Utils.bindParameters(stmt, F.asList(params));
 
-                        H2QueryInfo qryInfo = new H2QueryInfo(H2QueryInfo.QueryType.LOCAL, stmt, qry, qryId);
+                        H2QueryInfo qryInfo = new H2QueryInfo(H2QueryInfo.QueryType.LOCAL, stmt, qry, ctx.discovery().localNode(), qryId);
```
ygerzhedovich, line 633: seems we could use nodeId instead of ctx.discovery().localNode()

PR #2226, opened 2021-12-07
GG-34399 Add isFullSupport flag to compute task monitoring

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-34399

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/task/monitor/ComputeTaskStatus.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/task/monitor/ComputeTaskStatus.java
@@ -65,7 +65,7 @@ public class ComputeTaskStatus implements ComputeTaskStatusSnapshot {
     @Nullable private final Throwable failReason;
 
     /** Availability of changing task attributes. */
-    private final boolean fullSupported;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/task/monitor/ComputeTaskStatus.java
```diff
@@ -64,6 +64,9 @@
     /** Reason for the failure of the task. */
     @Nullable private final Throwable failReason;
 
+    /** Availability of changing task attributes. */
+    private final boolean fullSupported;
```
sergey-chugunov-1985, line 68: How about renaming the field and all corresponding accessor methods to a simpler `fullSupport`? In that case it matches with `GridTaskSessionInternal#isFullSupport`.

sergey-chugunov-1985 (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/task/monitor/ComputeTaskStatus.java
tkalkirill: Yes, I think it can be renamed.

PR #2229, opened 2021-12-09
GG-34419 Add createdBy for compute grid

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-34419


nva (commented):
LGTM. If possible, I would add a test with Ignite Authentication

PR #2233, opened 2021-12-10
GG-34397 Snapshot doesn't work for caches with TDE in case these caches have a lot of values

rpuch:
https://ggsystems.atlassian.net/browse/GG-34397

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
@@ -1261,7 +1261,6 @@ public abstract class AbstractDataPageIO<T extends Storable> extends PageIO impl
     /** {@inheritDoc} */
     @Override public void compactPage(ByteBuffer page, ByteBuffer out, int pageSize) {
         assertPageType(page);
-        assertPageType(out);
 
         // TODO May we compactDataEntries in-place and then copy compacted data to out?
         copyPage(page, out, pageSize);
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
@@ -237,6 +237,8 @@ public abstract class AbstractDataPageIO<T extends Storable> extends PageIO impl
      * @param freeListPageId Free list page ID.
      */
     public void setFreeListPageId(long pageAddr, long freeListPageId) {
+        assertPageType(pageAddr);
+
         PageUtils.putLong(pageAddr, FREE_LIST_PAGE_ID_OFF, freeListPageId);
     }
 
@@ -294,6 +296,7 @@ public abstract class AbstractDataPageIO<T extends Storable> extends PageIO impl
      */
     private void setFirstEntryOffset(long pageAddr, int dataOff, int pageSize) {
         assert dataOff >= ITEMS_OFF + ITEM_SIZE && dataOff <= pageSize : dataOff;
+        assertPageType(pageAddr);
 
         PageUtils.putShort(pageAddr, FIRST_ENTRY_OFF, (short)dataOff);
     }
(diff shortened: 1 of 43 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
```diff
@@ -386,16 +391,18 @@ public int getRowsCount(long pageAddr) {
     /**
      * @param pageAddr Page address.
      * @param c Closure.
-     * @param <T> Closure return type.
+     * @param <U> Closure return type.
      * @return Collection of closure results for all items in page.
      * @throws IgniteCheckedException In case of error in closure body.
      */
-    public <T> List<T> forAllItems(long pageAddr, CC<T> c) throws IgniteCheckedException {
+    public <U> List<U> forAllItems(long pageAddr, CC<U> c) throws IgniteCheckedException {
```
tkalkirill, line 398: I don't see the need to change T to U.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/IgniteWalRecoveryTest.java
```diff
@@ -1529,7 +1529,8 @@ else if (rec instanceof PageDeltaRecord) {
                             int realPageSize = data.length;
 
                             pageIO.compactPage(GridUnsafe.wrapPointer(bufPtr, realPageSize), buf, realPageSize);
-                            pageIO.compactPage(ByteBuffer.wrap(data), bufWal, realPageSize);
+                            ByteBuffer dataBuf = ByteBuffer.wrap(data).order(ByteOrder.nativeOrder());
```
tkalkirill, line 1532: Please inline

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
```diff
@@ -984,4 +982,22 @@ public static String printPage(long addr, int pageSize) {
 
         return sb.toString();
     }
+
+    /**
+     * Asserts that page type of the page stored at pageAddr matches page type of this PageIO.
+     *
+     * @param pageAddr address of a page to use for assertion
+     */
+    protected final void assertPageType(long pageAddr) {
```
tkalkirill, line 991: I don’t think this method is needed because it is called in hot code and we don’t have 100% guarantee that it will be removed when the asserts are turned off, such a check should only be where it is definitely needed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/PageIO.java
```diff
@@ -984,4 +982,22 @@ public static String printPage(long addr, int pageSize) {
...
+        assert getType(pageAddr) == getType() : "Expected type " + getType() + ", but got " + getType(pageAddr);
+    }
+
+    /**
+     * Asserts that page type of the page stored in the given buffer matches page type of this PageIO.
+     *
+     * @param buf   buffer where the page for assertion is stored
+     */
+    protected final void assertPageType(ByteBuffer buf) {
```
tkalkirill, line 1000: Same ^

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
rpuch: <T> on the method shadows with <T> on the class. It brings some confusion, so it seems that it's better to name them differently.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/IgniteWalRecoveryTest.java
rpuch: Inlined, Not sure that it became better: now we have a line wrap and so still 2 lines. Should I leave it like this, what do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/io/AbstractDataPageIO.java
tkalkirill: Let's not change the current code.

rpuch: This is an improvement. It removes some confusion that was in the code. It does not add any problems (at least I don't see them, and the compiler is happy).

Why would you like to retain the confusion?

tkalkirill (approved):

PR #2236, opened 2021-12-13
GG-34438 Upgraded log4j2 dependency to 2.15.0 to address CVE-2021-44228

antkr:
TC PR check.

### DIFF
--- a/modules/log4j2/src/main/java/org/apache/ignite/logger/log4j2/Log4J2Logger.java
+++ b/modules/log4j2/src/main/java/org/apache/ignite/logger/log4j2/Log4J2Logger.java
@@ -21,7 +21,6 @@ import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.Map;
 import java.util.UUID;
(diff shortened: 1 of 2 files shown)


modules/log4j2/src/main/java/org/apache/ignite/logger/log4j2/Log4J2Logger.java
```diff
@@ -17,11 +17,11 @@
 package org.apache.ignite.logger.log4j2;
 
 import java.io.File;
-import java.lang.reflect.Field;
 import java.net.URL;
 import java.nio.charset.Charset;
 import java.util.Map;
 import java.util.UUID;
+
```
ibelyakov, line 24: According to the code style guideline, there should be no empty lines between the imports.

antkr: Thanks, fixed with latest commit.

PR #2247, opened 2021-12-16
GG-34386 Output IgniteSystemProperties via ignite.sh

Berkof:
[GG-34386](https://ggsystems.atlassian.net/browse/GG-34386)

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
+++ b/modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
@@ -291,7 +291,7 @@ public final class IgniteSystemProperties {
      */
     @SystemProperty(value = "This property allows to override default behavior that rest processor " +
         "doesn't start on client node. If set {@code true} than rest processor will be started on client node",
-        defaults = "" + false)
+        defaults = "false")
     public static final String IGNITE_REST_START_ON_CLIENT = "IGNITE_REST_START_ON_CLIENT";
 
     /**
@@ -302,7 +302,7 @@ public final class IgniteSystemProperties {
      */
     @Deprecated
     @SystemProperty(value = "Changes output format of {@link GridRestCommand#CACHE_GET_ALL} from {k: v, ...} " +
-        "to [{\"key\": k, \"value\": v}, ...] to allow non-string keys output", defaults = "" + false)
+        "to [{\"key\": k, \"value\": v}, ...] to allow non-string keys output", defaults = "false")
     public static final String IGNITE_REST_GETALL_AS_ARRAY = "IGNITE_REST_GETALL_AS_ARRAY";
 
     /**
@@ -378,7 +378,7 @@ public final class IgniteSystemProperties {
      */
     @SystemProperty(value = "Boolean flag indicating whether GridToStringBuilder should throw RuntimeException " +
         "     * when building string representation of an object or should just print information about exception into the log " +
-        "     * and proceed", defaults = "" + false)
+        "     * and proceed", defaults = "false")
     public static final String IGNITE_TO_STRING_THROW_RUNTIME_EXCEPTION = "IGNITE_TO_STRING_THROW_RUNTIME_EXCEPTION";
 
     /**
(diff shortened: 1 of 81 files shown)


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -58,30 +154,40 @@
      * @deprecated Use {@link #IGNITE_LOG_INSTANCE_NAME}.
      */
     @Deprecated
+    @SystemProperty(value = "If this system property is present Ignite will include grid name into " +
```
ygerzhedovich: hm, should we use GridGain instead of Ignite in messages?
@sergey-chugunov-1985  WDYT?

modules/core/src/main/java/org/apache/ignite/startup/cmdline/CommandLineStartup.java
```diff
@@ -62,10 +82,50 @@
...
+
+    /** Classes with Ignite system properties. */
+    static final List<Class<?>> PROPS_CLS = new ArrayList<>(Arrays.asList(
+        IgniteSystemProperties.class,
+        ExchangeContext.class,
+        GridCacheMapEntry.class,
+        LocalDeploymentSpi.class,
+        GridCacheDatabaseSharedManager.class,
+
+        // TODO: GG-34386
```
sk0x50, line 96: this TODO refers to itself which does not seem correct.

modules/core/src/main/java/org/apache/ignite/startup/cmdline/CommandLineStartup.java
```diff
@@ -62,10 +82,50 @@
...
+        CacheContinuousQueryHandler.class,
+        OffheapReadWriteLock.class,
+        TcpCommunicationConfiguration.class
+    ));
+
     /**
      * Static initializer.
      */
     static {
+        String h2TreeCls = "org.apache.ignite.internal.processors.query.h2.database.H2Tree";
```
sk0x50, line 112: It would be nice to create a new issue that should improve this behavior and address `GridSystemProperties` too

sergey-chugunov-1985: Agree with @sk0x50 here: all sources of properties information have to be hardcoded which turns plugins support into a mess.

We need to create a ticket to support information providers of some kind.

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1418,21 +1983,28 @@
...
+        "greater than 0", type = Integer.class, defaults = "" + DFLT_TRANSACTION_TIME_DUMP_SAMPLES_PER_SECOND_LIMIT)
     public static final String IGNITE_TRANSACTION_TIME_DUMP_SAMPLES_PER_SECOND_LIMIT =
         "IGNITE_TRANSACTION_TIME_DUMP_SAMPLES_PER_SECOND_LIMIT";
 
     /**
      * Disables smart DR throttling. Default value is <code>false</code>.
      */
-    public static final String IGNITE_DISABLE_SMART_DR_THROTTLING =
-        "IGNITE_DISABLE_SMART_DR_THROTTLING";
+    @SystemProperty(value = "Disables smart DR throttling", defaults = "" + false)
```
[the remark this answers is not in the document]
Berkof: Just protect from mistakes and colour highlighting for boolean constant.

modules/core/src/main/java/org/apache/ignite/startup/cmdline/CommandLineStartup.java
Berkof: Removed

modules/core/src/main/java/org/apache/ignite/startup/cmdline/CommandLineStartup.java
Berkof: Created: https://issues.apache.org/jira/browse/IGNITE-16179

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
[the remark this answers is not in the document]
ygerzhedovich: @Berkof let's use general aproach, as I see the code use direct text representation for defaults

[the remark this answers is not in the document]
Berkof: done

PR #2250, opened 2021-12-16
GG-34423 B+Tree corruption while putAll from thin client

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-34423

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeReplaceRemoveRaceTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeReplaceRemoveRaceTest.java
@@ -17,12 +17,10 @@
 package org.apache.ignite.internal.processors.database;
 
 import java.io.Externalizable;
-import java.util.concurrent.BrokenBarrierException;
 import java.util.concurrent.CyclicBarrier;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicLong;
 import org.apache.ignite.IgniteCheckedException;
-import org.apache.ignite.IgniteException;
 import org.apache.ignite.configuration.DataRegionConfiguration;
 import org.apache.ignite.failure.FailureContext;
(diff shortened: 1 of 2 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeReplaceRemoveRaceTest.java
```diff
@@ -352,53 +347,24 @@ public TestPairTree(
...
-            tree.putx(new Pair(4, 0));
-            tree.putx(new Pair(6, 0));
-            tree.putx(new Pair(7, 0));
-
-            // Split root.
-            tree.putx(new Pair(5, 0));
-
-            // Split its left subtree.
-            tree.putx(new Pair(3, 0));
+        for (int i = 0; i < 100; i++) {
```
sergey-chugunov-1985, line 350: With number of iterations please fix javadoc as well. This number is mentioned there as well.

modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeReplaceRemoveRaceTest.java
```diff
@@ -419,4 +385,81 @@ public void testConcurrentPutRemove() throws Exception {
...
     }
+
+    /**
+     * Checks that there will be no corrupted B+tree during concurrent update and deletion
+     * of the same key that is contained in the inner and leaf nodes of the B+tree.
+     * 
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testConcurrentPutRemoveSameRow() throws Exception {
```
sergey-chugunov-1985, line 396: Do I understand correctly that this test works almost the same as the previous one? As I can see the only difference is that it tries to operate on a single key instead of two different ones.

I suggest to mention this in javadoc of the test. Lets add a phrase like that: `Test logic is the same as of <ref to another test>; the only difference is that it operates (puts and removes) on a single key.`

PR #2255, opened 2021-12-17
GG-32626: Request select where in sub select returns multiplicative data on multi-node cluster

AMashenkov:
https://ggsystems.atlassian.net/browse/GG-32626


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheSqlQuery.java
```diff
@@ -81,7 +81,7 @@
     /** Flag indicating that the query contains an OUTER JOIN from REPLICATED to PARTITIONED. */
     @GridToStringInclude
     @GridDirectTransient
-    private transient boolean hasOuterJoinReplicatedPartitioned;
+    private transient boolean trearPartitionedAsReplicated;
```
ygerzhedovich, line 84: trearPartitionedAsReplicated -> treatPartitionedAsReplicated

PR #2256, opened 2021-12-21
GG-34413 Fix node crashes with ClosedChannelException

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-34413


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/filehandle/FileWriteHandleImpl.java
```diff
@@ -500,7 +500,19 @@ private void fsync(MappedByteBuffer buf, int off, int len) throws IgniteCheckedE
...
+                        }
                     }
+                    else {
+                        if (log.isDebugEnabled()) {
+                            log.debug("Not enough space in wal segment to write segment switch, written="
+                                + written + ", switchSegmentRecSize=" + switchSegmentRecSize);
+                        }
+                    }
+
+                    flushOrWait(null);
```
tkalkirill, line 515: Please indicate why there is **null**.

SammyVimes: null is for unconditional flush, the same as 20  lines above

tkalkirill (approved):

PR #2269, opened 2022-01-12
GG-34556 Update jackson-databind version to bypass VULNDB-275302

solveme:
https://github.com/FasterXML/jackson-databind/issues/3328


aealeksandrov (approved):
LGTM

PR #2295, opened 2022-01-28
GG-30476 Added link to doc instead of site

IgGusev:


alievmirza (approved):
In general, I'm ok with the changes, one thing is that maybe it's better to use `https` instead of `http` for a link.

PR #2303, opened 2022-02-01
GG-34617 [IGNITE-16294] .NET: Fix System.Enum serialization

ptupitsyn:


isapego (approved):
Looks good to me.

modules/core/src/test/java/org/apache/ignite/platform/model/AccessLevel.java
```diff
@@ -0,0 +1,26 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
isapego, line 2: 2021

ptupitsyn: Fixed

PR #2321, opened 2022-02-08
GG-33888 [IGNITE-15504] .NET: Modernize project structure

ptupitsyn:
* Update all projects (csproj files) to use SDK-style format https://docs.microsoft.com/en-us/dotnet/core/project-sdk/overview#project-files
* Remove nuspec files, use csproj properties instead.
* Target netstandard2.0 in modern libraries (Core, Linq, loggers).
* Target net461 in legacy libraries (EF, ASP.NET).
* Target netcoreapp3.1 in examples (lowest supported .NET Core version at the moment).
* Use dotnet CLI to build binaries and packages: this works on any OS, even for legacy Windows-only libraries.
* Delete legacy NuGet tests.
* Update build.ps1 - remove complicated logic for nuget and binaries, use simple dotnet CLI commands.

As a result:
* Nothing should change for the users.
* Visual Studio is not required.
* The only requirement is .NET Core SDK 3.1.
* Build procedure is trivial and works on any OS.
* Future SDK upgrades are trivial.


isapego (approved):
Looks good to me.

PR #2324, opened 2022-02-09
GG-34468 BPlusTreeSelfTest#testMassiveRemove2_false flaky

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-34468

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
@@ -4814,6 +4814,9 @@ public abstract class BPlusTree<L, T extends L> extends DataStructure implements
                         freePage(tail.pageId, tail.page, tail.buf, tail.walPlc, true);
 
                         tail = tail.down;
+
+                        assert tail.sibling == null : tail;
+
                         cutRoot = true;
 
                         // Exit: we are done.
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
```diff
@@ -4802,14 +4802,21 @@ private boolean releaseForRetry(Tail<L> t) {
...
+                        cutRoot(t.lvl);
+                        freePage(t.pageId, t.page, t.buf, t.walPlc, false);
+
+                        t = t.down;
 
                         // Exit: we are done.
                     }
-                    else if (tail.sibling != null &&
+
+                    if (tail.sibling != null &&
```
ibessonov, line 4819: `tail` can point on removed page here, isn't it? I bet there's a bug here. Why do you need additional variable `t` in the loop above? You could just use `tail` and a `while` loop instead, this way `tail` value would be correct. Maybe I'm wrong here, please correct me if that's the case.

tkalkirill: Correct it

ibessonov (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
```diff
@@ -4802,14 +4802,27 @@ private boolean releaseForRetry(Tail<L> t) {
 
                     assert needReplaceInner != TRUE;
 
-                    if (tail.getCount() == 0 && tail.lvl != 0 && getRootLevel() == tail.lvl) {
+                    boolean cutRoot = false;
```
ibessonov, line 4805: Please remove this variable, we don't need ot

ibessonov (approved):

PR #2355, opened 2022-03-02
GG-34794: Fix B-Tree corruption when client creates cache with affinity field.

AMashenkov:


AMashenkov:
Correct fix is to backport GG-31261 [IGNITE-5795] Binary metadata is registering during start of cache

PR #2362, opened 2022-03-11
GG-34882 [IGNITE-16640] Peer class loading failure should not be treated as a critical node failure

rpuch:
https://ggsystems.atlassian.net/browse/GG-34882


SammyVimes (approved):
LGTM

PR #2368, opened 2022-03-15
GG-34902 [IGNITE-16549] .NET: Add IgniteClientConfiguration.SendServerExceptionStackTraceToClient

ptupitsyn:
* Propagate SendServerExceptionStackTraceToClient to .NET
* Java fix: Include stack trace for Compute errors
* Java fix: Include stack trace even when cause is null


isapego (approved):
Looks good to me.

PR #2371, opened 2022-03-17
GG-34919 Fix sqlline.bat windows shell scripts

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-34919


tledkov (commented):
OK with me

PR #2374, opened 2022-03-21
GG-34872 [IGNITE-16530] Java thin: add heartbeats

ptupitsyn:
Implement heartbeats in Java thin client: https://cwiki.apache.org/confluence/display/IGNITE/IEP-83+Thin+Client+Keepalive

  * `ClientConfiguration.heartbeatsEnabled`, default `false`.
  * `ClientConfiguration.heartbeatInterval`, default 30 seconds.
  * When heartbeats are enabled, get idle timeout from server, set effective heartbeat interval to `Math.min(heartbeatInterval, idleTimeout / 3)`.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/configuration/ClientConfiguration.java
+++ b/modules/core/src/main/java/org/apache/ignite/configuration/ClientConfiguration.java
@@ -129,6 +129,12 @@ public final class ClientConfiguration implements Serializable {
     /** Executor for async operations continuations. */
     private Executor asyncContinuationExecutor;
 
+    /** Whether heartbeats should be enabled. */
+    private boolean heartbeatEnabled;
+
+    /** Heartbeat interval, in milliseconds. */
+    private long heartbeatInterval = 30_000L;
+
     /**
      * @return Host addresses.
      */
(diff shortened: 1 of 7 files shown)


isapego (approved):

pkgonan:
@ptupitsyn @isapego 
Hi. I'm using the GridGain Ultimate License, and I'm thinking of posting a question through the Support Portal, but I'm asking because it would be better to contact the developers directly.

The application developed by us is accessing the GridGain Cluster using a thin client. If there is no request for a certain period of time in the application, the connection is cut off and an error occurs. I think this feature was developed to solve this problem.

I have a question.

First. When will 8.8.18 be released? We are waiting for this feature now.

Second. Can Java Thin HeartBeat function solve the above connection problem?

Third. In order to solve the above problem, should both the version of GridGain Cluster and the version of GridGain Thin Client be upgraded to 8.8.18? Or can GridGain Cluster use the current 8.8.17 as it is and upgrade only the version of GridGain Thin Client to 8.8.18?

pkgonan:
@ptupitsyn @isapego 
Hi. May I ask you to answer the above question? Thank you :)

ptupitsyn:
@pkgonan please forward those questions to the GridGain support team.

PR #2375, opened 2022-03-21
GG-34798 Fix RU does not work for nodes with TDE configured

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-34798

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/IgniteNodeAttributes.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/IgniteNodeAttributes.java
@@ -195,8 +195,8 @@ public final class IgniteNodeAttributes {
     /** User authentication enabled flag. */
     public static final String ATTR_AUTHENTICATION_ENABLED = ATTR_PREFIX + ".authentication.enabled";
 
-    @Deprecated
     /** Encryption master key digest. Only needed for compatibility with 8.7.X versions. */
+    @Deprecated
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/IgniteNodeAttributes.java
```diff
@@ -195,6 +195,10 @@
     /** User authentication enabled flag. */
     public static final String ATTR_AUTHENTICATION_ENABLED = ATTR_PREFIX + ".authentication.enabled";
 
+    @Deprecated
+    /** Encryption master key digest. Only needed for compatibility with 8.7.X versions. */
```
sashapolo, line 199: Please move the comment above the annotation, IDEA complains about it

modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -247,6 +248,9 @@ public GridEncryptionManager(GridKernalContext ctx) {
     @Override public void start() throws IgniteCheckedException {
         startSpi();
 
+        if (getSpi().masterKeyDigest() != null)
```
sashapolo, line 251: I think we should extract `getSpi().masterKeyDigest()` into a variable

sashapolo: Also, we should add a comment that this is for compatibility reasons

sashapolo (approved):

PR #2376, opened 2022-03-23
GG-34948 [IGNITE-16637] Update lucene version to 8.11.1

solveme:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridLuceneFile.java
```diff
@@ -105,6 +116,7 @@ void lockRef() {
     void releaseRef() {
         refCnt.decrementAndGet();
 
+        // TODO: this may trigger actual file deletion but looks like filename would still stay in GridLuceneDirectory.fileMap
```
ygerzhedovich, line 119: Let's create separate ticket to check that

solveme: Done

ygerzhedovich (approved):

PR #2377, opened 2022-03-23
GG-34694 Fix IgniteClientConnectionLimit flakiness

isapego:
Flakiness is caused by random appearance of messages of format `Client disconnected due to an error ...` which causes client disconnection be counted twice.

### DIFF
--- a/modules/platforms/cpp/thin-client-test/include/vector_logger.h
+++ b/modules/platforms/cpp/thin-client-test/include/vector_logger.h
@@ -67,7 +67,7 @@ namespace ignite_test
             const std::string& nativeErrorInfo)
         {
             if ((categoryFilter.empty() || category.find(categoryFilter) != std::string::npos) &&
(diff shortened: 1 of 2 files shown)


modules/platforms/cpp/thin-client-test/include/vector_logger.h
```diff
@@ -67,7 +67,7 @@ namespace ignite_test
             const std::string& nativeErrorInfo)
         {
             if ((categoryFilter.empty() || category.find(categoryFilter) != std::string::npos) &&
-                (messageFilter.empty() || message.find(messageFilter) != std::string::npos))
+                (!messageFilter.empty() && message.find(messageFilter) != std::string::npos))
```
ptupitsyn, line 70: Can you please explain the change? Why the second condition is different from the first one?

ptupitsyn (approved):

PR #2385, opened 2022-03-29
GG-34975: DR: Wrong update counter is written to entry on backup node in TX.

AMashenkov:
Fix the issue when 'excplicitVersion' object could be shared between different GridCacheMapEntries objects.
Restore correct partition counter to GridCacheVersion before writing to WAL on primary and backup nodes.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxEntry.java
```diff
@@ -558,6 +561,20 @@ public boolean addReader() {
...
 
+    /**
+     * @param noop Add no-op flag.
+     */
+    public void noop(boolean noop) {
+        setFlag(noop, TX_ENTRY_NOOP_ON_PRIMARY);
+    }
+
+    /**
+     * @return Add no-op flag.
```
SammyVimes, line 572:
```suggestion
     * @return {@code true} if noop flag is set, {@code false} otherwise.
```

SammyVimes (approved):
LGTM except one little comment

PR #2390, opened 2022-04-07
GG-35019 DR: Add an option to ignore duplicates in PartitionLogTree.

tledkov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/tree/updatelog/PartitionLogTree.java
```diff
@@ -108,7 +117,17 @@ public PartitionLogTree(
 
         cmp = Long.compare(updCntr, row.updCntr);
 
-        assert cmp != 0 || row.link == 0 /* search insertion poin */ || io.getLink(pageAddr, idx) == row.link /* remove row */;
+        /* remove row */
+        if (cmp == 0 && row.link != 0 /* search insertion poin */ && io.getLink(pageAddr, idx) != row.link) {
+            if (IgniteSystemProperties.getBoolean(IgniteSystemProperties.IGNITE_STRICT_CONSISTENCY_CHECK))
```
AMashenkov, line 122: Let's save the value to a boolean field.

AMashenkov (approved):

PR #2406, opened 2022-04-19
GG-35110 Make metric command to work with secured cluster

solveme:


modules/core/src/main/java/org/apache/ignite/internal/visor/metric/VisorMetricTask.java
```diff
@@ -69,6 +70,8 @@ protected VisorMetricJob(@Nullable VisorMetricTaskArg arg, boolean debug) {
             if (arg == null)
                 throw new IgniteException("VisorMetricTaskArg is null");
 
+            ignite.context().security().authorize(ADMIN_VIEW);
```
sk0x50, line 73: Do we really need this check here? In my understanding, visor task should be already authorized before running.

modules/control-utility/src/test/java/org/apache/ignite/util/GridCommandHandlerMetricTest.java
```diff
@@ -42,10 +42,10 @@
 import static org.apache.ignite.testframework.GridTestUtils.assertContains;
 
 /** Tests output of {@link CommandList#METRIC} command. */
-public class MetricCommandTest extends GridCommandHandlerClusterByClassAbstractTest {
+public class GridCommandHandlerMetricTest extends GridCommandHandlerClusterByClassAbstractTest {
```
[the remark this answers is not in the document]
solveme: In this case I think it would be easier to get back to MetricCommandTest, to have same naming in both projects

modules/core/src/main/java/org/apache/ignite/internal/visor/metric/VisorMetricTask.java
solveme: honestly I have no idea where it should be, I made it by analogy with other tasks

modules/core/src/main/java/org/apache/ignite/internal/visor/metric/VisorMetricTask.java
```diff
@@ -35,6 +35,7 @@
 import org.jetbrains.annotations.Nullable;
 
 import static org.apache.ignite.internal.processors.metric.impl.MetricUtils.SEPARATOR;
+import static org.apache.ignite.plugin.security.SecurityPermission.ADMIN_VIEW;
```
sk0x50, line 38: please remove unused import

PR #2417, opened 2022-05-11
GG-34893 Extract H2Tree comparator and make it versioned

zstan:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -155,6 +156,8 @@ public class H2Tree extends BPlusTree<H2Row, H2Row> {
     /** Whether index was created from scratch during owning node lifecycle. */
     private final boolean created;
 
+    private final boolean useFixedComparator;
```
ygerzhedovich, line 159: Let's add a comment reason and what does mean fixed and not fixed comparator

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -252,6 +255,21 @@ public H2Tree(
...
+
+            if (idxCreateVer != null) {
+                if ((idxCreateVer.compareTo(ver_8_7_40) >= 0 && idxCreateVer.minor() == 7) ||
+                    idxCreateVer.compareTo(ver_8_8_11) >= 0)
+                    useFixedComparator = true;
+                else
+                    useFixedComparator = false;
+            }
+            else
+                useFixedComparator = false;
```
AMashenkov, line 271:
```suggestion
            useFixedComparator = idxCreateVer != null &&(idxCreateVer.compareTo(ver_8_7_40) >= 0 && idxCreateVer.minor() == 7) ||
                    idxCreateVer.compareTo(ver_8_8_11) >= 0);                   
```

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2ValueCacheObject.java
```diff
@@ -180,6 +183,22 @@ public Object getObject(boolean cpy) {
...
 
+    /**
+     * Stub GG-33962, GG-34893 (new comparator) workaround.
+     * Versions < 8.7.40 and < 8.8.11 are used incorrect comparator and a simple comparator change
+     * leads to the tree corruprion. Thus for "old" versions is used appropriate old comparator,
+     * we take into account that later all indexes will be rebuild and only one ver of comparator
+     * will be used.
+     */
+    public void useCorrectComparator() {
+        useFixedComparator = true;
```
AMashenkov, line 194: Using different names for the method and the field look a bit confusing.
Will the next be better? WDYT?
```suggestion
    public void useLagacyComparator() {
        useLagacyComparator= true;
```

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
zstan: I found my approach is more readable as for me, i wan`t fix it ..

AMashenkov: `useFixedComparator = false` is a default value, thus similar 'else' branches looks too verbose.
What about the next way?
```
 if (idxCreateVer == null /* too old version */ ||
     idxCreateVer.compareTo(ver_8_7_40) < 0 || 
     idxCreateVer.minor() != 7 && idxCreateVer.compareTo(ver_8_8_11) < 0)
            useFixedComparator = false; // Fallback to legacy comparator due to compatibility reasons.
else 
     useFixedComparator = true;
```

zstan: done thanks

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2ValueCacheObject.java
AMashenkov: Still not clear.
Why do anyone want to use 'incorrect' comparator?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2ValueCacheObject.java
```diff
@@ -180,6 +183,24 @@ public Object getObject(boolean cpy) {
...
+     * leads to the tree corruprion. Thus for "old" versions is used appropriate old comparator,
+     * we take into account that later all indexes will be rebuild and only one ver of comparator
+     * will be used.
+     */
+    public void useLegacyComparator() {
+        useLegacyComparator = true;
+    }
+
+    /**
+     * @return {@code false} if legacy comparator is used.
```
AMashenkov, line 198:
```suggestion
     * @return {@code true} if legacy comparator is used, {@code false} otherwise.
```

AMashenkov (approved):

PR #2431, opened 2022-06-02
GG-35225 Enable keepBinary by default for sqlline

tledkov:


modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1027,6 +1028,12 @@ else if (res.status() != ClientListenerResponse.STATUS_SUCCESS)
 
                     throw e;
                 }
+                catch (BinaryObjectException e) {
+                    if (LOG.isLoggable(Level.FINE))
+                        LOG.log(Level.FINE, "Exception during sending an sql request.", e);
```
ygerzhedovich, line 1033: Maybe log possible solution here?

tledkov-gridgain: Not catch an idea. Add to message info about binary mode?

ygerzhedovich: Hmmm, may be you are right and add info to exception will be more good solution. Seems will be good to add solutions like a ''Probably you use specific non JDBC types, in this case try to set keepBinary=true for connection"

tledkov-gridgain: Fixed.

PR #2434, opened 2022-06-03
GG-35189 DiscoveryClientSocketTest.sslSocketTest fails regularly on TC and locally

rpuch:
https://ggsystems.atlassian.net/browse/GG-35189

This is caused by two problems:

1. Some JDK versions (including 1.8.0_261) contain a bug https://bugs.openjdk.java.net/browse/JDK-8241239 due to which close() on a socket is blocked when a write() is blocked on its output stream. This causes the test to always fail on such JDKs. The test models a situation which may be faced by our customers, so there is no point in working this around just in our test.
2. If the original 'handshake' takes long time (I saw it around 1 second), then the test may block on a write attempt for a long time (handshake time multiplied by 10), in my case 10 seconds, which makes the test fail.

This commit can't do anything with item 1, but it alleviates item 2: write wait time is bounded.


SammyVimes (approved):
LGTM!

PR #2443, opened 2022-06-10
GG-35238 Test fixed: different API is used to trigger writing binary metadata to avoid flaky failures

sergey-chugunov-1985:


rpuch (approved):

rpuch:
The patch looks good to me

tkalkirill (approved):

PR #2447, opened 2022-06-14
GG-35386 Fix flaky IgniteCacheMessageRecoveryIdleConnectionTest.testCacheOperationsIdleConnectionCloseTx

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-35386

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
@@ -2847,27 +2847,11 @@ public class GridNioServer<T> {
 
                 IOException err = new IOException("Failed to send message (connection was closed): " + ses);
 
-                try {
-                    filterChain.onSessionClosed(ses);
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -2878,6 +2862,19 @@ protected boolean close(final GridSelectorNioSessionImpl ses, @Nullable final Ig
...
                 }
 
+                try {
+                    filterChain.onSessionClosed(ses);
+                }
+                catch (IgniteCheckedException e1) {
+                    filterChain.onExceptionCaught(ses, e1);
+                }
+
+                if (outRecovery != null)
```
sergey-chugunov-1985, line 2872: As I can see the old version of code releases in and out recovery descriptors in finally block around cleaning up message requests.

In new version of code no finally block is protecting us from any exception there.

Do you see any risks here?

PR #2450, opened 2022-06-15
GG-34787 Fixed broken link in Javadoc

IgGusev:
https://ggsystems.atlassian.net/browse/GG-34787

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/IgniteLogger.java
+++ b/modules/core/src/main/java/org/apache/ignite/IgniteLogger.java
@@ -22,7 +22,7 @@ import org.apache.ignite.internal.util.tostring.GridToStringExclude;
  * This interface defines basic logging functionality used throughout the system. We had to
  * abstract it out so that we can use whatever logging is used by the hosting environment.
  * Currently, <a target=_new href="http://logging.apache.org/log4j/1.2/">log4j</a>,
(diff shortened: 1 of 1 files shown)


rpuch (approved):

modules/core/src/main/java/org/apache/ignite/IgniteLogger.java
```diff
@@ -22,7 +22,7 @@
  * This interface defines basic logging functionality used throughout the system. We had to
  * abstract it out so that we can use whatever logging is used by the hosting environment.
  * Currently, <a target=_new href="http://logging.apache.org/log4j/1.2/">log4j</a>,
- * <a target=_new href="http://docs.jboss.org/hibernate/orm/4.3/topical/html/logging/Logging">JBoss</a>,
+ * <a target=_new href="https://docs.jboss.org/hibernate/orm/4.3/manual/en-US/html_single/#configuration-logging">JBoss</a>,
```
rpuch, line 25: I think this link should be used instead: https://docs.jboss.org/hibernate/orm/5.4/topical/html_single/logging/Logging.html because it it about JBoss Logging.

The link you suggest talks about slf4j, while the original was about JBoss Logging: https://web.archive.org/web/20190404193125/http://docs.jboss.org/hibernate/orm/4.3/topical/html/logging/Logging

rpuch (changes requested):

rpuch (approved):

PR #2451, opened 2022-06-15
GG-35331 Fixed a link to java Fork/Join framework

IgGusev:
https://ggsystems.atlassian.net/browse/GG-35331

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/collision/jobstealing/JobStealingCollisionSpi.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/collision/jobstealing/JobStealingCollisionSpi.java
@@ -69,7 +69,7 @@ import static org.apache.ignite.events.EventType.EVT_NODE_LEFT;
  * from slower node and moved to the fast under-utilized node.
  * <p>
  * The design and ideas for this SPI are significantly influenced by
- * <a href="http://gee.cs.oswego.edu/dl/papers/fj.pdf">Java Fork/Join Framework</a>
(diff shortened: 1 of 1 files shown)


rpuch (approved):

modules/core/src/main/java/org/apache/ignite/spi/collision/jobstealing/JobStealingCollisionSpi.java
```diff
@@ -69,7 +69,7 @@
  * from slower node and moved to the fast under-utilized node.
  * <p>
  * The design and ideas for this SPI are significantly influenced by
- * <a href="http://gee.cs.oswego.edu/dl/papers/fj.pdf">Java Fork/Join Framework</a>
+ * <a href="https://docs.oracle.com/javase/tutorial/essential/concurrency/forkjoin.html">Java Fork/Join Framework</a>
```
sk0x50, line 72: I don't think that this change is the right one. The previous link is about the original Doug Lea's paper "A Java Fork/Join Framework". Perhaps, this link is just blocked from RU region.

IgGusev: I am not in RU region and still cannot access it. It also fails tests, which check outside of RU.

sk0x50: Oh... Looks like the link is really broken...

PR #2459, opened 2022-06-20
GG-35362 DR: Incremental active-active erroneously replicate entries back to DC where they have been replicated through FST

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/dr/GridDrType.java
```diff
@@ -33,5 +33,8 @@ public enum GridDrType {
     DR_LOAD,
 
     /** Replication during preload. */
-    DR_PRELOAD
-}
\ No newline at end of file
+    DR_PRELOAD,
+
+    /** Entry initiated in current dc. */
+    DR_NATIVE
```
AMashenkov, line 39: DR_IGNORE

AMashenkov, line 39:
```suggestion
    /** Got back an entry, which was created in current dc, during full state transfer. */
    DR_IGNORE
```

AMashenkov (approved):

AMashenkov (approved):

PR #2465, opened 2022-06-23
GG-34640 skip authentication for Probe and Version commands

dborunov-gg:
https://ggsystems.atlassian.net/browse/GG-34640


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/GridRestProcessor.java
```diff
@@ -111,6 +114,9 @@ public class GridRestProcessor extends GridProcessorAdapter {
     private static final String HTTP_PROTO_CLS =
         "org.apache.ignite.internal.processors.rest.protocols.http.jetty.GridJettyRestProtocol";
 
+    /** Commands, that are not required to be authenticated. */
```
nva, line 117:
```suggestion
    /** Commands that do not require authentication. */
```

nva (approved):

PR #2468, opened 2022-06-24
GG-35236 Log validation errors with DEBUG when processing snapshot-utility commands

rpuch:
https://ggsystems.atlassian.net/browse/GG-35236

Some exceptions are thrown to indicate that something is wrong with a user command, but they do not indicate that the system is in trouble. Such exceptions should not be logged with high levels in the core system. To distinguish them from other exceptions, we add special exception types.


SammyVimes (approved):
LGTM!

PR #2488, opened 2022-07-13
GG-35422 Java thin: Fix hang on deprecated SSL protocols

ptupitsyn:
TLSv1.1 is deprecated. It causes an exception in `GridNioSslFilter.onSessionOpened`. However, `GridNioClientConnectionMultiplexer` waits on `sesFut` first, which never completes in this case.

* Fix `GridNioClientConnectionMultiplexer` to wait on `sslHandshakeFut` first and fail on SSL exceptions instead of hanging.
* Fix `GridNioSslFilter` to propagate exception details correctly.
* Fix tests to use TLSv1.2 where needed.


isapego (approved):
Looks great!

PR #2502, opened 2022-07-29
GG-35556 Update SSL certs and keys for tests

isapego:
1. Re-generated SSL certs and keys;
2. Fixed generation scripts (they were not working with later `keytool` and did not generate all the files).


modules/core/src/main/java/org/apache/ignite/internal/util/nio/ssl/GridNioSslFilter.java
```diff
@@ -216,8 +213,12 @@ public boolean directMode() {
         try {
             GridNioFutureImpl<?> fut = ses.removeMeta(HANDSHAKE_FUT_META_KEY);
 
-            if (fut != null)
+            if (fut != null) {
```
ptupitsyn, line 216: I'm not sure if this condition is relevant to `rejectedSesCnt`.

ptupitsyn (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/util/nio/ssl/GridNioSslFilter.java
isapego: Tests show that it is at least more proper way to count them that it is done now.
But I agree, this needs additional investigation and maybe additional testing. Moved to a separate ticket (GG-35573) and removed these edits from PR.

ptupitsyn (approved):

PR #2510, opened 2022-08-09
GG-35462 DR: Support of rebuild DR b-tree through control.sh

AMashenkov:


modules/core/src/main/java/org/apache/ignite/internal/visor/dr/VisorDrRebuildTreeOperation.java
```diff
@@ -0,0 +1,29 @@
...
+ * limitations under the License.
+ */
+
+package org.apache.ignite.internal.visor.dr;
+
+/**
+ * Dr rebuild tree sub-commands.
+ */
+public enum VisorDrRebuildTreeOperation {
+    /** Default command */
```
korlov42, line 23: missed period at the end

korlov42 (approved):

PR #2514, opened 2022-08-18
GG-35591 Extend GridQueryStartedInfo, GridQueryFinishedInfo.

nva:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryFinishedInfo.java
```diff
@@ -45,12 +46,28 @@ public class GridQueryFinishedInfo {
     /** */
     private final long finishTime;
 
+    /** Query cancellable flag. */
```
ygerzhedovich, line 49: what reason to have the field?

nva: Looks like you're right. Deleted this field.

ygerzhedovich (approved):

PR #2517, opened 2022-08-22
GG-35450 C++: Java remote filters for CQ for thin client

isapego:
- Implemented Java remote filter for thin client.
- Added 2 tests.

### DIFF
--- a/modules/platforms/cpp/thin-client/include/ignite/thin/cache/event/java_cache_entry_event_filter.h
+++ b/modules/platforms/cpp/thin-client/include/ignite/thin/cache/event/java_cache_entry_event_filter.h
@@ -53,7 +53,7 @@ namespace ignite
                  */
                 class JavaCacheEntryEventFilter
                 {
-                    friend ignite::impl::thin::cache::CacheClientImpl;
+                    friend class ignite::impl::thin::cache::CacheClientImpl;
                 public:
                     /**
                      * Default constructor.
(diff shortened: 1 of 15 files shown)


modules/platforms/cpp/thin-client/include/ignite/thin/cache/event/java_cache_entry_event_filter.h
```diff
@@ -0,0 +1,129 @@
...
+                     * Add property.
+                     *
+                     * Template argument type should be copy-constructable and assignable. Also BinaryType class template
+                     * should be specialized for this type.
+                     *
+                     * @param name Property name.
+                     * @param value Property value.
+                     */
+                    template<typename T>
+                    void SetProperty(const std::string& name, const T& value)
```
ptupitsyn, line 106: We can set properties, but not get them - is this intentional?

ptupitsyn (approved):

modules/platforms/cpp/thin-client/include/ignite/thin/cache/event/java_cache_entry_event_filter.h
isapego: Well, yeah. I can add a method for getting property, but it will be hard to implement right and I don't think there are a lot of use cases where we'd like to get properties from such an object anyway.

PR #2525, opened 2022-09-05
GG-35677: Spark 3.2 module was added

ivgag:


aealeksandrov:
I don't see the examples for the new Spark? Do we plan to add it there?

PR #2531, opened 2022-09-15
GG-35521 Fixed an issue that could lead to B+Tree corruption during attempt to write a large entry to atomic cache

sk0x50:


modules/core/src/test/java/org/apache/ignite/internal/processors/database/IgniteDbPutGetAbstractTest.java
```diff
@@ -68,6 +80,24 @@ private IgniteEx ig() {
...
+    @Override protected IgniteConfiguration getConfiguration(String gridName) throws Exception {
+        IgniteConfiguration cfg = super.getConfiguration(gridName);
+
+        if (getTestIgniteInstanceIndex(gridName) == smallWalBufSizeNodeIdx) {
+            cfg.getDataStorageConfiguration()
+                .setWalBufferSize(DFLT_WAL_SEGMENT_SIZE / 4)
+                .setWalSegmentSize(DFLT_WAL_SEGMENT_SIZE / 4);
+        }
+
+        cfg.setFailureHandler((ignite, failureCtx) -> {
```
ibessonov, line 93: Don't we have `getFailureHandler()` in abstract test? This section can be moved to that specific method.

modules/core/src/test/java/org/apache/ignite/internal/processors/database/IgniteDbPutGetAbstractTest.java
```diff
@@ -271,6 +301,56 @@ public void testPutGetLarge() throws Exception {
...
+
+            // Primary node should be updated successfully,
+            // however, backup node should fail because of size of the new entry does not allow to write it to WAL.
+            assertThrows(
+                log,
+                () -> atomicCache.put(atomicBackupKey, newVal),
+                CacheException.class,
+                "Failed to update keys");
+
+            assertTrue(
```
ibessonov, line 342: Weird, there should be methods to compare arrays

modules/core/src/test/java/org/apache/ignite/internal/processors/database/IgniteDbPutGetAbstractTest.java
```diff
@@ -271,6 +301,56 @@ public void testPutGetLarge() throws Exception {
...
+                "Unexpected value.",
+                Arrays.equals((byte[])primaryNode.cache(atomicCache.getName()).get(atomicBackupKey), newVal));
+
+            assertTrue(
```
ibessonov, line 346: As well as here, you could write `assertThat(a, contains(b))` or something like that

sk0x50: Wow, `hamcrest` 2.x is available in Ignite codebase! Will use it.

modules/core/src/test/java/org/apache/ignite/internal/processors/database/IgniteDbPutGetAbstractTest.java
sk0x50: Nice catch! Will change of course.

ibessonov (approved):

PR #2536, opened 2022-09-17
GG-35721 Split IgnitePdsWithIndexingCoreTestSuite in two

rpuch:
https://ggsystems.atlassian.net/browse/GG-35721


rpuch:
This is irrelevant as the suite was timing out due to a hanging test (which is already fixed)

PR #2538, opened 2022-09-19
GG-35627 SQL: Do not yield separate query result for comments.

AMashenkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1228,6 +1228,9 @@ private void checkClusterState(QueryParserResult parseRes) {
 
                         assert cmd != null;
 
+                        if (cmd.noOp() && parseRes.remainingQuery() == null && parseRes.queryDescriptor().sql().isEmpty())
```
ygerzhedovich, line 1231:
```suggestion
                        if (cmd.noOp() && remainingQry == null && newQryDesc.sql().isEmpty())
```

PR #2590, opened 2022-10-05
GG-35682 SQL: improve IN statement execution in H2.

xtern:
Currently, if we have "IN" statement, we only use the index cursor when the "IN()" condition is set to the first column of the index.

For example we have
```
TABLE TAB1(int A, int B)
INDEX IDX1(A,B)
```
Example 1. Index cursor query example:
```
select * from TAB1 where A in (1,2) and B = 3

will be executed as follows

find(A=1)
 filter(B=3)
find(A=2)
 filter (B=3)
```
Example 2. Not index query cursor
```
select * from TAB1 where A = 3 and B (1,2)

will be executed as follows

find(A=3)
 filter(B in 1,2)
```

So the main idea behind this fix is that we can use the index cursor (for the "IN()" operator) not only for the first column, if all other conditions are "equal-constant" expressions.

```
select * from TAB1 where A in (1,2) and B = 3

should be executed as follows

find(A=1, B=3)
find(A=2, B=3)

select * from TAB1 where A = 3 and B (1,2)

should be executed as follows
find(A=3, B=1)
find(A=3, B=2)
```

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
@@ -1497,7 +1497,7 @@ public abstract class BPlusTree<L, T extends L> extends DataStructure implements
      * @return Found result or {@code null}.
      * @throws IgniteCheckedException If failed.
      */
-    public <R> R findOne(L row, TreeRowClosure<L, T> c, Object x) throws IgniteCheckedException {
+    public final <R> R findOne(L row, TreeRowClosure<L, T> c, Object x) throws IgniteCheckedException {
         checkDestroyed();
 
         GetOne g = new GetOne(row, c, x, false);
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
@@ -1497,7 +1497,7 @@ public abstract class BPlusTree<L, T extends L> extends DataStructure implements
      * @return Found result or {@code null}.
      * @throws IgniteCheckedException If failed.
      */
-    public final <R> R findOne(L row, TreeRowClosure<L, T> c, Object x) throws IgniteCheckedException {
+    public <R> R findOne(L row, TreeRowClosure<L, T> c, Object x) throws IgniteCheckedException {
         checkDestroyed();
 
         GetOne g = new GetOne(row, c, x, false);
--- a/modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
+++ b/modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
@@ -262,25 +262,23 @@ public class BasicIndexTest extends AbstractIndexingCommonTest {
     public void testInStatementUsesIndex() throws Exception {
         H2TreeIndex.h2TreeFactory = TestH2Tree::new;
 
-        inlineSize = 20;
+        inlineSize = 30;
 
         IgniteEx ig0 = startGrid();
         GridQueryProcessor qryProc = ig0.context().query();
 
-        String create = "CREATE TABLE PUBLIC.TEST_TABLE (f1 int, f2 int, f3 int, f4 int, CONSTRAINT PK_IDX PRIMARY KEY (f1, f2, f3))";
-        String createIdx = "CREATE INDEX PK_IDX ON PUBLIC.TEST_TABLE (f1, f2) INLINE_SIZE 30";
-        String insert = "INSERT INTO PUBLIC.TEST_TABLE (f1, f2, f3, f4) values (%d, %d, %d, 2)";
-
         Function<String, FieldsQueryCursor<List<?>>> execSql =
-                sql0 -> qryProc.querySqlFields(new SqlFieldsQuery(sql0), true);
+            sql0 -> qryProc.querySqlFields(new SqlFieldsQuery(sql0), true);
 
-        execSql.apply(create);
-        execSql.apply(createIdx);
+        execSql.apply("CREATE TABLE PUBLIC.TEST_TABLE (f1 int, f2 int, f3 int, f4 int, PRIMARY KEY (f1, f2, f3))");
+        execSql.apply("CREATE INDEX PK_IDX ON PUBLIC.TEST_TABLE (f1, f2, f3)");
+
+        String sqlInsert = "INSERT INTO PUBLIC.TEST_TABLE (f1, f2, f3, f4) values (%d, %d, %d, 2)";
 
         for (int i = 0; i < 10; ++i)
             for (int j = 0; j < 10; ++j)
                 for (int k = 0; k < 10; ++k)
-                    execSql.apply(String.format(insert, i, j, k));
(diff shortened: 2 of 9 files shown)


modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -101,6 +119,12 @@ public class BasicIndexTest extends AbstractIndexingCommonTest {
     /** Default table name. */
     private static final String TEST_TBL_NAME = "PUBLIC.TEST_TABLE";
 
+    /** The number of cursor scans. */
+    private static final AtomicInteger scanCntr = new AtomicInteger();
```
zstan, line 123: violates naming convention, need upper case.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -2393,4 +2459,68 @@ private enum SqlDataType {
             this.javaType = javaType;
         }
     }
+
+    /** */
+    static class TestH2Tree extends H2Tree {
```
zstan, line 2464: This approach looks good, but how can we be sure that scanCntr correlates with /* scanCount: 2 */ from issue ?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -229,6 +257,44 @@ public void testNoIndexesNoPersistence() throws Exception {
...
+
+        String sqlInsert = "INSERT INTO PUBLIC.TEST_TABLE (f1, f2, f3, f4) values (%d, %d, %d, 2)";
+
+        for (int i = 0; i < 10; ++i)
+            for (int j = 0; j < 10; ++j)
+                for (int k = 0; k < 10; ++k)
+                    execSql.apply(String.format(sqlInsert, i, j, k));
+
+        Consumer<String> execAssert = sql -> {
+            scanCntr.set(0);
```
zstan, line 284: seems you need to reset counter after each test ?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -229,6 +257,44 @@ public void testNoIndexesNoPersistence() throws Exception {
...
+            FieldsQueryCursor<List<?>> res =
+                execSql.apply(sql);
+
+            assertEquals(4, res.getAll().size());
+            assertEquals(4, scanCntr.get());
+        };
+
+        execAssert.accept("SELECT * FROM PUBLIC.TEST_TABLE WHERE f1 = 8 and f2 in (1, 5, 3, 7) and f3 = 5");
+        execAssert.accept("SELECT * FROM PUBLIC.TEST_TABLE WHERE f1 in (1, 5, 3, 7) and f2 = 8 and f3 = 5");
+        execAssert.accept("SELECT * FROM PUBLIC.TEST_TABLE WHERE f1 = 8 and f2 = 5 and f3 in (1, 5, 3, 7)");
```
zstan, line 295: plz append additionally tests for :
`"SELECT * FROM PUBLIC.TEST_TABLE WHERE f1 = 8 and f2 in (1) and f3 in (1, 5, 3, 7)"`
`"SELECT * FROM PUBLIC.TEST_TABLE WHERE f1 = 8 and f2 in (1, 5) and f3 in (1, 5, 3, 7)"
"SELECT * FROM PUBLIC.TEST_TABLE WHERE f1 in (1, 5) and f2 = 5 and f3 in (1, 5, 3, 7)"`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
```diff
@@ -1497,7 +1497,7 @@ public final <R> R findOne(L row, Object x) throws IgniteCheckedException {
      * @return Found result or {@code null}.
      * @throws IgniteCheckedException If failed.
      */
-    public final <R> R findOne(L row, TreeRowClosure<L, T> c, Object x) throws IgniteCheckedException {
+    public <R> R findOne(L row, TreeRowClosure<L, T> c, Object x) throws IgniteCheckedException {
```
zstan, line 1500: seems we needn`t touch this code, probably the right way is to fix the reporting code and adopt it for test usage, what do you think ?

zstan (commented):
intermediate review, i will proceed a bit minutes later

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -229,6 +257,44 @@ public void testNoIndexesNoPersistence() throws Exception {
...
+        inlineSize = 30;
+
+        IgniteEx ig0 = startGrid();
+        GridQueryProcessor qryProc = ig0.context().query();
+
+        Function<String, FieldsQueryCursor<List<?>>> execSql =
+            sql0 -> qryProc.querySqlFields(new SqlFieldsQuery(sql0), true);
+
+        execSql.apply("CREATE TABLE PUBLIC.TEST_TABLE (f1 int, f2 int, f3 int, f4 int, PRIMARY KEY (f1, f2, f3))");
+        execSql.apply("CREATE INDEX PK_IDX ON PUBLIC.TEST_TABLE (f1, f2, f3)");
```
zstan, line 274: why do we need such index ? test pass without it too.

zstan: Can you expand tests for case like :
`        execSql.apply("CREATE TABLE PUBLIC.TEST_TABLE (f1 int, f2 int, f3 int, f4 int, PRIMARY KEY (f1))");
        execSql.apply("CREATE INDEX PK_IDX ON PUBLIC.TEST_TABLE (f1, f2, f3)");`
        
`        execAssert.accept("SELECT * FROM PUBLIC.TEST_TABLE USE INDEX(PK_IDX) WHERE f1 = 8 and f2 in (1, 5, 3, 7) and f3 = 5");
        execAssert.accept("SELECT * FROM PUBLIC.TEST_TABLE USE INDEX(PK_IDX) WHERE f1 in (1, 5, 3, 7) and f2 = 8 and f3 = 5");
        execAssert.accept("SELECT * FROM PUBLIC.TEST_TABLE USE INDEX(PK_IDX) WHERE f1 = 8 and f2 = 5 and f3 in (1, 5, 3, 7)");`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/tree/BPlusTree.java
xtern: I updated the "reporting" code and redesigned the test to use a log listener.
Thanks.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
xtern: Thanks, I updated the test, but since the PK must be unique, I added a separate column for it.
Please take a look.

zstan (approved):
LGTM, plz proceed.

PR #2594, opened 2022-10-06
GG-35765 SQL: Fix deadlock when saving statistics data to the metastorage.

AMashenkov:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/stat/IgniteStatisticsConfigurationManager.java
```diff
@@ -421,42 +441,80 @@ public void updateStatistics(StatisticsObjectConfiguration... targets) {
...
+            });
+        }
 
-                    if (validate)
-                        validateDropRefresh(target, oldCfg);
+        chainFuture.listen(f -> {
+            if (f.error() != null)
+                resultFuture.onDone(f.error());
+            else
+                resultFuture.onDone(f.result() == null || f.result().booleanValue());
```
zstan, line 488: can result be a **null** here ?

zstan (commented):
from issue description i see that in time intervals t1 < t2 < t3 we obtain:
db-ckeckpoint-thread:
...
t2 -> lock.writeLock()
....

and
exchange-worker-thread:
...
t1 -> lock.readLock()
...
t3 -> lock.readLock()
It is correct, i can\`t see dead lock in description, can u fix description or explain me here ? 
`because the "exchange-worker" thread already has got the "checkpoint.readLock". At the same time, the "exchange-worker" thread is trying to get the "readLock" for the DistributedMetastorage but it can't because the "disco-notifier-worker" is trying to get the "writeLock"`

Explain plz - what section of async code help to avoid dead lock ?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/stat/IgniteStatisticsConfigurationManager.java
```diff
@@ -340,10 +351,19 @@ public void start() {
...
             distrMetaStorage.iterate(STAT_OBJ_PREFIX, (k, v) -> {
                 StatisticsObjectConfiguration cfg = (StatisticsObjectConfiguration)v;
 
-                updateLocalStatistics(cfg);
+                compoundFuture.add(updateLocalStatisticsAsync(cfg));
+            });
+
+            compoundFuture.markInitialized();
+
+            compoundFuture.listen(future -> {
```
zstan, line 364: previous approach was blocking and here we not waiting execution is it ok here ?

AMashenkov: Blocking operation under metastorage read-lock was the reason of hanging,
because disco-notify thread requires metastorage write-lock to finish CAS operation future.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/stat/IgniteStatisticsConfigurationManager.java
AMashenkov: Yes, here we convert 'null'->true

AMashenkov:
@zstan , I've added a test, TC looks ok.
Test hangs stably on main branch .

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/stat/StatisticsConfigurationTest.java
```diff
@@ -395,6 +395,50 @@ public void dropSingleColumnStatisticWhileNodeDown() throws Exception {
...
+     * - Ensure statistics for the table exists.
+     * - Disable StatisticsManagerConfiguration to prevent configuration changes in metastorage.
+     * - Drop table.
+     * - Re-activate the grid.
+     * - Ensures statistics for the table was dropped as well.
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testOrphanDataCleanup() throws Exception {
```
zstan, line 410: Not consider me as boring guy - but ticked about dead lock and we discuss test with dead lock and here i see orphan clearing and can`t link it with initial formulation, need some explanations here ..

zstan (approved):
LGTM

PR #2597, opened 2022-10-10
GG-35777 [IGNITE-17847] Dump threads if afterTest() times out

rpuch:
https://ggsystems.atlassian.net/browse/GG-35777


SammyVimes (approved):
LGTM

PR #2610, opened 2022-10-20
GG-35788 Fix serializable lambda serialization

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-35788

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
@@ -12854,4 +12854,20 @@ public abstract class IgniteUtils {
             }
         }
     }
+
+    public static boolean isLambda(Class<?> objectClass) {
+        return !objectClass.isPrimitive() && !objectClass.isArray()
+                && !objectClass.isAnonymousClass() && !objectClass.isLocalClass()
+                && objectClass.isSynthetic()
+                && classCannotBeLoadedByName(objectClass);
+    }
+
+    public static boolean classCannotBeLoadedByName(Class<?> objectClass) {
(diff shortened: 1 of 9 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -12854,4 +12854,20 @@ public static Socket acceptServerSocket(ServerSocket srvrSock) throws IOExceptio
             }
         }
     }
+
+    public static boolean isLambda(Class<?> objectClass) {
```
rpuch, line 12858: Let's create some unit tests for this method (at least that it can distinguish lambda a lambda from a non-lambda class)

rpuch (commented):
Is there a new test that would make sure that serializable lambdas really get serialized? I can only see technical changes to existing tests

modules/core/src/test/java/org/apache/ignite/internal/util/IgniteUtilsSelfTest.java
```diff
@@ -1595,6 +1596,71 @@ public void testResolveLocalAddressesWithHostNameDefinedAndLocalHostNameNotIgnor
...
+        assertTrue(IgniteUtils.isLambda(capturingOuterClassLambda.getClass()));
+        assertTrue(IgniteUtils.isLambda(methodReference.getClass()));
+    }
+
+    // Test nested class.
+    private static class TestNestedClass {
+    }
+
+    // Test inner class.
+    private static class TestInnerClass {
```
rpuch, line 1628: Should it be static then?

rpuch (approved):

PR #2616, opened 2022-10-31
GG-35823 Add freeMemPercentage metric for sql queries

shnus:


modules/indexing/src/test/java/org/apache/ignite/internal/mxbean/SqlQueryMXBeanImplSqlFreeMemTest.java
```diff
@@ -95,4 +95,32 @@ private void validateMemoryUsageOn(int nodeIdx, MemValidator validator) throws E
 
         validator.validate(free, maxMem);
     }
+
```
shnus, line 98: In order to not overcomplicate logic by supporting this test(SqlQueryMXBean is deprecated), I decided to copy here some validations

ingvard (approved):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/QueryMemoryManager.java
```diff
@@ -333,6 +333,14 @@ public long freeMemory() {
         return memoryLimit() - reserved();
     }
 
+    /**
+     * @return Percent of currently available memory for SQL queries on the node.
```
ygerzhedovich, line 337: What reason has the metric?  If some value could be evaluated by other values it must be done outside of the product.

PR #2632, opened 2022-11-15
GG-35529 ODBC: Fix uninitialized fields in meta

isapego:


ptupitsyn (approved):

ptupitsyn (commented):
Should we add a test for this scenario?

isapego:
I actually can not ass a test for this case as it was uninitialized data access. I actually caught by tests in very rare cases, but I can not make sure the unitialized data is non-zero.

PR #2636, opened 2022-11-16
GG-35915 Clean connections only once

xtern:
The test IgniteCacheQueryNodeFailTest.testNodeFailedSimpleQuery fails because in a rare case we may catch an AssertionError when the component stops

```
catch (Throwable e) {
    errOnStop = true;

    U.error(log, "Failed to stop component (ignoring): " + comp, e);

    if (e instanceof Error)
        throw (Error)e;
}
```
Errors re-throw here.

The assertion is thrown in ConnectionManager#recycle (if ConnectionManager#closeConnections has been called before).
```
void recycle(H2Connection conn) {
    boolean rmv = usedConns.remove(conn);

    assert rmv : "Connection isn't tracked [conn=" + conn + ']';
```

Those, to fix the fall of the test, it is enough to change this assert to an exception.

This fail case cannot be reproduced in the Apache Ignite 2.x codebase because the ConnectionManager stop is correctly implemented there. It can't be used after connections are closed - it throws an учсузешщт that the node is in the process of shutting down, which seems correct behavior.

### DIFF
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/ConnectionManager.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/ConnectionManager.java
@@ -32,6 +32,7 @@ import org.apache.ignite.internal.processors.query.QueryUtils;
 import org.apache.ignite.internal.processors.query.h2.opt.GridH2DefaultTableEngine;
 import org.apache.ignite.internal.processors.query.h2.opt.H2PlainRowFactory;
 import org.apache.ignite.internal.processors.timeout.GridTimeoutProcessor;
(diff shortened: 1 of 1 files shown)


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/ConnectionManager.java
```diff
@@ -188,6 +188,8 @@ public void onCacheDestroyed() {
      * Close all connections.
      */
     private void closeConnections() {
+        busyLock.block();
```
ygerzhedovich, line 191: if we has issue with double invocation the method then I don't understand how it will fix the problem? Second invocation will held here forever

korlov42 (approved):

PR #2649, opened 2022-12-05
GG-35955 IgniteDbMultiNodeWithIndexingEntryCompressionPutGetTest.testRandomRemove is flaky

rpuch:
https://ggsystems.atlassian.net/browse/GG-35955

* Make the test use less iterations as with current 30k it takes too much time on TC causing failures due to timeouts
* Actually randomize data (randomization was broken)


modules/core/src/test/java/org/apache/ignite/internal/processors/database/IgniteDbPutGetAbstractTest.java
```diff
@@ -762,14 +762,14 @@ public void testPutGetRandomUniqueMultipleObjects() throws Exception {
 
     /** */
     private static int[] generateUniqueRandomKeys(int cnt, Random rnd) {
-        int[] keys = new int[cnt];
+        Integer[] keys = new Integer[cnt];
```
sashapolo, line 765: Can we replace this code with `rnd.ints(cnt, 0, cnt).toArray()`?

sashapolo (approved):

PR #2659, opened 2022-12-09
GG-35841 Split brain if the coordinator node is in JVM pause state

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-35841

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/failure/AbstractFailureHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/failure/AbstractFailureHandler.java
@@ -61,9 +61,8 @@ public abstract class AbstractFailureHandler implements FailureHandler {
     }
 
     /**
-     * {@link ShutdownPolicy} to be used when handling critical situation by FailureProcessor mechanism.
-     *
-     * @return IMMEDIATE shutdown policy should be used as we treat any situation reached Failure Handler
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/failure/AbstractFailureHandler.java
```diff
@@ -59,6 +60,16 @@ public Set<FailureType> getIgnoredFailureTypes() {
         return !ignoredFailureTypes.contains(failureCtx.type()) && handle(ignite, failureCtx);
     }
 
+    /**
+     * {@link ShutdownPolicy} to be used when handling critical situation by FailureProcessor mechanism.
```
tkalkirill, line 64:
```suggestion
     * Returns {@link ShutdownPolicy} to be used when handling critical situation by FailureProcessor mechanism.
```

modules/core/src/main/java/org/apache/ignite/failure/AbstractFailureHandler.java
```diff
@@ -59,6 +60,16 @@ public Set<FailureType> getIgnoredFailureTypes() {
...
+     *
+     * @return IMMEDIATE shutdown policy should be used as we treat any situation reached Failure Handler
```
tkalkirill, line 66: Can describe in the body of the method javadoc and not in return value section?

PR #2673, opened 2022-12-21
GG-35786 Add default value for read and write buffers for thin client connections on server node

makedonskaya:


ptupitsyn (approved):

SammyVimes (approved):
LGTM

PR #2675, opened 2022-12-22
GG-36023 Ignore attempts to track free memory after the tracker is closed.

xtern:
This issue occurs because we are trying to call "release" on a closed memory tracker.
When we close, a partially read resultset from the thin client, the following happens:

- client connector (connector-thread) receives JdbcQueryCloseRequest.
- connector-thread closes cursors (see JdbcRequestHandler#closeQuery).
   When the cursor is about to close, it closes the associated memory tracker (see  RunningQueryManager#unregister)
   and sends GridQueryCancelRequest via communication (see GridReduceQueryExecutor#releaseRemoteResources).
- query-thread receives the GridQueryCancelRequest and closes the MapQueryResult, which in turn closes the H2ManagedLocalResult which in turn tries to call "release" on the associated (child) memory tracker, but this tracker already closed.

To fix the problem, I haven't yet come up with anything better than just allowing "release" to be called on a closed tracker.

The test changes look a little complicated, but they're actually simple.

1. We swap the inheritance hierarchy of the JdbcQueryMemoryTrackerSelfTest and JdbcThinQueryMemoryTrackerSelfTest classes (previously `JdbcQueryMemoryTrackerSelfTest extends JdbcThinQueryMemoryTrackerSelfTest`, and now `JdbcThinQueryMemoryTrackerSelfTest extends JdbcQueryMemoryTrackerSelfTest`).
2. We add our implementation to TestH2LocalResultFactory (to track errors in H2ManagedLocalResult).


zstan:
approach looks good to me but i have (as i think) more clear approach (send t Pavel via email) changes are a bit huge to comment them here. @xtern plz take a look on my mail and if it`s ok with you - lets fix your PR.

xtern:
> approach looks good to me but i have (as i think) more clear approach (send t Pavel via email) changes are a bit huge to comment them here. @xtern plz take a look on my mail and if it`s ok with you - lets fix your PR.

@zstan I have applied the suggested changes, thanks.

zstan:
@AMashenkov can u take a look on it plz ?

AMashenkov (approved):

PR #2682, opened 2022-12-29
GG-35987 Logs cache groups for which the index.bin is deleted

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-35987


SammyVimes (approved):
LGTM

PR #2683, opened 2022-12-29
GG-35960 PDS5 test suite fails with OutOfMemory error

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-35960

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
@@ -1763,7 +1763,8 @@ public class BPlusTreeSelfTest extends GridCommonAbstractTest {
         MAX_PER_PAGE = 5;
         CNT = 800;
 
-        // Sliding window size should be grater than cores to avoid races between puts and removes values in the tree.
+        // Sliding window size should be greater than the amount of CPU cores to avoid races between puts and removes
+        // in the tree.
         final int SLIDING_WINDOW_SIZE = CPUS * 2;
         final boolean DEBUG_PRINT = false;
 
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
@@ -214,8 +214,8 @@ public class BPlusTreeSelfTest extends GridCommonAbstractTest {
                     // Ignore.
                 }
 
-                // We should try to wait for all the futures to complete, since GridCompoundFuture#get(long) only waits
-                // for one of the futures to complete so that we don't want crash jvm due to PageMemory deallocation.
+                // We should try to wait for all the futures to complete, since GridCompoundFuture#get(long) fails if
+                // one of the futures fails so that we don't want crash jvm due to PageMemory deallocation.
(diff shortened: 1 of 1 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
```diff
@@ -1744,16 +1756,15 @@ public void doTestSizeForRandomPutRmvMultithreadedAsync(final int rmvPutSlidingW
...
-     *
      * @throws Exception if test failed
      */
     @Test
     public void testPutSizeLivelock() throws Exception {
         MAX_PER_PAGE = 5;
         CNT = 800;
 
-        final int SLIDING_WINDOW_SIZE = 16;
+        // Sliding window size should be grater than cores to avoid races between puts and removes values in the tree.
```
sashapolo, line 1766:
```suggestion
        // Sliding window size should be greater than the amount of CPU cores to avoid races between puts and removes in the tree.
```

modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
```diff
@@ -208,10 +208,22 @@ protected ReuseList createReuseList(int cacheId, PageMemory pageMem, long rootId
 
                 try {
                     asyncRunFut.cancel();
-                    asyncRunFut.get(60000);
+                    asyncRunFut.get(TimeUnit.MINUTES.toMillis(1));
```
ibessonov, line 211: Why do you need this get if you added new loop?

modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
```diff
@@ -208,10 +208,22 @@ protected ReuseList createReuseList(int cacheId, PageMemory pageMem, long rootId
...
                 }
                 catch (Throwable ex) {
-                    //Ignore
+                    // Ignore.
+                }
+
+                // We should try to wait for all the futures to complete, since GridCompoundFuture#get(long) fails if
```
sashapolo, line 217: I don't understand this comment: how does waiting for all futures prevent JVM crash?

sashapolo: oh, looks like I get it: the above compound future can fail, while other futures are still running

modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
tkalkirill: Fix it

modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
tkalkirill: Yep

modules/core/src/test/java/org/apache/ignite/internal/processors/database/BPlusTreeSelfTest.java
tkalkirill: Loop as an additional protection, in general, I think waiting on the future should be enough.

sashapolo (approved):

PR #2697, opened 2023-01-16
GG-35895 Fix communication check dangling connections

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-35895


modules/core/src/test/java/org/apache/ignite/internal/util/nio/GridNioSelfTest.java
```diff
@@ -755,6 +761,99 @@ public void testMultiThreadedSendReceive() throws Exception {
...
+                client.sendMessage(data, data.length);
+            }
+            catch (Exception e) {
+                error("Failed to send message.", e);
+
+                assert false : "Message sending failed: " + e;
+            }
+
+            // Wait until message is received by nio server as this means that connection is established.
+            latch.await(3, SECONDS);
```
sashapolo, line 797: you should use `assertTrue` here

sashapolo (approved):

PR #2699, opened 2023-01-19
GG-36198 Create tests for how deployed compute tasks are executed

rpuch:
https://ggsystems.atlassian.net/browse/GG-36198

### DIFF
--- a/modules/extdata/uri/META-INF/ignite.xml
+++ b/modules/extdata/uri/META-INF/ignite.xml
@@ -33,5 +33,6 @@
     -->
     <util:list id="tasks">
         <value>org.apache.ignite.spi.deployment.uri.tasks.GarHelloWorldTask</value>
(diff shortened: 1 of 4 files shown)


modules/extdata/uri/META-INF/ignite.xml
```diff
@@ -33,5 +33,6 @@
     -->
     <util:list id="tasks">
         <value>org.apache.ignite.spi.deployment.uri.tasks.GarHelloWorldTask</value>
+        <value>org.apache.ignite.spi.deployment.uri.tasks.SimpleNamedTask</value>
     </util:list>
 </beans>
```
SammyVimes, line 38: Should there be a newline?

SammyVimes (approved):
LGTM otherwise

modules/extdata/uri/META-INF/ignite.xml
rpuch: Maybe, but it wasn't before this change, so it's probably not a big deal

PR #2702, opened 2023-01-19
GG-35842 added info message for when sql offloading starts

kingnoskcaj:


solveme:
LGTM

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/disk/PlainExternalResult.java
```diff
@@ -112,6 +112,8 @@ private void spillRows() {
         if (F.isEmpty(rowBuff))
             return;
 
+        log.info("SQL offloading started");
```
sk0x50, line 115: IMHO, you need to check that the "info" level is enabled.
```
    if (log.isInfoEnabled())
        log.info("SQL offloading started);
```

kingnoskcaj:
This ticket is unnecessary as this was already implemented in https://ggsystems.atlassian.net/browse/GG-27744

PR #2714, opened 2023-02-02
GG-35941 Fixed incorrect handling of lost partitions by partition reconciliation tool

sk0x50:
https://ggsystems.atlassian.net/browse/GG-35941

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
@@ -140,22 +140,30 @@ public class ReconciliationAffectedEntries extends IgniteDataTransferObject {
             printer.accept("\nSKIPPED CACHES: " + skippedCachesCount() + "\n\n");
 
             for (PartitionReconciliationSkippedEntityHolder<String> skippedCache : skippedCaches) {
-                printer.accept("Following cache was skipped during partition reconciliation check cache=["
-                    + skippedCache.skippedEntity() + "], reason=[" + skippedCache.skippingReason() + "]\n");
+                String entity = skippedCache.skippedEntity();
+                String reason = skippedCache.skippingReason() != null ? skippedCache.skippingReason().reason() : null;
+
+                printer.accept(
+                    "Following cache was skipped during partition reconciliation check cache=[" + entity + "]" +
+                        (reason != null ? ", reason=[" + reason + ']' : "") + '\n');
             }
         }
 
         if (skippedEntries != null && !skippedEntries.isEmpty()) {
             printer.accept("\nSKIPPED ENTRIES: " + skippedEntriesCount() + "\n\n");
 
-            printer.accept(skippedEntries(skippedEntries));
+            printer.accept(skippedEntries(skippedEntries, includeSensitive));
         }
     }
 
     /**
      * @param skippedEntries Skipped entries.
+     * @param includeSensitive Indicates that sensitive information should be stored.
      */
-    private String skippedEntries(Map<String, Map<Integer, Set<PartitionReconciliationSkippedEntityHolder<PartitionReconciliationKeyMeta>>>> skippedEntries) {
+    private String skippedEntries(
+        Map<String, Map<Integer, Set<PartitionReconciliationSkippedEntityHolder<PartitionReconciliationKeyMeta>>>> skippedEntries,
+        boolean includeSensitive
+    ) {
         StringBuilder res = new StringBuilder();
 
         for (Map.Entry<String, Map<Integer, Set<PartitionReconciliationSkippedEntityHolder<PartitionReconciliationKeyMeta>>>>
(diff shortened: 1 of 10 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/PartitionReconciliationProcessor.java
```diff
@@ -537,6 +543,19 @@ private void detachWorkload(PipelineWorkload workload) {
             }
         }
 
+        /**
+         * Redirects processiong of skeppeed workload to collector.
```
sk0x50, line 547:
```suggestion
         * Redirects processing of skшppeed workload to the collector.
```

sk0x50, line 547:
```suggestion
         * Redirects processing of skipped workload to the collector.
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
```diff
@@ -352,6 +380,6 @@ public int skippedCachesCount() {
      * @return Skipped entries count.
      */
     public int skippedEntriesCount() {
-        return skippedEntries.size();
+        return skippedEntries.values().stream().flatMap(m -> m.values().stream()).mapToInt(Set::size).sum();
```
sanpwc, line 383: Seems that we miss tests for this.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
```diff
@@ -140,22 +140,30 @@ public void print(Consumer<String> printer, boolean includeSensitive) {
             printer.accept("\nSKIPPED CACHES: " + skippedCachesCount() + "\n\n");
 
             for (PartitionReconciliationSkippedEntityHolder<String> skippedCache : skippedCaches) {
-                printer.accept("Following cache was skipped during partition reconciliation check cache=["
-                    + skippedCache.skippedEntity() + "], reason=[" + skippedCache.skippingReason() + "]\n");
+                String entity = skippedCache.skippedEntity();
+                String reason = skippedCache.skippingReason() != null ? skippedCache.skippingReason().reason() : null;
```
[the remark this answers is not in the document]
sk0x50: `ReconciliationAffectedEntries` can be transferred via the network. In the case of the Rolling-Upgrade scenario, it is possible that the new value `SkippingReason#LOST_PARTITION` is not available on a particular node.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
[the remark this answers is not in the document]
sk0x50: This is already addressed here: https://github.com/gridgain/gridgain/pull/2714/files/#diff-80f98377c17e1d886378596f8b3646454a40d6c4cbbeaa98c0ec61c2a4654460R455

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/ReconciliationResultCollector.java
```diff
@@ -131,6 +141,9 @@ void appendRepairedEntries(
      * Represents a collector of inconsistent and repaired entries.
      */
     public static class Simple implements ReconciliationResultCollector {
+        /** */
+        private static final byte[] EMPTY_BYTES = new byte[0];
```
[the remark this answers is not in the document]
sk0x50: not a big deal. however, it would be nice to use an already defined constant.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
sk0x50: I extended my test to check this (compact and regular modes). Looks like there is only one case when `skipped` entries may appear - lost partitions.

sanpwc: Seems that ttl per entry is also the reason, is that correct?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/checker/processor/PartitionReconciliationLostPartitionsTest.java
```diff
@@ -0,0 +1,140 @@
...
+            grid(0),
+            new VisorPartitionReconciliationTaskArg.Builder()
+                .locOutput(false)
+                .repair(false)
+                .caches(Collections.singleton(DEFAULT_CACHE_NAME)));
+
+        assertEquals(
+            "Number of lost partitions does not match to the number of skipped entities.",
+            res.partitionReconciliationResult().skippedEntriesCount(),
+            lostParts.size());
```
sanpwc, line 115: Well, I didn't check startGridsMultiThreaded logic. The amount of an expected lost partitions differs from the amount of the caches that have such lost partitions, is that correct?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/objects/ReconciliationAffectedEntries.java
sk0x50: Yes, it can be a reason for skipping an entry, but it is not implemented https://ggsystems.atlassian.net/browse/GG-26267
I only found this simple check: https://github.com/gridgain/gridgain/blob/master/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/PartitionReconciliationProcessor.java#L186

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/checker/processor/PartitionReconciliationLostPartitionsTest.java
sk0x50: In general, `res.partitionReconciliationResult().skippedEntriesCount()` should return a total number of skipped entries that is the total number of lost partitions for all caches included in the check.

PR #2716, opened 2023-02-09
GG-36269 Fix duplicated rows in result in query with IN predicate.

xtern:
This is a fix for changes made in [GG-35682](https://github.com/gridgain/gridgain/commit/4e072cdde57ea732fbb4ba4de4a4f5f0833112ce)

To simplify the review the full version (including changes from previous patch) can be found here https://github.com/gridgain/gridgain/pull/2723

[GG-35682]: https://ggsystems.atlassian.net/browse/GG-35682?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
+++ b/modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
@@ -348,6 +348,7 @@ public class TableFilter implements ColumnResolver {
 
         // Remove conditions that cannot be used with the current index due to gaps.
         if (!hashIndex) {
+            boolean[] rmvConditionIndexes = null;
             boolean gapFound = false;
 
             for (int i = 0; i <= maxColumnIdx; i++) {
@@ -356,8 +357,19 @@ public class TableFilter implements ColumnResolver {
                     gapFound = true;
                 }
                 else if (gapFound) {
+                    if (rmvConditionIndexes == null) {
+                        rmvConditionIndexes = new boolean[indexConditions.size()];
+                    }
+
                     // Any condition associated with a column in the index after a gap cannot be used.
-                    indexConditions.remove(indexColumnConditions[i] - 1);
+                    rmvConditionIndexes[indexColumnConditions[i] - 1] = true;
+                }
+            }
+
+            if (rmvConditionIndexes != null) {
+                for (int i = rmvConditionIndexes.length - 1; i >= 0; i--) {
+                    if (rmvConditionIndexes[i])
+                        indexConditions.remove(i);
                 }
             }
         }
--- a/modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
+++ b/modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
@@ -313,7 +313,8 @@ public class TableFilter implements ColumnResolver {
             ((HashJoinIndex)index).prepare(session, indexConditions);
 
         boolean hashIndex = index.getIndexType().isHash();
-        ArrayList<Integer>[] indexColumnConditions = null;
+        boolean[] columnPresenceMask = null;
+        int[] columnIndexes = null;
         int maxColumnIdx = -1;
 
         // forget all unused index conditions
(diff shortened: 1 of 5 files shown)


modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
...
+
+                        continue;
+                    }
+
+                    if (hashIndex)
+                        continue;
+
+                    if (columnPresenceMask == null) {
+                        columnPresenceMask = new boolean[index.getColumns().length];
+                        columnIndexes = new int[indexConditions.size()];
```
AMashenkov, line 341: Why don't `new int[index.getColumns().length]`?

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
         if (index.getClass() == HashJoinIndex.class)
             ((HashJoinIndex)index).prepare(session, indexConditions);
 
+        boolean hashIndex = index.getIndexType().isHash();
+        boolean[] columnPresenceMask = null;
```
AMashenkov, line 316: java.util.BitSet can be used instead.

AMashenkov: Seems, you need only BitSet here to mark columns, which a condition was found for.
Then you can use `index.getColumns()` to find the first column, which condition is missed.
Then index conditions for the found columns and all the next ones can be removed.

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
...
+        // For example index has 3 columns (a,b,c),
+        // And we have the following conditions: c = 1 and a > 2.
+        // In this case we cannot use condition on 'c' and it can be removed.
+        for (int columnIdx = 0; columnIdx <= maxColumnIdx; columnIdx++) {
+            if (columnPresenceMask[columnIdx])
+                continue;
+
+            // We can remove any condition that uses an index column greater than a gap.
+            for (int i = columnIndexes.length - 1; i >= 0; i--) {
+                if (columnIndexes[i] > columnIdx)
```
AMashenkov, line 362: AFAIK index can be scanned in forward way only.
Should we check if range condition can be applied?

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
...
         // the indexConditions list may be modified here
         for (int i = 0; i < indexConditions.size(); i++) {
             IndexCondition condition = indexConditions.get(i);
             if (!condition.isAlwaysFalse()) {
                 Column col = condition.getColumn();
                 if (col.getColumnId() >= 0) {
-                    if (index.getColumnIndex(col) < 0) {
+                    int columnIdx = index.getColumnIndex(condition.getColumn());
+
+                    if (columnIdx < 0) {
```
AMashenkov, line 329: What is the case when we can get negative value here? for non-indexed column?
If so, then we should test that the gap in `columnIndexes` is correctly handled in the loop below, where unused index conditions are removed.

AMashenkov, line 329: When can this condition be true?

If the condition is true, `indexCondition.size()` will be changed.
Thus, if you are already created `columnIndexes[]` array, then the array will be larger than `indexConditions`.
Therefore, the array gap may lead to out-of-bound exception below,
when you will remove conditions with unused index columns.

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
...
+        // Removing all index conditions that cannot be applied because some index column is missing.
+        // For example index has 3 columns (a,b,c),
+        // And we have the following conditions: c = 1 and a > 2.
+        // In this case we cannot use condition on 'c' and it can be removed.
+        for (int columnIdx = 0; columnIdx <= maxColumnIdx; columnIdx++) {
+            if (columnPresenceMask[columnIdx])
+                continue;
+
+            // We can remove any condition that uses an index column greater than a gap.
+            for (int i = columnIndexes.length - 1; i >= 0; i--) {
```
AMashenkov, line 361:
```suggestion
            for (int i = indexConditions.size() - 1; i >= 0; i--) {
```
It is possible `columnIndex[]` is larger, but the array tail is empty.
This can happens when some conditions were removed after the `columnIndex[]` had beed created.

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
...
+                    if (hashIndex)
+                        continue;
+
+                    if (columnPresenceMask == null) {
+                        columnPresenceMask = new boolean[index.getColumns().length];
+                        columnIndexes = new int[indexConditions.size()];
                     }
+
+                    // Storing conditions using the order of their columns in the index to find possible gap.
+                    columnPresenceMask[columnIdx] = true;
```
AMashenkov, line 345:
```suggestion
                    columnUsedInConditions[columnIdx] = true;
```

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
```diff
@@ -312,21 +312,60 @@ public void prepare() {
...
+                    columnIndexes[i] = columnIdx;
```
AMashenkov, line 346:
```suggestion
                    conditionColumnIndices[i] = columnIdx;
```

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
xtern: AFAIK BitSet is good for large sets. In this case, we won't save much memory, and as far as I know, performance is worse with bitset than with boolean array.
But if you insist I can change the type to BitSet.

modules/h2/src/main/java/org/gridgain/internal/h2/table/TableFilter.java
xtern: Done, thanks!

AMashenkov (approved):

AMashenkov (changes requested):
Actually, it is better to revert previous broken fix or rework, because it is not clear if the state after this fix will be better than before [GG-35682]

Regardless the tests looks fine, we can't be sure index is used where it was used before [GG-35682]. It is possible we start fall into full-table scan in some cases.

PR #2719, opened 2023-02-13
GG-36211 Fixed SystemViewSelfTest.testTransactions

sk0x50:
https://ggsystems.atlassian.net/browse/GG-36211


alievmirza (approved):
In general, I'm ok with this patch, there is a minor comment, that is not necessary to fix: we could not wait for all threads to complete their puts in case of testing the puts to one cache.

sk0x50:
> In general, I'm ok with this patch, there is a minor comment, that is not necessary to fix: we could not wait for all threads to complete their puts in case of testing the puts to one cache.

Well, I would stick with the current fix. First of all, I don't think that it will reduce the test execution time. And I don't want to start a new round of getting the green visa from the MTCGA bot. :)

PR #2720, opened 2023-02-13
GG-36055 Utility methods to work with cluster's topology

sergey-chugunov-1985:


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/util/TopologyUtils.java
```diff
@@ -0,0 +1,65 @@
...
+import java.util.UUID;
+
+/**
+ * Class provides utility methods to work with nodes in topology:
+ * finding coordinator, filtering client nodes and so on.
+ */
+public final class TopologyUtils {
+    /**
+     * Obtains ID of coordinator node from the topology
+     * visible through {@link GridClientCompute client compute} interface at the moment of method call.
```
rpuch, line 34: Should we mention the rule to select the coordinator ('a server node with the least order property')?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/util/TopologyUtils.java
```diff
@@ -0,0 +1,65 @@
...
+            .nodes(node -> !node.isClient())
+            .stream()
+            .min(Comparator.comparingLong(GridClientNode::order))
+            .map(GridClientNode::nodeId)
+            .orElse(null);
+    }
+
+    /**
+     * Obtains any server node from the topology
+     * visible through {@link GridClientCompute client compute} interface at the moment of method call.
```
rpuch, line 52: Should we mention 'connectable' in the javadoc?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/DefragmentationCommand.java
```diff
@@ -53,7 +54,7 @@ public class DefragmentationCommand extends AbstractCommand<DefragmentationArgum
     /** {@inheritDoc} */
     @Override public Object execute(GridClientConfiguration clientCfg, Logger log) throws Exception {
         try (GridClient client = Command.startClient(clientCfg)) {
-            Optional<GridClientNode> firstNodeOpt = client.compute().nodes().stream().filter(GridClientNode::connectable).findFirst();
+            Optional<GridClientNode> firstNodeOpt = anyConnectableServerNode(client.compute());
```
rpuch, line 57: The change is not equivalent: code 'before' was taking the first node, while 'after' it takes any node. Is this because `compute().nodes()` ordering does not have any specific sense and might be considered as arbitrary?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/PersistenceCommand.java
```diff
@@ -58,7 +59,7 @@ public class PersistenceCommand extends AbstractCommand<PersistenceArguments> {
     /** {@inheritDoc} */
     @Override public Object execute(GridClientConfiguration clientCfg, Logger logger) throws Exception {
         try (GridClient client = Command.startClient(clientCfg)) {
-            Optional<GridClientNode> firstNodeOpt = client.compute().nodes().stream().findFirst();
+            Optional<GridClientNode> firstNodeOpt = anyConnectableServerNode(client.compute());
```
rpuch, line 62: 'before' a non-connectable node might be chosen, so the change is not equivalent. Is this a fix?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrRebuildPartitionLogCommand.java
```diff
@@ -55,7 +56,7 @@ public class DrRebuildPartitionLogCommand extends AbstractCommand<DrRebuildParti
     /** {@inheritDoc} */
     @Override public Object execute(GridClientConfiguration clientCfg, Logger logger) throws Exception {
         try (GridClient client = Command.startClient(clientCfg)) {
-            Optional<GridClientNode> firstNodeOpt = client.compute().nodes().stream().findFirst();
+            Optional<GridClientNode> firstNodeOpt = TopologyUtils.anyConnectableServerNode(client.compute());
```
rpuch, line 59: 'before' a non-connectable node might be chosen, so the change is not equivalent. Is this a fix?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/util/TopologyUtils.java
sergey-chugunov-1985: I addressed this comment, could you please take another look?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/util/TopologyUtils.java
sergey-chugunov-1985: I added this information to the javadoc, I hope it is clear enough.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/PersistenceCommand.java
sergey-chugunov-1985: In the case of persistence command the whole cluster contains only one node as persistence command works only with nodes in Maintenance Mode for persistence repair and clean-up.

So here it doesn't matter.

rpuch (approved):

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/DefragmentationCommand.java
sergey-chugunov-1985: From what I see in GridClientTopology implementation, current state of topology on a client side is managed by a HashMap<UUID, GridClientNodeImpl> collection that doesn't maintain any particular order of elements.

So I believe that in this case there is no semantic difference between "first node" and "any node" methods.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/dr/subcommands/DrRebuildPartitionLogCommand.java
sergey-chugunov-1985: I think this is not a problem as well and here is why: even if the code for finding the first node returns a non-empty Optional, value from this Optional is completely ignored just a few lines later when `null` nodeId is passed to `executeTaskByNameOnNode` method.

`null` there means that implementation will use a node from GridClient connection.

Moreover, even if we passed the node from Optional and that node turned out to be non-connectable, we would get an exception as `executeTaskByNameOnNode` method filters nodes by `connectable` flag internally.

PR #2731, opened 2023-03-09
GG-36261 Implemented handling for a situation of unavailable class in CacheConfiguration

sergey-chugunov-1985:


modules/core/src/test/java/org/apache/ignite/p2p/P2PCustomSqlFunctionsConfigurationTest.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /**
+     * Test verifies that dynamic cache request issued by a client node with a class unavailable on server side
+     * fails with a proper exception.
+     * However, no client nor server nodes fail as a result.
+     *
+     * @throws Exception If test fails to set up necessary environment.
+     */
+    @Test
+    public void testClientStartsDynamicCacheWithUnavailableClass() throws Exception {
```
sk0x50, line 150: Let's add the following test scenario:
 - start one server and one client node
 - create a new dynamic cache (let's assume that the required class is available on both nodes)
 - start a new server/client node that does not contain the required class in its classpath

sk0x50:
The proposed solution has a few limitations:
 - mutating discovery messages does not work with Zookeeper discovery SPI, in general
 - also this mutating leads to the fact that the processing time required to start a new cache will be increased (we cannot send this such a message to the next node in the ring until it is handled by the current node)

Let's take a look at `@SerializeSeparately` annotation. You can just add this annotation to the `sqlFuncCls` field and check the tests. Perhaps, this approach does not solve the case with a static configuration, but I hope it can be fixed as well.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/DynamicCacheChangeBatch.java
```diff
@@ -40,6 +47,14 @@ public class DynamicCacheChangeBatch implements DiscoveryCustomMessage {
     /** Discovery custom message ID. */
     private IgniteUuid id = IgniteUuid.randomUuid();
 
+    /** Request ID -> Initiator Node ID mapping extracted from {@link DynamicCacheChangeBatch#reqs} collection
```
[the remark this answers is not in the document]
sergey-chugunov-1985: It turned out that implementation based on SerializeSeparately annotation doesn't need these changes in Batch message, so I reverted them.

modules/core/src/test/java/org/apache/ignite/p2p/P2PCustomSqlFunctionsConfigurationTest.java
sergey-chugunov-1985: Done, new scenario added and works as expected (new node fails to join the cluster).

modules/core/src/test/java/org/apache/ignite/p2p/P2PCustomSqlFunctionsConfigurationTest.java
```diff
@@ -0,0 +1,351 @@
...
+
+    /** */
+    private static final String UNAVAILABLE_TO_SERVER_SQL_FUNCTIONS_CLASS_NAME =
+        "org.apache.ignite.tests.p2p.cache.UnavailableToServerCustomSqlFunctionsClass";
+
+    static {
+        try {
+            CONFIGURATION_CLASS_LOADER = new URLClassLoader(
+                new URL[]{new URL(GridTestProperties.getProperty("p2p.uri.cls.second"))},
+                GridP2PScanQueryWithTransformerTest.class.getClassLoader());
```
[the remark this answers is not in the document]
sergey-chugunov-1985: Done.

modules/core/src/test/java/org/apache/ignite/p2p/P2PCustomSqlFunctionsConfigurationTest.java
```diff
@@ -0,0 +1,351 @@
...
+    public static final String DYNAMIC_CACHE_TEMPLATE = "DYNAMIC_CACHE_TMPL_*";
+
+    /** */
+    private static final String UNAVAILABLE_TO_SERVER_SQL_FUNCTIONS_CLASS_NAME =
+        "org.apache.ignite.tests.p2p.cache.UnavailableToServerCustomSqlFunctionsClass";
+
+    static {
+        try {
+            CONFIGURATION_CLASS_LOADER = new URLClassLoader(
+                new URL[]{new URL(GridTestProperties.getProperty("p2p.uri.cls.second"))},
```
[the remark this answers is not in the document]
sergey-chugunov-1985: Fixed as well.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheConfigurationEnricher.java
```diff
@@ -141,7 +142,15 @@ public CacheGroupDescriptor enrich(CacheGroupDescriptor desc, boolean affinityNo
             }
         }
         catch (Exception e) {
-            throw new IgniteException("Failed to enrich cache configuration [cacheName=" + ccfg.getName() + "]", e);
+            StringBuilder msgBldr = new StringBuilder("Failed to enrich cache configuration [cacheName=")
```
sk0x50, line 145: I think that we need to update `CacheConfigurationSplitterImpl` as well. Let's assume the following RU scenario:
 - start one server node using gg version that does not support serializing `sqlFunCls`
 - start a new node with the fix
 - request a new dynamic cache from the new node
 
 In my understanding, it can be a problem for the first node to properly process the request because the real value of `sqlFuncCls` will be placed in the enrichment.

sergey-chugunov-1985: @sk0x50 I wrote exactly that compatibility test and it seems working well. The reason for that is in `CacheConfigurationEnricher` code: it doesn't care about which fields to process and which not and deserialize them all and set into cache configuration.

So no additional logic is needed here.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -3135,7 +3135,8 @@ private GridCacheSharedContext createSharedContext(
         if (validationRes != null)
             return new IgniteNodeValidationResult(node.id(), validationRes);
 
-        return ValidationOnNodeJoinUtils.validateNode(node, discoData, marsh, ctx, this::cacheDescriptor);
+        return ValidationOnNodeJoinUtils.validateNode(node, discoData, marsh, ctx, this::cacheDescriptor,
```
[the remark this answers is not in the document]
sergey-chugunov-1985: I wrote such test and it hangs indeed. Implementing a fix for it is tricky, so I created a separate ticket and muted the test at the moment: [GG-36586](https://ggsystems.atlassian.net/browse/GG-36586).

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ValidationOnNodeJoinUtils.java
```diff
@@ -171,12 +179,26 @@ public class ValidationOnNodeJoinUtils {
                     }
                 }
 
-                DynamicCacheDescriptor locDesc = cacheDescProvider.apply(cacheInfo.cacheData().config().getName());
+                //check is performed only on affinity nodes for the cache requested to start
+                if (CU.affinityNode(ctx.discovery().localNode(), joiningNodeCacheCfg.getNodeFilter())) {
```
vldpyatkov, line 183: We will skip the check if coordinator node is filtered by affinity.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
```diff
@@ -1088,11 +1088,25 @@ else if (op0 instanceof SchemaAddQueryEntityOperation) {
     public void initQueryStructuresForNotStartedCache(DynamicCacheDescriptor cacheDesc) throws IgniteCheckedException {
         QuerySchema schema = cacheDesc.schema() != null ? cacheDesc.schema() : new QuerySchema();
 
+        if (ctx.discovery().localNode().isClient())
+            enrichConfiguration(cacheDesc);
```
vldpyatkov, line 1092: This place looks very suspicious, because cache configuration class has many fields marked with SerializeSeparately annotation. Here we deserialize all of them, but required only sqlFuncCls.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ValidationOnNodeJoinUtils.java
sergey-chugunov-1985: Good catch, I removed this check and marked to ignore a jUnit test relying in this behavior.

modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridQueryProcessor.java
sergey-chugunov-1985: I fixed this. The fix isn't ideal but it enables to pass a filter function to configuration enricher that allows only a specific fields to be deserialized.
Now only sqlFuncCls field will be deserialized by this call.

PR #2733, opened 2023-03-13
GG-36405 Connection is closed immediately without writing back TLS error

SammyVimes:
https://ggsystems.atlassian.net/browse/GG-36405


modules/core/src/main/java/org/apache/ignite/spi/discovery/tcp/TcpDiscoverySpi.java
```diff
@@ -1686,7 +1687,7 @@ protected void writeToSocket(Socket sock, TcpDiscoveryAbstractMessage msg, byte[
             out.flush();
         }
         catch (IOException e) {
-            err = e;
+            err = maybeSslError(sock, e);
```
sergey-chugunov-1985, line 1690: I suggest to rename this method to `clearUpSslEssence` to better reflect its semantics: find out if exception's context indicates it is an SSL-related error and wrap it into SSLException if needed.

sergey-chugunov-1985 (changes requested):

sergey-chugunov-1985 (approved):

PR #2736, opened 2023-03-16
GG-36242 [IGNITE-18591] Java thin client: Endpoints discovery

ptupitsyn:


modules/core/src/main/java/org/apache/ignite/internal/client/thin/ProtocolBitmaskFeature.java
```diff
@@ -37,6 +37,9 @@ public enum ProtocolBitmaskFeature {
      */
     CLUSTER_API(2),
 
+    /** Client discovery. */
+    CLUSTER_GROUP_GET_NODES_ENDPOINTS(3),
```
isapego, line 41: See `ClientOperation.java`
```suggestion
    CLUSTER_GROUP_GET_NODE_ENDPOINTS(3),
```

isapego (approved):

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ProtocolBitmaskFeature.java
ptupitsyn: `NODES_ENDPOINTS` is the correct name, because we returns a list of nodes with multiple endpoints per node.

ptupitsyn: Let's keep it the same way as in Ignite to make merges easier.

PR #2739, opened 2023-03-17
GG-35372 Failed to activate cluster with control.sh

josephdanrev:
Don't get clusterName() when cluster is INACTIVE


josephdanrev (commented):
Updated checkstyle

PR #2753, opened 2023-03-27
GG-36410 Shade AWS dependencies in Control Center Agent

andreybell91:


andreybell91:
Related to CCA https://github.com/ggprivate/control-center-agent/pull/483

nva (commented):
May be we should move all shaded deps to different package (org.gridgain.control.dependencies, for example) and mute it all at once?

andreybell91:
> May be we should move all shaded deps to different package (org.gridgain.control.dependencies, for example) and mute it all at once?

It may make sense. I've created the task to analyze https://ggsystems.atlassian.net/browse/GG-36509

nva (approved):

PR #2763, opened 2023-04-04
GG-36321 SQL: Print warning if primary key is potentially misconfigured.

AMashenkov:
Also forbid to run COPY command from Java API because this was is broken. Seemd, we assumed COPY will run only from JDBC Thin Client.

### DIFF
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
@@ -1976,6 +1976,15 @@ public class IgniteH2Indexing implements GridQueryIndexing {
     @Override public boolean registerType(GridCacheContextInfo cacheInfo, GridQueryTypeDescriptor type, boolean isSql)
         throws IgniteCheckedException {
         validateTypeDescriptor(type);
+
+        // Print warning if _key is of custom user type and no key fields are configured, thus
+        // key object can't be created and used for COPY FROM operation.
+        // Note: QueryEntity can't be properly validated instantly due to compatibility reasons, and
+        // warning can't be printed in QE due to absence of logger.
+        if (type.keyClass() == Object.class && F.isEmpty(type.primaryKeyFields())) {
+            log.warning("Key of user type has no fields configured for table=" + type.tableName());
+        }
+
         schemaMgr.onCacheTypeCreated(cacheInfo, this, type, isSql);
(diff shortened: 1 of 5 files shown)


modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/QueryEntityValidationSelfTest.java
```diff
@@ -279,4 +290,38 @@ public void testNotUniqueNameInAnnotation() {
...
+    /**
+     * Validates a warning is logged when query entity key is potentially misconfigured.
+     */
+    @Test
+    public void testNonAnnotatedField() {
+        CacheConfiguration<TestKeyNoAnnotatedFields, TestValue> ccfg = new CacheConfiguration<>();
+        ccfg.setName(CACHE_NAME);
+        ccfg.setIndexedTypes(TestKeyNoAnnotatedFields.class, TestValue.class);
+
+        LogListener logListener = LogListener.matches("Key of user type has no fields configured for table=TESTVALUE").build();
```
zstan, line 313: I suggest to expand such a message, probably we need to highlight @QuerySqlField  annotation ?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1976,6 +1976,15 @@ public UpdateResult executeUpdateOnDataNode(
     @Override public boolean registerType(GridCacheContextInfo cacheInfo, GridQueryTypeDescriptor type, boolean isSql)
         throws IgniteCheckedException {
         validateTypeDescriptor(type);
+
+        // Print warning if _key is of custom user type and no key fields are configured, thus
+        // key object can't be created and used for COPY FROM operation.
+        // Note: QueryEntity can't be properly validated instantly due to compatibility reasons, and
+        // warning can't be printed in QE due to absence of logger.
```
zstan, line 1983: what does QE abbr mean?

zstan (approved):
fill minor comments, overall looks good

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/QueryEntityValidationSelfTest.java
AMashenkov: Annotation is just one of the ways to configure key fields.

PR #2765, opened 2023-04-05
GG-36507 .NET: Add IgniteClientConfiguration.EnableClusterDiscovery

ptupitsyn:
* Add `IgniteClientConfiguration.EnableClusterDiscovery`.
* Allow any combinations of `EnableClusterDiscovery` and `EnablePartitionAwareness`:
  * Only `EnablePartitionAwareness`: use configured endpoints to route requests.
  * Only `EnableClusterDiscovery`: use discovered endpoints for failover, but maintain only one connection.
  * Both enabled: discover all, connect to all.
  * Both disabled: maintain one connection, use only configured endpoints for failover.

### DIFF
--- a/modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs
+++ b/modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs
@@ -591,27 +591,34 @@ namespace Apache.Ignite.Core.Impl.Client
         {
             _affinityTopologyVersion = affinityTopologyVersion;
 
-            if (_discoveryTopologyVersion < affinityTopologyVersion.Version && _enableDiscovery)
+            if (!_config.EnablePartitionAwareness && !_enableDiscovery)
             {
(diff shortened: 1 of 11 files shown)


modules/platforms/dotnet/Apache.Ignite.Core/Impl/Client/ClientFailoverSocket.cs
```diff
@@ -508,7 +508,12 @@ private void OnNewDefaultConnection()
                 );
             }
 
-            if (!_socket.Features.HasFeature(ClientBitmaskFeature.ClusterGroupGetNodesEndpoints))
+            if (!_config.EnablePartitionAwareness || !_config.EnableClusterDiscovery)
```
isapego, line 511: What if user wants Cluster Discovery but without Partition Awareness? Just for failover?

ptupitsyn: Good point. I've discarded this use case as weird initially, but there in no reason not to support it.

Fixed.

isapego (approved):

PR #2767, opened 2023-04-05
GG-33516 Reset lost partitions tool to automatically discover affected caches

ashapkin:
https://ggsystems.atlassian.net/browse/GG-33516

### DIFF
--- a/modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
+++ b/modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
@@ -38,7 +38,6 @@ import org.apache.ignite.internal.commandline.cache.reset_lost_partitions.CacheR
 import org.apache.ignite.internal.commandline.cache.reset_lost_partitions.CacheResetLostPartitionsTaskResult;
 import org.apache.ignite.internal.processors.cache.distributed.dht.topology.GridDhtPartitionState;
 
-import static org.apache.ignite.internal.commandline.CommandLogger.optional;
 import static org.apache.ignite.internal.commandline.CommandLogger.or;
 import static org.apache.ignite.internal.commandline.TaskExecutor.BROADCAST_UUID;
 import static org.apache.ignite.internal.commandline.TaskExecutor.executeTaskByNameOnNode;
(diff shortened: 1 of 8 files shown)


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
```diff
@@ -16,35 +16,51 @@
...
  */
 public class ResetLostPartitions extends AbstractCommand<Set<String>> {
     /** {@inheritDoc} */
     @Override public void printUsage(Logger logger) {
         String CACHES = "cacheName1,...,cacheNameN";
-        String description = "Reset the state of lost partitions for the specified caches.";
+        String description = "Reset the state of lost partitions for the specified or all affected caches.";
 
-        usageCache(logger, RESET_LOST_PARTITIONS, description, null, CACHES);
+        usageCache(logger, RESET_LOST_PARTITIONS, description, null, CACHES,
```
sk0x50, line 58: In my understanding, it should be as follows: 
`usageCache(logger, RESET_LOST_PARTITIONS, description, null, or(CACHES, ALL_CACHES_MODE));`

If I am not mistaken, the current implementation only allows two options:
 - reset_lost_partitions cache_name
 - reset_lost_partitions --all

Should we allow using reset_lost_partitions without any arguments (assuming that this case equals to --all option)?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
```diff
@@ -67,9 +85,47 @@ public class ResetLostPartitions extends AbstractCommand<Set<String>> {
...
+        if (caches.size() > 0)
+            logger.info("The following caches have LOST partitions: " + CommandLogger.join(",", caches) + ".");
+        else
+            logger.info("No caches with LOST partition has been found found.");
+    }
+
     /** {@inheritDoc} */
     @Override public void parseArguments(CommandArgIterator argIter) {
-        caches = argIter.nextStringSet("Cache names");
+
```
sk0x50, line 117: Please remove the empty line here

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
```diff
@@ -67,9 +85,47 @@ public class ResetLostPartitions extends AbstractCommand<Set<String>> {
...
+        String nextArg = argIter.nextArg("");
```
sk0x50, line 118: It would be nice to provide a detailed message. Your implementation assumes that it should be a cache name, or `--all` option.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
```diff
@@ -67,9 +85,47 @@ public class ResetLostPartitions extends AbstractCommand<Set<String>> {
...
+                        if (partition.getState() == GridDhtPartitionState.LOST) {
+                            caches.add(group.getGroupName());
+                            break;
+                        }
+                    }
+                }
+            }
+        }
+
+        if (caches.size() > 0)
```
sk0x50, line 109: Please use `!isEmpty()` instead of `size()`

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
ashapkin: Nothing has changed if no option is specified. In that case, it's going to use the old behavior - return nothing and no additional hints or information is printed.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/cache/ResetLostPartitions.java
ashapkin: Fixed

PR #2768, opened 2023-04-05
GG-36202 Restrict possibility to create two caches with the same schemas and index names

zstan:


AMashenkov:
Let's add a test for dynamic caches and dynamic indexes if we don't have one.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/ErroneousQueryEntityConfigurationTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+        stopGrid("srv1");
+
+        if (persistent)
+            startGrid("srv1");
+    }
+
+    /** */
+    @Test
+    public void testErroneousCacheConfigFromClientNodeWithPersistence() throws Exception {
```
AMashenkov, line 134:
```suggestion
    public void testErroneousCacheConfigFromClientNode() throws Exception {
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/ErroneousQueryEntityConfigurationTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+            client.getOrCreateCaches(ccfgs);
+        }, CacheException.class, "Duplicate index name");
+
+        server.cluster().state(ClusterState.ACTIVE);
+    }
+
+    /** */
+    @Test
+    public void teststartErroneousCacheConfigThroughDynamicCachesWithPersistence() throws Exception {
```
AMashenkov, line 158:
```suggestion
    public void teststartErroneousCacheConfigThroughDynamicCaches() throws Exception {
```

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/ErroneousQueryEntityConfigurationTest.java
```diff
@@ -0,0 +1,294 @@
...
+        if (persistent) {
+            stopGrid("srv1");
+
+            startGrid("srv1");
+        }
+    }
+
+    /** */
+    @Test
+    public void testEqualIndexAreConfiguredOnServerAndJoinedNodeWithPersistence() throws Exception {
```
AMashenkov, line 211:
```suggestion
    public void testEqualIndexAreConfiguredOnServerAndJoinedNode() throws Exception {
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridLocalConfigManager.java
```diff
@@ -298,6 +315,62 @@ private void addCacheFromConfiguration(
...
 
+    /**
+     * Validates already processed cache configuration instead newly defined.
+     *
+     * @param cacheConfigs Already processed caches.
+     * @param cfg Currently processed cache config.
+     * @return Error message, if supplied configuration is incorrect.
+     * @throws IgniteException If misconfigured.
+     */
+    @Nullable public static String validateIncomingConfiguration(
```
xtern, line 326: Since we only checking the index names then maybe we should change the method name to something like "validateIncomingIndexConfiguration(".

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/ErroneousQueryEntityConfigurationTest.java
```diff
@@ -0,0 +1,294 @@
...
+
+            client.getOrCreateCaches(ccfgs);
+        }, CacheException.class, "Duplicate index name");
+
+        server.cluster().state(ClusterState.ACTIVE);
+    }
+
+    /** */
+    @Test
+    public void teststartErroneousCacheConfigThroughDynamicCaches() throws Exception {
```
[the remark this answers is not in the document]
zstan: it`s my fail, previously this tests runs without parameters, i`l fix it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridLocalConfigManager.java
zstan: ok, but i prefer to stay it as is.

zstan:
@xtern thanks, i fix all your comments, besides one.

xtern (approved):

PR #2779, opened 2023-04-11
GG-36506 Reencryption is sometimes not resumed after restart

rpuch:
https://ggsystems.atlassian.net/browse/GG-36506


modules/core/src/main/java/org/apache/ignite/internal/managers/encryption/GridEncryptionManager.java
```diff
@@ -598,6 +598,8 @@ public void onLocalJoin() {
 
         Map<Integer, Object> encKeysFromCluster = (Map<Integer, Object>)data.commonData();
 
+        log.info("encKeysFromCluster: " + encKeysFromCluster);
```
tkalkirill, line 601: Are you sure we need to log this?

rpuch: Good catch, removed this

tkalkirill (approved):

PR #2783, opened 2023-04-12
Avoid eternal hangs while dumping locks for diagnostics

rpuch:
https://ggsystems.atlassian.net/browse/GG-36548

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/diagnostic/pagelocktracker/SharedPageLockTracker.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/diagnostic/pagelocktracker/SharedPageLockTracker.java
@@ -190,35 +190,48 @@ public class SharedPageLockTracker {
         Collection<PageLockTracker<?>> trackers = threadStacks.values();
         List<ThreadPageLockState> threadPageLockStates = new ArrayList<>(threadStacks.size());
 
-        for (PageLockTracker<?> tracker : trackers) {
-            boolean acquired = tracker.acquireSafePoint();
+        Set<PageLockTracker<?>> acquiredTrackers = new HashSet<>();
 
-            //TODO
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/diagnostic/pagelocktracker/SharedPageLockTracker.java
```diff
@@ -190,35 +190,48 @@ public synchronized SharedPageLockTrackerDump dump() {
...
-            assert acquired;
-        }
+        try {
+            for (PageLockTracker<?> tracker : trackers) {
+                boolean acquired = tracker.acquireSafePoint();
+
+                if (acquired)
+                    acquiredTrackers.add(tracker);
+
+                //TODO
```
SammyVimes, line 202: There is no ticket, I think it can be removed. Also it looks a bit strange because we have 
`if (acquired)` in the previous line. This is for tests, right?

SammyVimes (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/diagnostic/pagelocktracker/SharedPageLockTracker.java
rpuch: As for the TODO, I left it in the state which it had before the fix. A TODO without a ticket reference is against the rules, but it still makes sense as a marker. I think we should keep it, maybe someone will do something about it in the future.

`acquiredTrackers` is used to keep track (no pun intended!) of the trackers on which we acquired locks to have ability to release not-yet-released locks reliably in the `finally` block, hence the `if`. This behavior is not used by any tests.

SammyVimes (approved):
LGTM

PR #2791, opened 2023-04-19
GG-36552 .NET: Fix deadlock on continuous query close

ptupitsyn:
Deadlock condition:
* Thread 1 processes continuous query events, takes `listenerLock`, then `PlatformContinuousQueryImpl.lock`
* Thread 2 processes continuous query close, takes `PlatformContinuousQueryImpl.lock`, then `listenerLock`

Fix: there is no way to ensure the same order of locks, so we resort to `tryLock` in `PlatformContinuousQueryImpl`.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
@@ -180,6 +180,7 @@ public class PlatformContinuousQueryImpl implements PlatformContinuousQuery {
 
     /** {@inheritDoc} */
     @Override public void onUpdated(Iterable evts) throws CacheEntryListenerException {
+        // onUpdated is not under listenerLock, so there is no deadlock possibility like in evaluate
         lock.readLock().lock();
 
         try {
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/cache/query/PlatformContinuousQueryImpl.java
```diff
@@ -169,12 +174,21 @@ private static CacheEntryEventFilter getJavaFilter(Object filter, GridKernalCont
...
-        lock.readLock().lock();
+        // First, use startLatch to ensure full initialization.
+        try {
+            startLatch.await();
+        } catch (InterruptedException e) {
+            throw new IgniteInterruptedException(e);
+        }
+
+        if (!lock.readLock().tryLock())
+            return; // Query is being closed (see tryLock reasoning in #evaluate).
```
sk0x50, line 191: It seems to me, we should throw a new exception here. 
`throw new CacheEntryListenerException("Failed to notify listener because it has been closed.");`

ptupitsyn: I don't think it makes sense to throw an exception from `onUpdate` when the query is closed. The user has stopped the continuous query, which means he is not interested in updates anymore. We can just ignore them.

I've replaced existing `throw` in `onUpdate` with `return` as well, to be consistent.

sk0x50 (approved):

isapego (approved):

PR #2797, opened 2023-04-23
GG-35372 Failed to activate cluster with control.sh

josephdanrev:
Handle scenario when changing states from INACTIVE causing null exceptions


sergey-chugunov-1985 (approved):
@josephdanrev I added two minor commits to shape the code better, otherwise patch looks good to me.

Please merge it to master branch with squash operation.

PR #2798, opened 2023-05-02
GG-36605 Add date to JavaLoggerFormatter message layout

solveme:


modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -557,6 +557,10 @@ public abstract class IgniteUtils {
     private static final DateTimeFormatter THREAD_DUMP_FMT =
         DateTimeFormatter.ofPattern("yyyy/MM/dd HH:mm:ss z").withZone(ZoneId.systemDefault());
 
+    /** ISO-8601 compliant date-time format. */
+    public static final DateTimeFormatter ISO_DATE_FMT =
+        DateTimeFormatter.ofPattern("yyyy-MM-dd'T'HH:mm:ss,SSS").withZone(ZoneId.systemDefault());
```
sashapolo, line 562: can we use `DateTimeFormatter.ISO_LOCAL_DATE_TIME` instead?

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+            Thread.currentThread().getName() + "][" +
+            category + "] " +
+            message;
+    }
+
+    static void assertSimilar(String expected, String actual) {
+        assertTrue("\nExpected: " + expected + "\nActual:   " + actual, actual.startsWith(expected));
+    }
+
+}
```
sashapolo, line 178: Missing line break, please check your IDE settings

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+
+        // then
+        assertSimilar(buildExpectedString(msg, expectedCategory), result);
+    }
+
+    void assertFormattedOutput(String category, String expectedCategory) {
+        assertFormattedOutput("Power overwhelming", category, expectedCategory);
+    }
+
+    static final DateTimeFormatter ISO_DATE_FORMAT =
```
sashapolo, line 149: What is this variable for?

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+ * Java log formatter test.
+ */
+@GridCommonTest(group = "Logger")
+public class JavaLoggerFormatterTest {
+
+    private static JavaLoggerFormatter formatter;
+
+    @BeforeClass
+    public static void beforeClass() throws Exception {
+        formatter = new JavaLoggerFormatter();
```
sashapolo, line 43: what's the point of having this method? `formatter` field can be initialized directly in the declaration

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+        System.setProperty("IGNITE_JAVA_LOGGER_DATE_FORMAT", "foobar");
+
+        // when
+        DateTimeFormatter fmt = JavaLoggerFormatter.resolveDateFormat();
+
+        // then
+        assertEquals(BASE_DATE, fmt.format(Instant.ofEpochMilli(BASE_MILLIS)));
+    }
+
+    void assertFormattedOutput(String msg, String category, String expectedCategory) {
```
sashapolo, line 134: Please make all helper methods in this class `private`

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
```diff
@@ -19,47 +19,72 @@
...
-    public static final String ANONYMOUS_LOGGER_NAME = "UNKNOWN";
+    static final String ANONYMOUS_LOGGER_NAME = "UNKNOWN";
+
+    /** Default date-time format. */
+    static final DateTimeFormatter DEFAULT_DATE_FMT = IgniteUtils.ISO_DATE_FMT;
+
+    /** Log message date-time formatter. */
+    static final DateTimeFormatter DATE_FMT = resolveDateFormat();
+
+    /** Property for overriding default dat-time format. */
```
sashapolo, line 44:
```suggestion
    /** Property for overriding default date-time format. */
```

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
```diff
@@ -19,47 +19,72 @@
...
+    static final String IGNITE_JUL_DATE_FORMAT_PROPERTY = "IGNITE_JAVA_LOGGER_DATE_FORMAT";
 
     /** {@inheritDoc} */
     @Override public String format(LogRecord record) {
         String threadName = Thread.currentThread().getName();
 
         String logName = record.getLoggerName();
 
-        if (logName == null)
+        if (logName == null || logName.equals(""))
```
sashapolo, line 53: `logName.isEmpty()`?

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
```diff
@@ -19,47 +19,72 @@
...
+    static final String ANONYMOUS_LOGGER_NAME = "UNKNOWN";
+
+    /** Default date-time format. */
+    static final DateTimeFormatter DEFAULT_DATE_FMT = IgniteUtils.ISO_DATE_FMT;
+
+    /** Log message date-time formatter. */
+    static final DateTimeFormatter DATE_FMT = resolveDateFormat();
+
+    /** Property for overriding default dat-time format. */
+    static final String IGNITE_JUL_DATE_FORMAT_PROPERTY = "IGNITE_JAVA_LOGGER_DATE_FORMAT";
```
sashapolo, line 45: How is this property related to the issue described in the ticket? What problem are we solving here?

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+        return "[" +
+            BASE_DATE + "][" +
+            Level.INFO + "][" +
+            Thread.currentThread().getName() + "][" +
+            category + "] " +
+            message;
+    }
+
+    static void assertSimilar(String expected, String actual) {
+        assertTrue("\nExpected: " + expected + "\nActual:   " + actual, actual.startsWith(expected));
```
sashapolo, line 175: Can be written shorter: `assertThat(actual, startsWith(expected));`

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+    static String buildExpectedString(@NotNull String message, @Nullable String category) {
+        return "[" +
+            BASE_DATE + "][" +
+            Level.INFO + "][" +
+            Thread.currentThread().getName() + "][" +
+            category + "] " +
+            message;
+    }
+
+    static void assertSimilar(String expected, String actual) {
```
sashapolo, line 174: I'd propose to rename this method to `assertStartsWith`

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+    @Test
+    public void testThrowable() {
+        // given
+        final String msg = "Greed is good";
+        final String category = "probe";
+        final String exceptionMsg = "Not enough minerals";
+
+        LogRecord record = buildLogRecord(msg, category);
+        try {
+            throw new RuntimeException(exceptionMsg);
```
sashapolo, line 90: Why do you need to "throw" this exception? What's wrong with `record.setThrown(new RuntimeException(exceptionMsg)`?

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
```diff
@@ -0,0 +1,178 @@
...
+        assertSimilar(expected, result);
+    }
+
+    /**
+     * Should use date format from property
+     */
+    @Test
+    public void testCorrectCustomFormatProperty() {
+        // given
+        System.setProperty("IGNITE_JAVA_LOGGER_DATE_FORMAT", "yyyy/MM/dd HH:mm:ss");
```
sashapolo, line 110: There's a `WithSystemProperty` annotation, I think it should be used instead (but current approach is also fine)

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
```diff
@@ -19,47 +19,72 @@
...
 
-            ex = "\n" + stackTrace;
+    @NotNull
+    static DateTimeFormatter resolveDateFormat() {
+        String customDateFormat = IgniteSystemProperties.getString(IGNITE_JUL_DATE_FORMAT_PROPERTY);
+        if (customDateFormat != null) {
+            try {
+                return DateTimeFormatter.ofPattern(customDateFormat).withZone(ZoneId.systemDefault());
+            } catch (Exception e) {
+                System.err.println("Incorrect JavaLogger date-time format: " + customDateFormat);
```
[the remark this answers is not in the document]
solveme: I guess that throwing an exception might fail node startup, we don't need fail fast here

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
```diff
@@ -19,47 +19,72 @@
...
+            t.printStackTrace(new PrintWriter(sw));
+        }
 
-            String stackTrace = sw.toString();
+        return sw.toString();
+    }
 
-            ex = "\n" + stackTrace;
+    @NotNull
+    static DateTimeFormatter resolveDateFormat() {
```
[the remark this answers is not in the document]
solveme: made it package to allow access for testing purpose

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
solveme: indeed, will change it

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
solveme: I thought about it also, but ISO_LOCAL_DATE_TIME doesn't print millis

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
solveme: Agree, made out of habit

modules/core/src/test/java/org/apache/ignite/logger/java/JavaLoggerFormatterTest.java
solveme: This formatter was created for composing expected date format. I defined it to guard us from occasional change of date format inside JavaLoggerFormatter (e.g. you can't use code under test for testing this code). But seems, indeed, usage of IgniteUtils.ISO_DATE_FMT should be safe enough

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
solveme: Will do

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
sashapolo: Why do we need millis, since ISO_LOCAL_DATE_TIME prints nanoseconds?

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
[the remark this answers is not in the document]
sashapolo: oh, it behaves differently depending on the system variable, missed that

modules/core/src/main/java/org/apache/ignite/logger/java/JavaLoggerFormatter.java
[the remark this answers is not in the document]
sashapolo: How will this failure look like then? Why do we print something to stderr? Is this behavior consistent with other loggers?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
solveme: Are you sure about nanos? From ISO_LOCAL_DATE_TIME javadoc:
The ISO date-time formatter that formats or parses a date-time without an offset, such as '2011-12-03T10:15:30'.

Anyway, millis are useful when you need to understand what happened first (there can be too many events in during whole second), but nanos would be an overkill.

Moreover proposed format yyyy-MM-dd'T'HH:mm:ss,SSS matches our default format used by ignite-log4j2:
<PatternLayout pattern="[%d{ISO8601}][%-5p][%t][%c{1}]%notEmpty{[%markerSimpleName]} %m%n"/>

2023-05-08T09:04:06,626 INFO  [pgrunner-#node-1] [o.a.i.i.IgniteKernal%node-1] Logger: Log4J2Logger

sashapolo: > Are you sure about nanos?

yes, because it uses `ISO_LOCAL_TIME` internally, which states `One to nine digits for the nano-of-second. As many digits will be output as required`

Anyway, I won't insist, current approach is also fine, so it's up to you

sashapolo (approved):

PR #2802, opened 2023-05-10
GG-36637 Update DeactivateCommand.java

josephdanrev:
Improve code to handle particular case with test


sergey-chugunov-1985 (approved):
Patch looks good to me, it is ready to be merged to master branch.

PR #2804, opened 2023-05-15
GG-34108 Fixed REST authentication failure

sk0x50:
https://ggsystems.atlassian.net/browse/GG-34108

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/security/client/AdditionalSecurityCheckTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/security/client/AdditionalSecurityCheckTest.java
@@ -16,25 +16,52 @@
 
 package org.apache.ignite.internal.processors.security.client;
 
+import java.util.concurrent.CountDownLatch;
+import java.util.concurrent.atomic.AtomicReference;
 import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteAuthenticationException;
+import org.apache.ignite.IgniteException;
 import org.apache.ignite.Ignition;
 import org.apache.ignite.client.ClientAuthenticationException;
 import org.apache.ignite.client.IgniteClient;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.internal.IgniteEx;
 import org.apache.ignite.internal.client.GridClient;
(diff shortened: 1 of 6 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/security/client/AdditionalSecurityCheckTest.java
```diff
@@ -152,4 +179,61 @@ public void testAdditionalPasswordServerFail() throws Exception {
...
+
+        assertTrue("Failed to wait for starting node.", startLatch.await(10, SECONDS));
+
+        try (GridClient client = GridClientFactory.start(getGridClientConfiguration())) {
+            err.set(client.checkLastError());
+        }
+
+        contLatch.countDown();
+
+        startFut.get();
```
tkalkirill, line 233: Please add timeout

modules/core/src/test/java/org/apache/ignite/internal/processors/security/client/AdditionalSecurityCheckTest.java
```diff
@@ -152,4 +179,61 @@ public void testAdditionalPasswordServerFail() throws Exception {
...
+
+        Exception unexpectedErr = err.get();
+
+        assertNull("Unexpected error [err=" + unexpectedErr + ']', unexpectedErr);
```
tkalkirill, line 237:
```suggestion
        Exception unexpectedErr = err.get();

        assertNull(err.get());
```

tkalkirill (approved):

modules/core/src/test/java/org/apache/ignite/internal/processors/security/client/AdditionalSecurityCheckTest.java
sk0x50: I don't like assertions without a message. On the other hand, in general, we should avoid reading err atomics twice. So, I would like to stay with the current code.

modules/core/src/test/java/org/apache/ignite/internal/processors/security/client/AdditionalSecurityCheckTest.java
sk0x50: Yep, will fix it.

PR #2809, opened 2023-05-30
GG-36614 Duplicated index creation using SQL and already existing QueryFields index leads to node start-up failure

zstan:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
```diff
@@ -1059,48 +1055,51 @@ public boolean rebuildFromHashInProgress() {
...
     /**
      * Add index that is in an intermediate state and is still being built, thus is not used in queries until it is
      * promoted.
      *
      * @param idx Index to add.
+     * @return {@code true} If index not intersects by name with already exists one.
      * @throws IgniteCheckedException If failed.
      */
-    public void proposeUserIndex(Index idx) throws IgniteCheckedException {
+    public boolean proposeUserIndex(Index idx, boolean ifNotExists) throws IgniteCheckedException {
```
xtern, line 1079: Where `ifNotExists` flag is used in this method? :thinking:

zstan:
@xtern thanks for review, all issues are fixed, TC in progress. Can i move forward ?

xtern:
Thanks for the contribution, in general it looks like the patch completely solves the main problem described in the issue. 

But as was discussed privately, we need to file a ticket about incorrect `IF NOT EXISTS` behavior, when index name has not been specified.

xtern (approved):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/opt/GridH2Table.java
```diff
@@ -1059,48 +1055,51 @@ public boolean rebuildFromHashInProgress() {
...
+
             Index oldTmpIdx = tmpIdxs.put(idx.getName(), (GridH2IndexBase)idx);
 
             assert oldTmpIdx == null;
         }
         finally {
             unlock(true);
         }
+
+        return true;
```
AMashenkov, line 1102: Looks like the method always returns true.

zstan: got it

PR #2814, opened 2023-06-07
GG-36785 [IGNITE-14257] Add copy constructor to ClientCacheConfiguration

myskov:
https://ggsystems.atlassian.net/browse/GG-36785


isapego (approved):
Overall looks good, though I'd added a test for copy constructor to make sure we won't miss adding any new fields to it in future.

PR #2818, opened 2023-06-15
GG-34913 Include consistentId in local node metrics log

josephdanrev:
https://ggsystems.atlassian.net/browse/GG-34913

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
@@ -2237,7 +2237,7 @@ public class IgniteKernal implements IgniteEx, IgniteMXBean, Externalizable {
 
             msg.nl()
                 .a("Metrics for local node (to disable set 'metricsLogFrequency' to 0)").nl()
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2237,7 +2237,7 @@ private void ackNodeMetrics(DecimalFormat dblFmt,
 
             msg.nl()
                 .a("Metrics for local node (to disable set 'metricsLogFrequency' to 0)").nl()
-                .a("    ^-- Node [id=").a(id).a(name() != null ? ", name=" + name() : "").a(", uptime=")
+                .a("    ^-- Node [id=").a(id).a(", consistentId=").a(localNode().consistentId()).a(name() != null ? ", name=" + name() : "").a(", uptime=")
```
sk0x50, line 2240: It seems to me, you can use the `locNode` variable that is already defined in the scope of this method instead of using `localNode()`

sk0x50 (approved):

PR #2830, opened 2023-06-27
GG-36810 SQL: JSON path functions

vadimkolodingg:


modules/h2/pom.xml
```diff
@@ -70,6 +70,27 @@
...
+                <exclusion>
+                    <groupId>org.slf4j</groupId>
+                    <artifactId>slf4j-api</artifactId>
+                </exclusion>
+            </exclusions>
+        </dependency>
+        <dependency>
+            <groupId>net.minidev</groupId>
+            <artifactId>json-smart</artifactId>
+            <version>2.4.10</version>
```
ygerzhedovich, line 87: why 2.4.10 version? As I see it contains few CVE. Also available 2.5.0 version

PR #2832, opened 2023-06-28
GG-36879 Dynamic column re-creation ignores modified column constraints.

xtern:
This patch includes fix for both `ADD COLUMN` (GG-36879) and `DROP COLUMN` (GG-36907) commands.

ADD COLUMN issue:
We forgot to set the `precision` and `scale` for the added column in `QueryEntity`, so after loading it from disk, we got the previous value.
Note: we do not support ADD COLUMN with a default value, so there is no need to set it.

DROP COLUMN issue:
We forgot to clean scale/precision/notnull/default column attributes in `QueryEntity` when removing field. So after reading entity from disk, we can get the previous values if column was re-created.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
@@ -1671,7 +1671,23 @@ public class QueryUtils {
      * @return {@code true} if the field is removed. Otherwise returns {@code false}.
      */
     public static boolean removeField(QueryEntity entity, String alias) {
-        return entity.getFields().remove(fieldNameByAlias(entity, alias)) != null;
+        String fieldName = fieldNameByAlias(entity, alias);
+
+        if (entity.getFields().remove(fieldName) != null) {
+            Set<String> notNUllFields = entity.getNotNullFields();
+
+            if (notNUllFields != null) {
+                notNUllFields.remove(fieldName);
+            }
+
+            entity.getDefaultFieldValues().remove(fieldName);
+            entity.getFieldsPrecision().remove(fieldName);
+            entity.getFieldsScale().remove(fieldName);
+
+            return true;
+        }
+
+        return false;
     }
 
     /**
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
@@ -1674,10 +1674,10 @@ public class QueryUtils {
         String fieldName = fieldNameByAlias(entity, alias);
 
         if (entity.getFields().remove(fieldName) != null) {
-            Set<String> notNUllFields = entity.getNotNullFields();
+            Set<String> notNullFields = entity.getNotNullFields();
 
-            if (notNUllFields != null) {
-                notNUllFields.remove(fieldName);
+            if (notNullFields != null) {
+                notNullFields.remove(fieldName);
             }
 
             entity.getDefaultFieldValues().remove(fieldName);
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
@@ -1117,12 +1117,12 @@ public class CommandProcessor {
                 dfltValues.put(e.getKey(), dfltVal);
 
             int precisionVal = convertH2ColumnPrecision(col.getType());
-            if (precisionVal != -1) {
+            if (precisionVal != QueryField.UNDEFINED_PRECISION) {
                 precision.put(e.getKey(), precisionVal);
             }
 
             int scaleVal = convertH2ColumnScale(col.getType());
-            if (scaleVal != -1) {
+            if (scaleVal != QueryField.UNDEFINED_SCALE) {
                 scale.put(e.getKey(), scaleVal);
             }
         }
@@ -1216,7 +1216,7 @@ public class CommandProcessor {
                 return (int)type.getPrecision();
         }
 
-        return -1;
+        return QueryField.UNDEFINED_PRECISION;
     }
 
     private static int convertH2ColumnScale(TypeInfo type) {
@@ -1224,7 +1224,7 @@ public class CommandProcessor {
             type.getScale() < H2Utils.DECIMAL_DEFAULT_SCALE)
             return type.getScale();
 
-        return -1;
+        return QueryField.UNDEFINED_SCALE;
     }
 
     /**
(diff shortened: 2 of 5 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryUtils.java
```diff
@@ -1671,7 +1671,23 @@ public static String fieldNameByAlias(QueryEntity entity, String alias) {
      * @return {@code true} if the field is removed. Otherwise returns {@code false}.
      */
     public static boolean removeField(QueryEntity entity, String alias) {
-        return entity.getFields().remove(fieldNameByAlias(entity, alias)) != null;
+        String fieldName = fieldNameByAlias(entity, alias);
+
+        if (entity.getFields().remove(fieldName) != null) {
+            Set<String> notNUllFields = entity.getNotNullFields();
```
zstan, line 1677:
```suggestion
            Set<String> notNullFields = entity.getNotNullFields();
```

xtern: Fixed, thanks.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -889,10 +889,30 @@ else if (cmdH2 instanceof GridSqlAlterTableAddColumn) {
...
 
+                        int precision = -1;
+                        int scale = -1;
+
+                        if (col.column().getType().getValueType() == Value.DECIMAL) {
+                            if (col.precision() < H2Utils.DECIMAL_DEFAULT_PRECISION) {
+                                precision = col.precision();
+                            }
+
+                            if (col.scale() < H2Utils.DECIMAL_DEFAULT_SCALE) {
```
zstan, line 900: why you operate with "<" and not with "!=" here ?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/IgnitePersistentStoreSchemaLoadTest.java
```diff
@@ -271,28 +247,81 @@ private CountDownLatch checkpointLatch(IgniteEx node) {
...
+        runSql("alter table \"Person\" drop column " +
+            "\"city\", \"str1\", \"str2\", \"num1\", \"num2\"", node, schema);
 
-        node.context().query().querySqlFields(
-            new SqlFieldsQuery("alter table \"Person\" drop column \"city\"").setSchema(schema), false)
-            .getAll();
+        runSql("alter table \"Person\" add column (" +
+                "\"rate\" decimal(10, 2) not null," +
+                "\"str1\" char(10) not null," +
+                "\"str2\" varchar," +
```
zstan, line 258: seems TIMESTAMP(5,1) already have precision but it not processed for now, probably some other types are not covered too ...

xtern: I tried creating timestamp/float with custom precision/scale but it looks like we don't support (ignoring) precision/scale for TIMESTAMP/FLOAT.

For the following table
```
CREATE TABLE Person(ID INTEGER PRIMARY KEY, TIME1 TIMESTAMP, TIME2 TIMESTAMP(10,3), TIME3 FLOAT(5))
```

the following query

```
SELECT * FROM sys.table_columns WHERE column_name LIKE 'TIME%'
```

 gives the following results:

 column | table | schema | affinity | incriment | default | nullable | PK | precision | scale | type
 ---------------------------------------------------------------------------------------------------
TIME1 | PERSON | PUBLIC | false | false | null | true | false | -1 | -1 | java.sql.Timestamp
TIME2 | PERSON | PUBLIC | false | false | null | true | false | -1 | -1 | java.sql.Timestamp
TIME3 | PERSON | PUBLIC | false | false | null | true | false | -1 | -1 | java.lang.Float

Thus, they are not affected by this issue.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
xtern: This code is already in use, I have extracted it into a separate method for clarity.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/IgnitePersistentStoreSchemaLoadTest.java
xtern: I added a timestamp/float into a test to avoid possible confusions in the future.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/IgnitePersistentStoreSchemaLoadTest.java
```diff
@@ -321,27 +361,40 @@ public Person(int id) {
...
+        /** */
+        @QuerySqlField
+        protected Timestamp time1;
 
-        /** {@inheritDoc} */
-        @Override public int hashCode() {
-            int res = id;
+        /** */
+        @QuerySqlField(notNull = true)
+        protected Timestamp time2;
```
zstan, line 390: what about @QuerySqlField(precision = 10, notNull = true) ?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/IgnitePersistentStoreSchemaLoadTest.java
xtern: After discussion, the test changes were reverted back :roll_eyes: .

modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/IgnitePersistentStoreSchemaLoadTest.java
xtern: After discussion, the test changes were reverted back roll_eyes

zstan (approved):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -1204,6 +1204,29 @@ else if (createTbl.wrapKey()) {
...
+                return (int)type.getPrecision();
+        }
+        else if (type.getValueType() == Value.STRING ||
+            type.getValueType() == Value.STRING_FIXED ||
+            type.getValueType() == Value.STRING_IGNORECASE) {
+            if (type.getPrecision() < H2Utils.STRING_DEFAULT_PRECISION)
+                return (int)type.getPrecision();
+        }
+
+        return -1;
```
AMashenkov, line 1219: Let's avoid magic numbers and add add a constants `UNDEFINED_PRECISION` & `UNDEFINED_SCALE`

xtern: Fixed, thanks.

AMashenkov (approved):

PR #2835, opened 2023-07-01
GG-35711 ignite.sh ignores -DIGNITE_QUIET=false specified via JVM_OPTS

josephdanrev:
https://ggsystems.atlassian.net/browse/GG-35711


solveme:
LGTM

PR #2839, opened 2023-07-17
GG-36991 Remove unnecessary commons-logging dependency from ignite-spring module

solveme:


modules/jcl/pom.xml
```diff
@@ -43,7 +43,7 @@
         <dependency>
             <groupId>commons-logging</groupId>
             <artifactId>commons-logging</artifactId>
-            <version>1.1.1</version>
+            <version>1.2</version>
```
sashapolo, line 46: I don't understand this. Why do we still have this dependency if the ticket states that it should be removed?

solveme: The goal is to remove commons-logging from ignite-spring. ignite-jcl still requires this dependency (e.g. what if user wants to run GG without ignite-spring but with ignite-jcl). Task to remove ignite-jcl is already in backlog for 8.9.x

sashapolo (approved):

modules/jcl/pom.xml
sashapolo: ok, got it

PR #2840, opened 2023-07-25
GG-36345 Disabling WAL for index rebuilding activity

sergey-chugunov-1985:


modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/WalDisabledDuringIndexRecreateTest.java
```diff
@@ -0,0 +1,404 @@
...
+                }
+            };
+        };
+
+        try {
+            testLog = new ListeningTestLogger(log);
+
+            awaitRebuild();
+
+            assertTrue("Rebuild must not succeed", false);
```
zstan, line 200: fail(String msg) need to be used instead

modules/indexing/src/test/java/org/apache/ignite/internal/processors/database/WalDisabledDuringIndexRecreateTest.java
```diff
@@ -0,0 +1,404 @@
...
+        FileWALPointer walStartPtr = createDataAndDeleteIndexBin();
+
+        testLog = new ListeningTestLogger(log);
+
+        awaitRebuild();
+
+        assertEquals(
+            0,
+            countWalGrpRecords(wp -> ((FileWALPointer)wp).compareTo(walStartPtr) > 0, CU.cacheGroupId(cacheName(), cacheGroupName()))
+        );
```
zstan, line 143: probably we need additional check that index is in operable\enabled state after rebuid ? and further new kv updates will touch index at all

zstan (commented):
i fill some comments, plz check them

PR #2845, opened 2023-08-07
GG-36429 Currently there are a few gaps to cover in terms of how we generate, check and manage product licenses. It's required to be addressed.

JKonSir:
-- added task to collect PDS allocation size metrics
-- added methods to collect host RAM size and allocated PDS size


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2271,6 +2296,25 @@ private void ackNodeMetrics(DecimalFormat dblFmt,
         }
     }
 
+    private long allocatedPDSSize() {
+        DataStorageConfiguration dsCfg = ctx.config().getDataStorageConfiguration();
+
+        if (dsCfg == null)
```
ibelyakov, line 2302: Based on the IgnitionEx#initializeDataStorageConfiguration(cfg) implementation the DataStorageConfiguration can't be null as long as it's not a client node. Should we add "ctx.clientNode()" check here?

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1469,6 +1504,26 @@ private static Iterable<ClusterNode> nodesPerJvm(Iterable<ClusterNode> nodes) {
         return grpMap.values();
     }
 
+    /**
+     * Returns one representative node for each host.
+     *
+     * @param nodes Nodes.
+     * @return Collection which contains only one representative node for each host.
+     */
+    private static Iterable<ClusterNode> nodesPerHost(Iterable<ClusterNode> nodes) {
```
ibelyakov, line 1513: It looks like that we already have the "IgniteUtils#neighborhood(nodes)" method, which does a similar thing and which is already used in the GridEntLicenseProcessor.

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
```diff
@@ -1449,6 +1451,39 @@ public static double offheapSize(Iterable<ClusterNode> nodes, int precision) {
...
+     * Gets total allocated persistence size in GB rounded to specified precision.
+     *
+     * @param nodes Nodes.
+     * @param precision Precision.
+     * @return Total allocated PDS size in GB.
+     */
+    public static double allocatedPDSSize(Iterable<ClusterNode> nodes, int precision) {
+        double totalAllocatedPDS = .0;
+
+        for (ClusterNode n : nodesPerJvm(nodes)) {
```
ibelyakov, line 1464: Shouldn't we summarize the allocated pds size for each node in one Jvm? Since they will use separate persistent directories.

ibelyakov (commented):
Please check the comments in the PR.

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2271,6 +2296,25 @@ private void ackNodeMetrics(DecimalFormat dblFmt,
...
+            return 0;
+
+        long res = ctx.grid().dataRegionMetrics(DFLT_DATA_REG_DEFAULT_NAME).getTotalAllocatedSize();
+
+        DataRegionConfiguration[] dataRegions = dsCfg.getDataRegionConfigurations();
+
+        if (dataRegions != null) {
+            for (DataRegionConfiguration dataReg : dataRegions) {
```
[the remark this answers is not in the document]
JKonSir: added the filter

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
JKonSir: added client node check

modules/core/src/main/java/org/apache/ignite/internal/util/IgniteUtils.java
JKonSir: Thank you for this. It's my rude mistake. It's fixed to summarize the allocated PDS size of all non-client nodes.

PR #2849, opened 2023-08-10
GG-37101 Improve logging when message deserialization fails

rpuch:
https://ggsystems.atlassian.net/browse/GG-37101

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
@@ -19,6 +19,7 @@ package org.apache.ignite.internal.direct.stream.v2;
 import java.lang.reflect.Array;
 import java.nio.ByteBuffer;
 import java.util.ArrayList;
+import java.util.Arrays;
 import java.util.BitSet;
 import java.util.Collection;
 import java.util.Iterator;
@@ -271,10 +272,10 @@ public class DirectByteBufferStreamImplV2 implements DirectByteBufferStream {
     private int readItems;
 
     /** */
-    private Object[] objArr;
+    Object[] objArr;
 
     /** */
-    private Collection<Object> col;
+    List<Object> col;
 
     /** */
     private Map<Object, Object> map;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -271,10 +272,10 @@ public class DirectByteBufferStreamImplV2 implements DirectByteBufferStream {
     private int readItems;
 
     /** */
-    private Object[] objArr;
+    Object[] objArr;
```
tkalkirill, line 275: I propose to return the private access modifier, use `GridTestUtils#setFieldValue` in tests.

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
```diff
@@ -1803,7 +1804,42 @@ protected Object read(MessageCollectionItemType type, MessageReader reader) {
...
+    }
+
+    /**
+     * Builds a string showing the simple types of the elements of the list; if the list is longer than the
+     * limit, then only its final elements are considered and ellipsis represents the omitted part of the list.
+     *
+     * @param list List to represent.
+     * @return String like [Integer, String] if there are exactly 2 elements of types Integer and String, respectively.
+     */
+    private String lastTypesWithLimit(List<Object> list) {
```
tkalkirill, line 1820: Do I understand correctly that this method can be static, and we can specifically test it?

I propose specifically to test this method better and make the limit an argument of the method, as an option, move this method to an auxiliary class.

modules/core/src/test/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2Test.java
```diff
@@ -0,0 +1,102 @@
+/*
+* Copyright 2019 GridGain Systems, Inc. and Contributors.
```
tkalkirill, line 2:
```suggestion
* Copyright 2023 GridGain Systems, Inc. and Contributors.
```

modules/core/src/main/java/org/apache/ignite/internal/direct/stream/v2/DirectByteBufferStreamImplV2.java
rpuch: Made the method static and package local to make it easier to test it. I don't think it should be extracted to a common tool yet as it's only used in one class for now. If we need it elsewhere, we'll do it.

modules/core/src/test/java/org/apache/ignite/testsuites/IgniteMarshallerSelfTestSuite.java
```diff
@@ -33,6 +34,7 @@
 import org.junit.runner.RunWith;
 import org.junit.runners.Suite;
 
+
```
tkalkirill, line 37:
```suggestion
```

tkalkirill (approved):

PR #2860, opened 2023-08-30
GG-36947 ContainsKey operation should not fetch entry value into heap

sk0x50:
https://ggsystems.atlassian.net/browse/GG-36947

This patch addresses **containsKey** and **containsKeys** for atomic caches. This optimization is not used when the on-heap cache is enabled or platform cache is used. Neither of these methods triggers removing expired entries that were touched during execution.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
@@ -39,7 +39,11 @@ public class CacheObjectShadow extends CacheObjectAdapter {
         valBytes = EMPTY_BYTES;
     }
 
-    /** */
+    /**
+     * Creates a new instance of CacheObjectShadow with the given type.
+     *
+     * @param type Type of cache object.
+     **/
     public CacheObjectShadow(byte type) {
         this.type = type;
         valBytes = EMPTY_BYTES;
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
@@ -27,7 +27,7 @@ import static org.apache.ignite.internal.util.IgniteUtils.EMPTY_BYTES;
  */
 public class CacheObjectShadow extends CacheObjectAdapter {
     /** */
-    private static final long  serialVersionUID = 0L;
+    private static final long serialVersionUID = 0L;
 
     /** Type of cache object. */
     private byte type;
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
@@ -1,0 +1,87 @@
+/*
+ * Copyright 2023 GridGain Systems, Inc. and Contributors.
+ *
+ * Licensed under the GridGain Community Edition License (the "License");
+ * you may not use this file except in compliance with the License.
+ * You may obtain a copy of the License at
+ *
+ *     https://www.gridgain.com/products/software/community-edition/gridgain-community-edition-license
+ *
+ * Unless required by applicable law or agreed to in writing, software
+ * distributed under the License is distributed on an "AS IS" BASIS,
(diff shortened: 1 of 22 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheIgniteOutOfMemoryExceptionTest.java
```diff
@@ -200,10 +251,250 @@ public void testHugeEntry() throws Exception {
...
+            cache.put(k, new byte[blobSize]);
+
+        long usedMemory = Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
+        long totalFreeMemory = Runtime.getRuntime().maxMemory() - usedMemory;
+
+        // Let's occupy all free memory.
+        List<Object> unused = new ArrayList<>();
+        IgniteBiTuple<HotSpotDiagnosticMXBean, VMOption> mbean = disableHeapDumpOnOutOfMemoryError();
+        try {
+            while (blobSize < totalFreeMemory) {
```
ibessonov, line 438: This condition doesn't include possible GC

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheIgniteOutOfMemoryExceptionTest.java
```diff
@@ -200,10 +251,250 @@ public void testHugeEntry() throws Exception {
...
+        int blobSize = (int) (512 / keys.size() * U.MB);
+
+        for (Integer k : keys)
+            cache.put(k, new byte[blobSize]);
+
+        long usedMemory = Runtime.getRuntime().totalMemory() - Runtime.getRuntime().freeMemory();
+        long totalFreeMemory = Runtime.getRuntime().maxMemory() - usedMemory;
+
+        // Let's occupy all free memory.
+        List<Object> unused = new ArrayList<>();
```
ibessonov, line 435: There's a change that JIT will remove this variable, because no one reads it. You can add some assertion at the end, that it's not empty, for example

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheObjectShadow.java
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    /** {@inheritDoc} */
+    @Override public <T> @Nullable T value(CacheObjectValueContext ctx, boolean cpy) {
+        return null;
+    }
+
+    /** {@inheritDoc} */
+    @Override public <T> @Nullable T value(CacheObjectValueContext ctx, boolean cpy, ClassLoader ldr) {
+        return null;
```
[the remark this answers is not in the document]
sk0x50: Makes sense to me. Fixed

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheIgniteOutOfMemoryExceptionTest.java
sk0x50: This loop was reworked.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheIgniteOutOfMemoryExceptionTest.java
sk0x50: Yes, you are right. The corresponding check/usage was added.

ibessonov (approved):

vldpyatkov (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManager.java
```diff
@@ -136,6 +136,14 @@ public interface IgniteCacheOffheapManager {
      */
     @Nullable public CacheDataRow read(GridCacheContext cctx, KeyCacheObject key) throws IgniteCheckedException;
 
+    /**
+     * @param cctx Cache context.
+     * @param key Key.
+     * @return Cached row, if available, null otherwise.
+     * @throws IgniteCheckedException If failed.
+     */
+    @Nullable public CacheDataRow find(GridCacheContext cctx, KeyCacheObject key) throws IgniteCheckedException;
```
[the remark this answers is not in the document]
sk0x50: Fixed

PR #2863, opened 2023-09-04
GG-36938 import/export iceberg

vadimkolodingg:
https://ggsystems.atlassian.net/browse/GG-36938

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/sql/command/SqlBulkLoadCommand.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/sql/command/SqlBulkLoadCommand.java
@@ -19,6 +19,7 @@ package org.apache.ignite.internal.sql.command;
 import org.apache.ignite.internal.processors.bulkload.BulkLoadCsvFormat;
 import org.apache.ignite.internal.processors.bulkload.BulkLoadFormat;
 import org.apache.ignite.internal.processors.bulkload.BulkLoadAckClientParameters;
+import org.apache.ignite.internal.processors.bulkload.BulkLoadIcebergFormat;
 import org.apache.ignite.internal.processors.bulkload.BulkLoadLocation;
 import org.apache.ignite.internal.processors.bulkload.BulkLoadLocationFile;
 import org.apache.ignite.internal.processors.bulkload.BulkLoadLocationQuery;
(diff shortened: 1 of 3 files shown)


modules/core/src/main/java/org/apache/ignite/internal/sql/command/SqlBulkLoadCommand.java
```diff
@@ -335,6 +350,35 @@ private void parseParameters(SqlLexer lex) {
...
+     * @return properties map.
+     */
+    private static Map<String, String> parseProperties(SqlLexer lex) {
+        if (!"PROPERTIES".equals(lex.lookAhead().token())) {
+            return Collections.emptyMap();
+        }
+        Map<String, String> props = new HashMap<>();
+        skipIfMatchesKeyword(lex, "PROPERTIES");
+        skipIfMatches(lex, SqlLexerTokenType.PARENTHESIS_LEFT);
+        do {
```
ygerzhedovich, line 365:
```suggestion

        skipIfMatchesKeyword(lex, "PROPERTIES");
        skipIfMatches(lex, SqlLexerTokenType.PARENTHESIS_LEFT);

        Map<String, String> props = new HashMap<>();
        
        do {
```

modules/core/src/main/java/org/apache/ignite/internal/sql/command/SqlBulkLoadCommand.java
```diff
@@ -412,6 +456,16 @@ public BulkLoadLocation into() {
         return into;
     }
 
+    /**
+     * Propertes (case-sensitive) passed to command.
+     * Syntax: PROPERTIES ('k1' = 'v1', 'k2' = 'v2', ...)
+     * Properties keys/values are case-sensitive.
+     * @return immutable properties
```
ygerzhedovich, line 463:
```suggestion
     * @return Immutable properties.
```

modules/core/src/main/java/org/apache/ignite/internal/sql/command/SqlBulkLoadCommand.java
```diff
@@ -335,6 +350,35 @@ private void parseParameters(SqlLexer lex) {
         }
     }
 
+    /**
+     * Parses the optional case-sensitive properties with syntax PROPERTIES ('k1' = 'v1', 'k2' = 'v2', ...)
```
ygerzhedovich, line 354:
```suggestion
     * Parses the optional case-sensitive properties with syntax PROPERTIES ('k1' = 'v1', 'k2' = 'v2', ...).
```

modules/core/src/main/java/org/apache/ignite/internal/sql/command/SqlBulkLoadCommand.java
ygerzhedovich, line 463: Are you sure that is immutable? Let's add a test in case it is important.

ygerzhedovich (approved):

PR #2868, opened 2023-09-06
GG-36951 Acknowledge Communication messages faster

rpuch:
Two mechanisms are added to send message acknowledgements faster:

* When accrued size of received messages since last ack exceeds a threshold
* On timeout since last ack


ptupitsyn (approved):
.NET-related changes look good to me.

sashapolo (approved):

PR #2879, opened 2023-09-14
GG-37218 copy from/into order insensitive

vadimkolodingg:
https://ggsystems.atlassian.net/browse/GG-37218


modules/core/src/test/java/org/apache/ignite/internal/sql/SqlParserBulkLoadSelfTest.java
```diff
@@ -207,8 +207,25 @@ public void testCopyParseProperties() {
...
+                "copy from 'any.file' into Person (_key, age, firstName, lastName) format csv")
+                .nextCommand();
+        new SqlParser(null,
+                "copy into 'any.file' from Person (_key, age, firstName, lastName) format csv")
+                .nextCommand();
+
+        assertParseError(null,
+                "copy into 'any.file' into Person (_key, age, firstName, lastName) format csv",
+                "(expected both locations: \"FROM\", \"INTO\")");
+        assertParseError(null,
```
ygerzhedovich, line 227: I don't see reverse order test, like a `copy from Person (_key, age, firstName, lastName) format csv into 'any.file'`, and so one

ygerzhedovich (approved):

PR #2881, opened 2023-09-18
GG-37304 .NET: Add UnwrapNullablePrimitiveTypes binary configuration mode

ptupitsyn:
**Problem**

GridGain.NET (in any mode - thin client, thick client, server) writes object fields of nullable primitive types (such as `int?`) using `BinaryWriter.WriteObject`, which is inconsistent with Java behavior, where `BinaryFieldAccessor` uses non-public methods in `BinaryWriterExImpl` to write nullable primitives, producing different metadata.

**Solution**

Add `BinaryConfiguration.UnwrapNullablePrimitiveTypes` flag which changes the behavior to match Java
* .NET specific
* Works for thin client and embedded modes the same way
* Off by default (compatibility), but recommended to enable when possible
* Log a warning when not enabled


modules/core/src/test/java/org/apache/ignite/platform/PlatformNullablePrimitivesTask.java
```diff
@@ -0,0 +1,177 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
isapego, line 2: I believe, we should put actual date in new files?
```suggestion
 * Copyright 2023 GridGain Systems, Inc. and Contributors.
```

modules/platforms/dotnet/Apache.Ignite.Core/Ignition.cs
```diff
@@ -367,6 +367,27 @@ private static void LogDiagnosticMessages(IgniteConfiguration cfg, ILogger log)
...
+        {
+            if (log == null)
+                return;
+
+            if (cfg == null || !cfg.UnwrapNullablePrimitiveTypes)
+            {
+                log.Warn(
+                    $"{nameof(BinaryConfiguration)}.{nameof(BinaryConfiguration.UnwrapNullablePrimitiveTypes)} " +
+                    "is not enabled. It is recommended to enable this setting, " +
+                    "unless old behavior is required for compatibility reasons.");
```
isapego, line 390: I'm not sure guys from CS would approve on this. Better check with @DirectXceriD just to be sure.

isapego (approved):
Overall looks good to me.

modules/platforms/dotnet/Apache.Ignite.Core/Ignition.cs
ptupitsyn: Discussed in Slack, no objections from @slukyano and @DirectXceriD  .

PR #2882, opened 2023-09-18
GG-37364 NullPointerException in org.apache.ignite.internal.IgniteKernal#allocatedPDSSize

JKonSir:
-- add checking that cluster is ACTIVE before collect information about allocated PDS


antkr (approved):
lgtm, please proceed with merge

PR #2896, opened 2023-09-26
GG-36804 Fix partition-log-tree corruption and NPE in defragmentation.

ibessonov:
https://ggsystems.atlassian.net/browse/GG-36804

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
@@ -3077,7 +3077,7 @@ public class IgniteCacheOffheapManagerImpl implements IgniteCacheOffheapManager 
             }
 
             if (oldRow != null && oldRow.version().updateCounter() != 0) {
-                if (oldTombstone && tombstoneRow == null)
+                if (isIncrementalDrEnabled(cctx) && oldTombstone && tombstoneRow == null)
                     cctx.dr().onTombstoneCleaned(partId, oldRow.version().updateCounter());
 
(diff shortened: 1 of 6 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
```diff
@@ -3074,12 +3074,13 @@ else if (tombstoneRow != null) {
...
-                if (oldRow != null && oldRow.version().updateCounter() != 0) {
-                    if (oldTombstone && tombstoneRow == null)
-                        cctx.dr().onTombstoneCleaned(partId, oldRow.version().updateCounter());
+            }
 
-                    removeFromLog(new UpdateLogRow(cctx.cacheId(), oldRow.version().updateCounter(), oldRow.link()));
-                }
+            if (oldRow != null && oldRow.version().updateCounter() != 0) {
+                if (oldTombstone && tombstoneRow == null)
+                    cctx.dr().onTombstoneCleaned(partId, oldRow.version().updateCounter());
```
AMashenkov, line 3081:
```suggestion
                if (isIncrementalDrEnabled(cctx) && oldTombstone && tombstoneRow == null)
                    cctx.dr().onTombstoneCleaned(partId, oldRow.version().updateCounter());
```
Otherwise, if DR is disabled, it may lead to unexpected behaviour or NPE.

AMashenkov (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteCacheOffheapManagerImpl.java
ibessonov: Thank you!

sashapolo (approved):
I don't see Andrey's comment being fixed, but otherwise looks good

PR #2899, opened 2023-09-26
GG-37427 Change IGNITE_BINARY_SORT_OBJECT_FIELDS default from false to true

stasusov:


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1096,8 +1097,9 @@ public final class IgniteSystemProperties {
      * <p>
      * NOTICE: Should be the default in Apache Ignite 3.0
      */
-    @SystemProperty("Enables fields to be written by BinaryMarshaller in sorted order. " +
-        "By default, the natural order is used")
+    @SystemProperty(value = "Enables fields to be written by BinaryMarshaller in sorted order. " +
+        "By default, the natural order is used",
```
ygerzhedovich, line 1101:
```suggestion
    @SystemProperty(value = "Enables fields to be written by BinaryMarshaller in sorted order. Use natural order in case the property is disabled " +
        "By default, the sorted order is used",
```

PR #2900, opened 2023-09-26
GG-37428 Change P2P deploymentMode from SHARED to ISOLATED

stasusov:


modules/core/src/main/java/org/apache/ignite/configuration/IgniteConfiguration.java
```diff
@@ -145,7 +145,7 @@ public class IgniteConfiguration {
     public static final long DFLT_DISCOVERY_STARTUP_DELAY = 60000;
 
     /** Default deployment mode (value is {@link DeploymentMode#SHARED}). */
-    public static final DeploymentMode DFLT_DEPLOYMENT_MODE = DeploymentMode.SHARED;
+    public static final DeploymentMode DFLT_DEPLOYMENT_MODE = DeploymentMode.ISOLATED;
```
ibessonov, line 148: Javadoc of deployment mode is not updated. There it says that default is "SHARED"

stasusov: Fixed, thank you

PR #2906, opened 2023-09-26
GG-37435 Change DataStorageConfiguration#metricsEnabled from false to true

stasusov:


modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestProcessorAbstractSelfTest.java
```diff
@@ -3529,6 +3535,8 @@ private static class NodeConsistentIdFilter implements IgnitePredicate<ClusterNo
 
         if (memoryMetricsEnabled)
             dsCfg.setMetricsEnabled(true).setWalMode(NONE);
+        else
+            dsCfg.setMetricsEnabled(false);
```
sashapolo, line 3539: What about the WAL mode?

sergey-chugunov-1985: I don't think we need to set WAL mode in this case. As you can see from just three lines of code above when the flag `memoryMetricsEnabled` is set to true it also enables persistence. And vice versa - when it is set to false, persistence is disabled thus no need for WAL configuration.

sashapolo (approved):

PR #2912, opened 2023-09-28
GG-37438 [IGNITE-8801] Forbid operations with atomic caches in transactions by default

sk0x50:
https://ggsystems.atlassian.net/browse/GG-37438

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
@@ -31,9 +31,12 @@ import static org.apache.ignite.IgniteSystemProperties.IGNITE_ALLOW_ATOMIC_OPS_I
  */
 public class CacheOperationContext implements Serializable {
     /** */
-    //TODO IGNITE-8801 remove this and set default as `false`.
-    public static final boolean DFLT_ALLOW_ATOMIC_OPS_IN_TX =
-        IgniteSystemProperties.getBoolean(IGNITE_ALLOW_ATOMIC_OPS_IN_TX, true);
+    private static final boolean DFLT_ALLOW_ATOMIC_OPS_IN_TX = false;
+
+    /** */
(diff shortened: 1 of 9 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -31,9 +31,12 @@
  */
 public class CacheOperationContext implements Serializable {
     /** */
-    //TODO IGNITE-8801 remove this and set default as `false`.
-    public static final boolean DFLT_ALLOW_ATOMIC_OPS_IN_TX =
-        IgniteSystemProperties.getBoolean(IGNITE_ALLOW_ATOMIC_OPS_IN_TX, true);
+    private static final boolean DFLT_ALLOW_ATOMIC_OPS_IN_TX = false;
+
+    /** */
+    public static final boolean allowAtomicOpsInTx() {
```
vldpyatkov, line 37: The method name should be changed because we already have a method called `setAllowAtomicOpsInTx`, but it is not related to the static one. How about `isAtomicOpsAllowInTxByDefault`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
```diff
@@ -31,9 +31,12 @@
...
+        return IgniteSystemProperties.getBoolean(IGNITE_ALLOW_ATOMIC_OPS_IN_TX, DFLT_ALLOW_ATOMIC_OPS_IN_TX);
```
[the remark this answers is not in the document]
sk0x50: Thanks! Fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheOperationContext.java
sk0x50: Renamed to `defaultAllowAtomicOpsInTx()`

vldpyatkov (approved):

PR #2913, opened 2023-09-28
GG-37433 Default value of IGNITE_QUIET was changed to false.

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-37433

### DIFF
--- a/bin/include/parseargs.sh
+++ b/bin/include/parseargs.sh
@@ -43,7 +43,7 @@ fi
 
 CONFIG=${DEFAULT_CONFIG:-}
 INTERACTIVE="0"
-QUIET="-DIGNITE_QUIET=true"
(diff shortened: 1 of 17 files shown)


sk0x50 (dismissed):

sk0x50:
Please update the following javadoc in accordance with the new default value: https://github.com/gridgain/gridgain/blob/8.9-master/modules/core/src/main/java/org/apache/ignite/IgniteLogger.java#L56

Also, we need to update startup scripts to reflect the required changes. https://github.com/gridgain/gridgain/blob/8.9-master/bin/include/parseargs.sh#L46

sergeyuttsel:
Fixed.

sk0x50 (approved):

PR #2918, opened 2023-10-03
GG-37431 [IGNITE-18719] Change SqlFieldsQuery#lazy to true by default.

zstan:


modules/clients/src/test/java/org/apache/ignite/internal/jdbc2/JdbcLocalCachesSelfTest.java
```diff
@@ -35,6 +36,7 @@
 /**
  * Test JDBC with several local caches.
  */
+@Ignore("https://ggsystems.atlassian.net/browse/IGN-22834")
```
ygerzhedovich, line 39: Please use IGNiTE link

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
```diff
@@ -360,18 +363,19 @@ public boolean isReplicatedOnly() {
...
-     * OutOfMemoryError. Use this flag as a hint for Ignite to fetch result set lazily, thus minimizing memory
+     * {@link OutOfMemoryError}. Use this flag as a hint for Ignite to fetch result set lazily, thus minimizing memory
      * consumption at the cost of moderate performance hit.
      * <p>
-     * Defaults to {@code false}, meaning that the whole result set is fetched to memory eagerly.
+     * Defaults to {@code true}, meaning that the only first page of result set is fetched to memory.
      *
      * @param lazy Lazy query execution flag.
      * @return {@code this} For chaining.
+     * @deprecated Use {@link #setPageSize(int)} instead.
```
ygerzhedovich, line 378: why you deprecate these methods? How I can use `setPageSize` instead of these methods?

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
```diff
@@ -81,7 +81,10 @@ public class SqlFieldsQuery extends Query<List<?>> {
     /** */
     private boolean replicatedOnly;
 
-    /** Lazy mode is default since Ignite v.2.8. */
+    /**
+     * Lazy mode is default since GG 8.9.
+     * @deprecated Use {@link #setPageSize(int)} instead.
```
AMashenkov, line 86: How does `pageSize` helps to achieve lazy behavior?

AMashenkov (approved):

PR #2921, opened 2023-10-05
GG-37519 Execution DML through multi statement throws exception

zstan:


modules/core/src/main/java/org/apache/ignite/internal/processors/query/GridMultiStatementQueryCancel.java
```diff
@@ -0,0 +1,21 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ygerzhedovich, line 2: 2023 :)

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1285,7 +1298,7 @@ else if (parseRes.isDml()) {
             GridNearTxLocal tx = ctx.cache().context().tm().tx();
 
             if (tx != null && tx.mvccSnapshot() != null &&
-                (!(e instanceof IgniteSQLException) || /* Parsing errors should not rollback Tx. */
+                (!(e instanceof IgniteSQLException) || /* Parsing errors should not roll back Tx. */
```
ygerzhedovich, line 1301: looks as mistake - let's rollback the change )

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/IgniteH2Indexing.java
```diff
@@ -1294,6 +1307,18 @@ else if (parseRes.isDml()) {
         }
     }
 
+    /** Initiate new cancel holder, chaining cancelation activity with previous state. */
+    private GridQueryCancel wrapCancel(GridQueryCancel inst) {
+        GridQueryCancel cancel0 = new GridMultiStatementQueryCancel();
+        try {
+            inst.add(cancel0::cancel);
```
ygerzhedovich, line 1314: cyclic cancel? it looks incorrect

zstan: yes cyclic, but i can`t see the problem here, can you clarify ? It just fast return.
```
    public synchronized void cancel() {
        if (canceled)
            return;
```

zstan:
@ygerzhedovich new visa obtained, plz take a look

PR #2923, opened 2023-10-09
GG-37407 Fixed flaky test JdbcThinQueryMemoryTrackerSelfTest.testPartialResultRead

xtern:
The test is flaky because the local handler of GridQueryCancelRequest executes asynchronously ([at this line](https://github.com/gridgain/gridgain/blob/master/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridReduceQueryExecutor.java#L1058) `send()` is called with `runLocParallel=true`) and `CloseResultCntr` may update a little later than expected.

### DIFF
--- a/modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinQueryMemoryTrackerSelfTest.java
+++ b/modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinQueryMemoryTrackerSelfTest.java
@@ -34,6 +34,7 @@ import org.junit.Test;
 import sun.reflect.generics.reflectiveObjects.NotImplementedException;
 
 import static org.apache.ignite.internal.util.IgniteUtils.MB;
+import static org.apache.ignite.testframework.GridTestUtils.waitForCondition;
 
 /**
  * Query memory manager for local queries.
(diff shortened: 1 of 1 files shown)


AMashenkov (approved):

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinQueryMemoryTrackerSelfTest.java
```diff
@@ -77,7 +78,9 @@ public void testPartialResultRead() throws Throwable {
             }
         }
 
-        assertEquals(localResults.size(), closedResultCntr.get());
+        // A wait is required because the query cancel request handler runs asynchronously.
+        boolean success = waitForCondition(() -> localResults.size() == closedResultCntr.get(), getTestTimeout());
```
ygerzhedovich, line 82: I would prefer small fixed timeout for such case, we should't wait 5 mins in case any problem, Le'ts use for example 10 secs

ygerzhedovich (approved):

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinQueryMemoryTrackerSelfTest.java
xtern: Done

PR #2929, opened 2023-10-11
GG-37437 Changed "allowOverwrite for Streamer" from false to true.

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-37437


modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerImpl.java
```diff
@@ -316,6 +318,13 @@ public DataStreamerImpl(
         this.cacheName = cacheName;
         this.flushQ = flushQ;
 
+        if (getBoolean(IGNITE_DATA_STREAMER_ALLOW_OVERWRITE, true)) {
+            rcvr = DataStreamerCacheUpdaters.individual();
+        }
+        else {
+            rcvr = ISOLATED_UPDATER;
+        }
```
vldpyatkov, line 326: Here, you can rewrite the part like this:
`
        if (getBoolean(IGNITE_DATA_STREAMER_ALLOW_OVERWRITE)) {
            rcvr = ISOLATED_UPDATER;
        }
        else {
            rcvr = DataStreamerCacheUpdaters.individual();
        }
`
without duplication default value.
But it is your choice.

vldpyatkov (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/datastreamer/DataStreamerImpl.java
sergeyuttsel: This approach doesn't work because default value from SystemProperty annotation is not used as default in `getBoolean` method.

PR #2943, opened 2023-10-27
GG-37242 Fix for WAL compressor logic: it stops on the previous successful CP instead of the end of WAL archive

sergey-chugunov-1985:


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/WalCompactionTest.java
```diff
@@ -134,6 +158,74 @@ private RolloverRecord() {
...
+
+        assertTrue(GridTestUtils.waitForCondition(
+                () -> wal(ig).lastArchivedSegment() > lastCpIdx,
+                getTestTimeout()));
+
+        Collection<?> queue = GridTestUtils.getFieldValue(wal(ig), "segmentAware", "segmentCompressStorage", "segmentsToCompress");
+
+        assertTrue(GridTestUtils.waitForCondition(queue::isEmpty, getTestTimeout()));
+
+        fileIoFactory.startFailing();
```
ibessonov, line 199: Would it be easier to stop the node and delete end marker? Is there a utility code for that?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/reader/IgniteWalReaderTest.java
```diff
@@ -1915,11 +1915,20 @@ else if (entry instanceof MarshalledDataEntry) {
      * @param ignite ignite instance.
      * @param recordsToWrite count.
      */
-    private void putDummyRecords(Ignite ignite, int recordsToWrite) {
+    private void putDummyRecords(Ignite ignite, int recordsToWrite, boolean forceCheckpoint) throws IgniteCheckedException {
         IgniteCache<Object, Object> cache0 = ignite.cache(CACHE_NAME);
 
-        for (int i = 0; i < recordsToWrite; i++)
+        for (int i = 0; i < recordsToWrite / 2; i++)
```
ibessonov, line 1921: Why so strange?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/db/wal/WalCompactionTest.java
sergey-chugunov-1985: I tried this, it turned out that when we allow a CP to finish successfully it may result in compressor making progress and removing WAL segments that it shouldn't remove.
So I decided it is better to implement a 'fair' scenario with CP failing 'for real'.

ibessonov: Ok, didn't think about it

ibessonov (approved):

PR #2946, opened 2023-10-30
gg-36839 Update condition to show freelist logging

josephdanrev:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
```diff
@@ -448,12 +449,14 @@ public long freeSpace() {
             }
         }
 
-        if (dataPages > 0) {
+        reusePages = bucketsSize.get(REUSE_BUCKET);
```
sergey-chugunov-1985, line 452: An idea for minor improvement here: in the for cycle above we already obtained this value but threw it away because it was the value for reuse bucket (see line 426).

How about adding `else` branch there and assign `reusePages` variable there?

sergey-chugunov-1985 (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/freelist/AbstractFreeList.java
josephdanrev: Didn't see that before. That was a great idea. I made the improvement.

sergey-chugunov-1985 (approved):

PR #2959, opened 2023-11-16
GG-37670 Add handshake timeout check to GridNioSslHandler

ptupitsyn:


isapego (approved):

modules/core/src/main/java/org/apache/ignite/internal/util/nio/ssl/GridNioSslHandler.java
```diff
@@ -56,6 +56,9 @@ class GridNioSslHandler extends ReentrantLock {
     /** Handshake duration threshold; warning is logged when exceeded. */
     private static final long LONG_HANDSHAKE_THRESHOLD_MS = 1000;
 
+    /** Handshake timeout; exception is thrown when exceeded. */
+    private static final long HANDSHAKE_TIMEOUT_MS = 3000;
```
solveme, line 60: This looks like a hard limit, what about making it configurable via java properties just in case?

ptupitsyn: Good point. Added `IGNITE_SSL_HANDSHAKE_TIMEOUT` system property.

ptupitsyn: Good point, dome.

PR #2960, opened 2023-11-16
GG-37687 AssertionError when calling cache.get inside transaction which contains CacheEntryProcessor with remove

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-37687


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/CacheEntryProcessorSelfTest.java
```diff
@@ -0,0 +1,68 @@
...
+
+                    return oldValue;
+                });
+
+                assertNull(cache.get(key));
+
+                tx.commit();
+            }
+
+            assertNull(cache.get(key));
```
denis-chudov, line 65: I'd rather also add the check that the value is not null after tx rollback.

denis-chudov (approved):

PR #2961, opened 2023-11-16
GG-37743 Fixed NPE on GridCacheMapEntry#innerRemove CacheConfiguration.setStatisticsEnabled is enabled.

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-37743


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheMapEntry.java
```diff
@@ -1720,13 +1720,17 @@ protected Object keyValue(boolean cpy) {
 
             drReplicate(drType, null, newVer, topVer);
 
-            if (metrics && cctx.statisticsEnabled()) {
+            if (metrics && cctx.statisticsEnabled() && tx != null) {
```
vldpyatkov, line 1723: This code can move down where the checking of the transaction is already, but this way is also correct. Your coice.

vldpyatkov (approved):

PR #2970, opened 2023-11-24
GG-37739 MulticastIpFinder is replaced with VmIpFinder in tests, examples and as a default IpFinder implementation

sergey-chugunov-1985:
https://ggsystems.atlassian.net/browse/GG-37739


tkalkirill (approved):
There are no comments, but maybe the standard address needs to be passed through TDA.

isapego (approved):

PR #2973, opened 2023-11-27
GG-37393 StorageSize metric shows incorrect results

josephdanrev:


ashapkin (changes requested):
A couple of things:
- there must be a test to validate the fix and original issue
- the code doesn't compile
I filed a separate PR https://github.com/gridgain/gridgain/pull/2999/files 
Let's close this PR

ibessonov (approved):

PR #2976, opened 2023-11-29
GG-37923 Add @QuerySqlTable annotation

sdarlington:


modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QuerySqlTable.java
```diff
@@ -0,0 +1,53 @@
...
+    String name() default "";
+
+    /**
+     * Key name. Can be used to denote the key as a whole, for example, if the
+     * key is a scalar.
+     *
+     * @return Key name
+     */
+    String keyFieldName() default "";
+}
```
slukyano, line 53:
```suggestion
}

```

modules/core/src/main/java/org/apache/ignite/cache/QueryEntity.java
```diff
@@ -794,6 +803,16 @@ private static void processAnnotationsInClass(boolean key, Class<?> cls, QueryEn
             throw new CacheException("Recursive reference found in type: " + cls.getName());
 
         if (parent == null) { // Check class annotation at top level only.
+            QuerySqlTable tableCls = cls.getAnnotation(QuerySqlTable.class);
+            if (tableCls != null) {
+                if (key) {
+                    throw new CacheException("@QuerySqlTable annotation can only be used on a value class (" + cls.getName() + ")");
```
slukyano, line 809:
```suggestion
                    throw new CacheException("@QuerySqlTable annotation can only be used on a value class [tableCls=" + cls.getName() + "]");
```

modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QuerySqlTable.java
```diff
@@ -0,0 +1,53 @@
+/*
```
slukyano, line 1: We use a different copyright header in GG.

slukyano (changes requested):
Looks good, left a few minor suggestions, and let's add a few tests.

PR #2982, opened 2023-12-08
GG-37886 Add new event types for service method invocation on a particular service instance.

JKonSir:
-- add new corresponding event types
-- add recording of new events
-- add tests

### DIFF
--- a/modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
+++ b/modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
@@ -99,7 +99,7 @@ public class ServiceEventSubjectIdSelfTest extends GridCommonAbstractTest {
 
                         return true;
                     }
-                },  EventType.EVT_SERVICE_METHOD_EXECUTION_STARTED,
+                }, EventType.EVT_SERVICE_METHOD_EXECUTION_STARTED,
                 EventType.EVT_SERVICE_METHOD_EXECUTION_FINISHED,
                 EventType.EVT_SERVICE_METHOD_EXECUTION_FAILED
             );
--- a/modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
+++ b/modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
@@ -16,13 +16,18 @@
 
 package org.apache.ignite.internal;
 
+import java.util.ArrayList;
+import java.util.Collection;
+import java.util.Iterator;
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
 import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteServices;
 import org.apache.ignite.Ignition;
 import org.apache.ignite.client.ClientServices;
 import org.apache.ignite.client.IgniteClient;
 import org.apache.ignite.configuration.ClientConfiguration;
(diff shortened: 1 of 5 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/service/ClientServiceInvokeRequest.java
```diff
@@ -176,7 +177,23 @@ public ClientServiceInvokeRequest(BinaryReaderExImpl reader) {
 
                 PlatformServices.convertArrayArgs(args, method);
 
-                res = proxy.invokeMethod(method, args, null);
+                UUID subjId;
+
+                if (ctx.kernalContext().security().enabled())
+                    subjId = ctx.kernalContext()
```
sk0x50, line 183: Perhaps, it should be `ctx.securityContext().subject().id()`. WDYT?

modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
```diff
@@ -0,0 +1,301 @@
...
+import java.util.UUID;
+import java.util.concurrent.CountDownLatch;
+
+import static java.util.concurrent.TimeUnit.MILLISECONDS;
+import static org.apache.ignite.events.EventType.EVTS_SERVICE_EXECUTION;
+import static org.apache.ignite.events.EventType.EVT_SERVICE_METHOD_EXECUTION_FAILED;
+import static org.apache.ignite.events.EventType.EVT_SERVICE_METHOD_EXECUTION_FINISHED;
+import static org.apache.ignite.events.EventType.EVT_SERVICE_METHOD_EXECUTION_STARTED;
+
+public class ServiceEventSubjectIdSelfTest extends GridCommonAbstractTest {
```
sk0x50, line 50: I would like to see a test with security enabled.

modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
```diff
@@ -0,0 +1,301 @@
...
+     */
+    @Test
+    public void testSimpleServiceThinClient() throws Exception {
+        latch = new CountDownLatch(2);
+
+        ClientConfiguration cfg = new ClientConfiguration();
+
+        cfg.setAddresses("127.0.0.1:10801");
+
+        thinClient = Ignition.startClient(cfg);
```
sk0x50, line 132: Please use try-with resource statement instead:
```
    try (IgniteClient cli = Ignition.startClient(cfg) {
        ...
    }
```

modules/clients/src/test/java/org/apache/ignite/internal/ServiceEventSubjectIdSelfTest.java
```diff
@@ -0,0 +1,301 @@
...
+        Iterator<ServiceEvent> it = evts.iterator();
+
+        assert it.hasNext();
+
+        ServiceEvent evt = it.next();
+
+        assert evt != null;
+
+        assertEquals(EVT_SERVICE_METHOD_EXECUTION_STARTED, evt.type());
+        assertEquals(nodeId, evt.subjectId());
```
sk0x50, line 158: So, the test expects that the subject ID is the ID of the server node. Is it correct? Should it be the ID of the thin client? Just because the thin client is an initiating node. right?

sk0x50 (approved):

PR #2987, opened 2023-12-11
GG-36585 Test IgnitePdsCheckpointMapSnapshotTest.testRestartWithCheckpointMapSnapshot is flaky on master

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-36585


sergey-chugunov-1985 (approved):
It looks like this fix closes cases when we can get lost updates but shouldn't introduce issues if snapshot writing process falls behind checkpoint writing process.
In that case I expect the system to read the latest successful cp snapshot and gather remaining information from WAL archive.

PR #2989, opened 2023-12-12
GG-38000 Fixed server nodes crashing if the cache directory contains illegal chars.

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-38000

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
@@ -2065,14 +2065,6 @@ public class IgnitionEx {
                         "like TcpDiscoverySpi)");
 
                 for (CacheConfiguration ccfg : userCaches) {
-                    if (!CU.isCacheTemplateName(ccfg.getName())
-                            && CU.containsInvalidFileNameChars(ccfg, cfg.getDataStorageConfiguration()))
-                        throw new IgniteCheckedException(
-                            "Cache start failed. Cache or group name contains the characters " +
-                                "that are not allowed in file names [cache=" + ccfg.getName() +
-                                (ccfg.getGroupName() == null ? "" : ", group=" + ccfg.getGroupName()) + ']'
-                        );
-
                     if (CU.isReservedCacheName(ccfg.getName()))
                         throw new IgniteCheckedException("Cache name cannot be \"" + ccfg.getName() +
                             "\" because it is reserved for internal purposes.");
--- a/modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
@@ -2065,6 +2065,13 @@ public class IgnitionEx {
                         "like TcpDiscoverySpi)");
 
                 for (CacheConfiguration ccfg : userCaches) {
+                    if (CU.containsInvalidFileNameChars(ccfg, cfg.getDataStorageConfiguration()))
+                        throw new IgniteCheckedException(
+                            "Cache start failed. Cache or group name contains the characters " +
+                                "that are not allowed in file names [cache=" +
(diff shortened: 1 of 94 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
```diff
@@ -1084,6 +1085,15 @@ else if (encMgr.masterKeyDigest() != null &&
             }
         }
 
+        if (containsInvalidFileNameChars(ccfg, ctx.config().getDataStorageConfiguration())) {
```
sk0x50, line 1088: I see that this change was originally proposed by Nikita under [IGNITE-20867](https://issues.apache.org/jira/browse/IGNITE-20867), however, it is a bit weird to me that we postponed this validation on this phase, I mean this validation will be done in the discovery thread when a DynamicCacheChangeRequest is received by server node. This can lead to hanged PME in the following case:
 - let's assume we have 2 nodes cluster: the first node is the coordinator on Linux and the second one is the server node on Windows.
  - start cache with the name that contains an invalid symbol on Windows only.
  - the coordinator will successfully process the message and trigger PME
  - the second node will reject the request.
  
Let's try to check this case. 
Perhaps, the validation can be moved to another place, please take a look at IgniteKernal.getOrCreateCache0 -> CU.validateNewCacheName(cacheName);

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheUtils.java
```diff
@@ -1108,6 +1111,14 @@ public static int cacheGroupId(String cacheName, @Nullable String grpName) {
         return grpName != null ? CU.cacheId(grpName) : CU.cacheId(cacheName);
     }
 
+    /**
+     * @param ccfg Cache configuration.
+     * @return Group name if it is specified, otherwise cache name.
+     */
+    public static String cacheOrGroupName(CacheConfiguration<?, ?> ccfg) {
```
sk0x50, line 1118: I don't see any reason for duplicating this method. We already have CacheGroupContext.cacheOrGroupName. We can either extract this from CacheGroupContext and move it to GridCacheUtils, or just use the existing one.

modules/core/src/main/java/org/apache/ignite/internal/IgnitionEx.java
```diff
@@ -2065,6 +2065,14 @@ public void initializeDefaultCacheConfiguration(IgniteConfiguration cfg) throws
                         "like TcpDiscoverySpi)");
 
                 for (CacheConfiguration ccfg : userCaches) {
+                    if (!CU.isCacheTemplateName(ccfg.getName())
```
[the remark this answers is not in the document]
sergeyuttsel: It can be reproduced on window only. So seems it impossible or we need to create c++/.net test.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
sergeyuttsel: I moved it to GridCacheProcessor#dynamicStartCache

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheUtils.java
sergeyuttsel: hanks. I removed it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/ClusterCachesInfo.java
sk0x50: After a private discussion it was decided to move this validation to `CU.validateNewCacheName`

sk0x50 (approved):

PR #2993, opened 2023-12-13
GG-37798 Check that cache exists on reset lost partitions.

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-37798

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
@@ -4038,18 +4038,17 @@ public class GridCacheProcessor extends GridProcessorAdapter {
      * Attempts to register a cache locally if it has not already been registered.
      *
      * @param cacheNames Cache names.
-     * @return Future that will be completed when all caches are checked.
+     * @return Future that will be completed when all caches are registered.
      */
-    public IgniteInternalFuture<?> checkCaches(Collection<String> cacheNames) {
+    public IgniteInternalFuture<?> registerCachesLocally(Collection<String> cacheNames) {
         List<IgniteInternalFuture<?>> futures = new ArrayList<>();
 
         for (String cacheName : cacheNames) {
-            IgniteInternalCache<Object, Object> cache0 = internalCache(cacheName);
+            IgniteInternalCache<Object, Object> cache = internalCache(cacheName);
 
             // Try to start cache, there is no guarantee that cache will be instantiated.
-            if (cache0 == null) {
+            if (cache == null)
                 futures.add(dynamicStartCache(null, cacheName, null, false, false, true));
-            }
         }
 
         return futures.stream().collect(IgniteCollectors.toCompoundFuture());
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
@@ -4035,6 +4035,27 @@ public class GridCacheProcessor extends GridProcessorAdapter {
     }
 
     /**
+     * Attempts to register a cache locally if it has not already been registered.
+     *
+     * @param cacheNames Cache names.
+     * @return Future that will be completed when all caches are checked.
+     */
+    public IgniteInternalFuture<?> checkCaches(Collection<String> cacheNames) {
+        List<IgniteInternalFuture<?>> futures = new ArrayList<>();
+
+        for (String cacheName : cacheNames) {
+            IgniteInternalCache<Object, Object> cache0 = internalCache(cacheName);
+
+            // Try to start cache, there is no guarantee that cache will be instantiated.
+            if (cache0 == null) {
+                futures.add(dynamicStartCache(null, cacheName, null, false, false, true));
+            }
+        }
+
+        return futures.stream().collect(IgniteCollectors.toCompoundFuture());
+    }
+
+    /**
      * Save cache configuration to persistent store if necessary.
      *
      * @param desc Cache descriptor.
--- a/modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
@@ -16,19 +16,21 @@
 
 package org.apache.ignite.cache;
 
-import java.util.Arrays;
-import org.apache.ignite.Ignite;
 import org.apache.ignite.IgniteCache;
 import org.apache.ignite.cluster.ClusterState;
 import org.apache.ignite.configuration.CacheConfiguration;
 import org.apache.ignite.configuration.DataRegionConfiguration;
 import org.apache.ignite.configuration.DataStorageConfiguration;
 import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.internal.IgniteEx;
 import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
 import org.junit.Test;
 
+import java.util.Arrays;
+import java.util.Set;
+
 /**
- *
+ * Tests for the reset lost partition by client node.
  */
 public class ResetLostPartitionsFromClientTest extends GridCommonAbstractTest {
     /** {@inheritDoc} */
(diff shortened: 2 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -4034,6 +4034,27 @@ public IgniteInternalFuture<?> resetCacheState(Collection<String> cacheNames) th
         return initiateCacheChanges(reqs).stream().collect(IgniteCollectors.toCompoundFuture());
     }
 
+    /**
+     * Attempts to register a cache locally if it has not already been registered.
+     *
+     * @param cacheNames Cache names.
+     * @return Future that will be completed when all caches are checked.
+     */
+    public IgniteInternalFuture<?> checkCaches(Collection<String> cacheNames) {
```
sk0x50, line 4043: It seems to me, that the name of the method is meaningless. Check caches for what? Please rename it.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -4034,6 +4034,27 @@ public IgniteInternalFuture<?> resetCacheState(Collection<String> cacheNames) th
...
+        List<IgniteInternalFuture<?>> futures = new ArrayList<>();
+
+        for (String cacheName : cacheNames) {
+            IgniteInternalCache<Object, Object> cache0 = internalCache(cacheName);
+
+            // Try to start cache, there is no guarantee that cache will be instantiated.
+            if (cache0 == null) {
```
sk0x50, line 4050: Please do not use brackets for a single line `if` statement.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+import org.apache.ignite.configuration.DataRegionConfiguration;
+import org.apache.ignite.configuration.DataStorageConfiguration;
+import org.apache.ignite.configuration.IgniteConfiguration;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+/**
+ *
+ */
+public class ResetLostPartitionsFromClientTest extends GridCommonAbstractTest {
```
sk0x50, line 33: It would be nice to provide javadoc for the test class.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+        cleanPersistenceDir();
+    }
+
+    /** {@inheritDoc} */
+    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        IgniteConfiguration cfg = super.getConfiguration(igniteInstanceName);
+
+        cfg.setDataStorageConfiguration(new DataStorageConfiguration()
+                .setDefaultDataRegionConfiguration(new DataRegionConfiguration()
+                        .setPersistenceEnabled(true)));
```
sk0x50, line 56: I think that we can simplify the test and it allows us to use the in-memory cluster. It should be faster.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+    public void testResetLostPartitions() throws Exception {
+        startGrids(2);
+
+        Ignite node = startClientGrid(3);
+
+        node.cluster().state(ClusterState.ACTIVE);
+
+        String cacheName = "test";
+
+        CacheConfiguration<Integer, String> cacheCfg = new CacheConfiguration<>(cacheName);
```
sk0x50, line 76: Let's specify the partition loss policy for this cache as follows: `cacheCfg.setPartitionLossPolicy(PartitionLossPolicy.READ_WRITE_SAFE)`. In this case, we can use the in-memory cluster instead of persistence.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+        cacheCfg.setBackups(0);
+
+        IgniteCache<Integer, String> cache = node.getOrCreateCache(cacheCfg);
```
sk0x50, line 79: Just start the cache from the coordinator node. So, the descriptor will be available on the client, but the cache will not be started on it.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        for (int i = 0; i < 100; i++) {
+            cache.put(i, "val-" + i);
+        }
+
+        node.close();
```
sk0x50, line 85: After previous test changes, stopping and restarting the client is not needed, just stop one server node.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,100 @@
...
+
+        startGrid(1);
+
+        node = startClientGrid(3);
+
+        node.resetLostPartitions(Arrays.asList(cacheName));
+
+        cache = node.cache(cacheName);
+
+        for (int i = 0; i < 100; i++)
```
sk0x50, line 97: It seems to me, this check for lost partitions is not fully correct. Let's use:
```
    Set<Integer> lostParts = node.cachex("test").context().topology().lostPartitions();
    assertTrue(lostParts.isEmpty());
```

sk0x50:
In general, the proposed way to fix the issue is good and correct.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
sergeyuttsel: Tried to fix.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
sergeyuttsel: Fixed

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
sergeyuttsel: Fixed

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
sergeyuttsel: Thanks. Fixed.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
sergeyuttsel: Fixed

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
sergeyuttsel: Fixed

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
sergeyuttsel: Fixed

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
sergeyuttsel: Thanks. Fixed.

modules/core/src/test/java/org/apache/ignite/cache/ResetLostPartitionsFromClientTest.java
```diff
@@ -0,0 +1,99 @@
...
+     *
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testResetLostPartitions() throws Exception {
+        IgniteEx ignite = startGrids(2);
+
+        IgniteEx client = startClientGrid(3);
+
+        client.cluster().state(ClusterState.ACTIVE);
```
sk0x50, line 73: You don't need to activate the in-memory cluster. So, it can be removed.

sergeyuttsel: Yes, thanks.

sk0x50 (approved):

PR #2994, opened 2023-12-14
GG-37994 Task event can record attributes.

antkr:
https://ggsystems.atlassian.net/browse/GG-37994


modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
```diff
@@ -101,6 +130,41 @@ public void testMultiThreaded() throws Exception {
...
+    public void testAttributesToEventsPropagation() throws Exception {
+        final int intParam = 1;
+        final String textParam = "text";
+
+        IgniteEx n = startGrid(0);
+
+        n.cluster().state(ACTIVE);
+
+        IgnitePredicate<TaskEvent> lsnr = evt -> {
+            System.out.println("Received task event [evt=" + evt.name() + ", taskName=" + evt.taskName() + ", taskAttributes=" + evt.attributes() + ']');
```
tkalkirill, line 140: Why sout not log?

modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
```diff
@@ -101,6 +130,41 @@ public void testMultiThreaded() throws Exception {
...
+
+            if (evt.type() != EVT_TASK_STARTED && evt.attributes().isEmpty()) {
```
tkalkirill, line 142: I think here you need to do `assertEquals(EVT_TASK_STARTED, evt.type())` also `assertFalse(evt.attributes().isEmpty())` and not a condition.

I don’t quite understand how it is checked that the listener has worked accurately and correctly, maybe we can wrap the result in the future and wait for it?

modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
```diff
@@ -101,6 +130,41 @@ public void testMultiThreaded() throws Exception {
...
+
+            return true;
+        };
+
+        n.events().localListen(lsnr, EVTS_TASK_EXECUTION);
+
+        // Generate task events.
+        IgniteFuture<Void> taskFut0 = n.compute().runAsync(new CallableWithSessionAttributes(intParam, textParam));
+
+        taskFut0.get();
```
tkalkirill, line 155: Without a timeout, it's not safe.

tkalkirill:
I think that you still need to check that there will be no backward compatibility problems in a heterogeneous cluster.

modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
tkalkirill: I corrected it myself.

modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
tkalkirill: Fixed

modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
tkalkirill: fixed

vldpyatkov (approved):

modules/core/src/test/java/org/apache/ignite/session/GridSessionSetTaskAttributeSelfTest.java
```diff
@@ -114,6 +203,29 @@ private void checkTask(int num) {
         assert (Integer)res == SPLIT_COUNT : "Invalid result [num=" + num + ", fut=" + fut + ']';
     }
 
+    /**
+     *
+     */
+    @ComputeTaskSessionFullSupport
+    private static class CallableWithSessionAttributes implements IgniteRunnable {
```
ashapkin, line 210: Typo: Callable but implements Runnable.

PR #2998, opened 2023-12-18
GG-37896 Module java.base does not "opens java.security.cert" to unnamed module

vmalin:
Environment scripts fixed.


solveme (commented):
LGTM

PR #3007, opened 2023-12-22
GG-38104 Introduce new method IgniteCache.touch(key)

sk0x50:
https://ggsystems.atlassian.net/browse/GG-38104

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryEx.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryEx.java
@@ -1119,9 +1119,19 @@ public interface GridCacheEntryEx {
      * Updates ttl for this entry and stores its value in storage.
      *
      * @param expiryPlc Expiry policy to be uased to calculate new ttl value.
+     * @return Cached object.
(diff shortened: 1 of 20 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheEntryEx.java
```diff
@@ -1110,9 +1110,28 @@ public void updateIndex(
...
+    public CacheObject touchTtl(@Nullable IgniteCacheExpiryPolicy expiryPlc) throws GridCacheEntryRemovedException;
+
+    /**
+     * Updates ttl for this entry and stores its value in storage.
+     *
+     * @param expiryPlc Expiry policy to be uased to calculate new ttl value.
+     * @return Cached object.
+     * @throws GridCacheEntryRemovedException If entry is obsolete (entry was removed).
+     */
+    public EntryGetResult touchTtlVersioned(@Nullable IgniteCacheExpiryPolicy expiryPlc) throws GridCacheEntryRemovedException;
```
denis-chudov, line 1134: It's unclear from the javadoc and from the method names what is the difference between them and what are their use cases.

denis-chudov (approved):

PR #3012, opened 2023-12-26
GG-37993 SQL functions should ignore data sensitive configuration

zstan:


modules/core/src/main/java/org/apache/ignite/internal/binary/BinaryEnumObjectImpl.java
```diff
@@ -271,37 +271,42 @@ private <T> T uncachedValue(Class<?> cls) throws BinaryObjectException {
         return false;
     }
 
+    /** {@inheritDoc} */
+    @Override public String getString() {
+        // 1. Try deserializing the object.
+        try {
+            Object val = deserialize();
+
+            return new SB().a(val).toString();
```
AMashenkov, line 280:
```suggestion
            return Objects.toString(val);
```

AMashenkov (approved):

xtern (approved):

PR #3017, opened 2023-12-27
GG-37609 Updated zookeeper library from 3.4.14 to 3.8.3 to fix CVE-2023-44981.

sergeyuttsel:
https://ggsystems.atlassian.net/browse/GG-37609


modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/ZookeeperDiscoverySpiTestUtil.java
```diff
@@ -18,29 +18,44 @@
...
+    public static final String ZK_ENABLE_ADMIN_SERVER = "admin.enableServer";
+
     /**
      * @param instances Number of instances in cluster.
+     * @param customProps Custom configuration properties for every server.
      * @return Test cluster.
      */
-    public static TestingCluster createTestingCluster(int instances) {
-        return createTestingCluster(instances, 0);
+    public static TestingCluster createTestingCluster(int instances, @Nullable Map<String,Object>[] customProps) {
```
sk0x50, line 45: Let's introduce 
```
public static TestingCluster createTestingCluster(int instances) {
    return createTestingCluster(instances, 0, null);
}
```

It will reduce the number of changes in other places, I think.

sk0x50, line 45: Why is the `customProps` just ignored and not passed to the underlying call?

modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/internal/ZookeeperDiscoverySegmentationAndConnectionRestoreTest.java
```diff
@@ -254,15 +253,12 @@ public void testSegmentation3() throws Exception {
             srvs.get(0).stop();
             srvs.get(1).stop();
 
-            QuorumPeer qp = srvs.get(2).getQuorumPeer();
-
-            // Zookeeper's socket timeout [tickTime * initLimit] + 5 additional seconds for other logic
-            assertTrue(l.await(qp.getTickTime() * qp.getInitLimit() + 5000, TimeUnit.MILLISECONDS));
+            assertTrue(l.await(30, TimeUnit.MILLISECONDS));
```
sk0x50, line 256: Could you please clarify why this code was changed?

sergeyuttsel: Of course 30 milliseconds is not enough. Increased to 60 seconds. I think it is not necessary to calculate the timeout in any special way like `qp.getTickTime() * qp.getInitLimit() + 5000`.

modules/zookeeper/src/test/java/org/apache/ignite/spi/discovery/zk/ZookeeperDiscoverySpiTestUtil.java
sergeyuttsel: fixed

sergeyuttsel: I checked it again. I think we don't need the customProps.

sk0x50 (approved):

PR #3023, opened 2023-12-29
GG-37820 Implement an option to use "syncfs" during checkpoints.

ibessonov:
https://ggsystems.atlassian.net/browse/GG-37820

### DIFF
--- a/modules/core/pom.xml
+++ b/modules/core/pom.xml
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!--
- Copyright 2019 GridGain Systems, Inc. and Contributors.
+ Copyright 2024 GridGain Systems, Inc. and Contributors.
 
  Licensed under the GridGain Community Edition License (the "License");
  you may not use this file except in compliance with the License.
@@ -71,6 +71,12 @@
             <groupId>org.jetbrains</groupId>
             <artifactId>annotations</artifactId>
             <version>${jetbrains.annotations.version}</version>
+        </dependency>
+
+        <dependency>
+            <groupId>net.java.dev.jna</groupId>
+            <artifactId>jna</artifactId>
+            <version>${jna.version}</version>
         </dependency>
 
         <dependency>
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
@@ -200,12 +200,7 @@ public class CheckpointWorkflow {
         this.igniteInstanceName = igniteInstanceName;
         this.checkpointCollectPagesInfoPool = initializeCheckpointPool();
 
-        if (SyncFsUtils.SYNC_FS_ENABLED) {
-            if (SyncFsUtils.isEnabled())
-                log.info("syncfs is enabled with a threshold of " + SyncFsUtils.SYNC_FS_THRESHOLD + " files.");
-            else
-                log.warning("syncfs checkpoint optimization is enabled, but not supported by the OS.");
-        }
+        SyncFsUtils.logStartup(log);
     }
 
     /**
@@ -632,13 +627,13 @@ public class CheckpointWorkflow {
      * Tries to call {@code syncfs} on the stores if it's enabled and there are enough stores to exceed the threshold.
      *
      * @param updStores Map with updated page stores.
-     * @return {@code true} is {@code syncfs} as been called, {@code false} otherwise.
+     * @return {@code true} if {@code syncfs} has been called, {@code false} otherwise.
      * @throws IgniteCheckedException If failed.
      */
     public boolean trySyncFs(
         ConcurrentMap<PageStore, CheckpointPageStoreInfo> updStores
     ) throws IgniteCheckedException {
-        if (SyncFsUtils.isEnabled()) {
+        if (SyncFsUtils.isActive()) {
             if (updStores.size() > SyncFsUtils.SYNC_FS_THRESHOLD) {
                 SyncFsUtils.syncfs(updStores);
 
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
@@ -25,6 +25,8 @@ import java.util.List;
 import java.util.Map;
 import java.util.Set;
 import java.util.UUID;
+import java.util.concurrent.ConcurrentHashMap;
+import java.util.concurrent.ConcurrentMap;
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.ForkJoinPool;
 import java.util.concurrent.ForkJoinTask;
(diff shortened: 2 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsUtils.java
```diff
@@ -0,0 +1,146 @@
...
+
+    /**
+     * Value of the {@link #IGNITE_CHECKPOINT_SYNC_FS_ENABLED} system property. {@code true} by default.
+     */
+    public static final boolean SYNC_FS_ENABLED = getBoolean(IGNITE_CHECKPOINT_SYNC_FS_ENABLED, true);
+
+    /**
+     * Value of the {@link #IGNITE_CHECKPOINT_SYNC_FS_THRESHOLD} system property. {@code 1024} by default.
+     */
+    public static final int SYNC_FS_THRESHOLD = getInteger(IGNITE_CHECKPOINT_SYNC_FS_THRESHOLD, 1024);
```
rpuch, line 63: How is this 1024 obtained?

rpuch: Also, is this 1024 files per checkpoint, or 1024 files per filesystem per checkpoint? Could it happen that just 2 files need to be fsynced on one FS and 10k files on another one? If yes, will this cause FS sync on both filesystems?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsUtils.java
```diff
@@ -0,0 +1,146 @@
...
+            }
+        }
+
+        // Maps file store name to a path from that file store. Second pass is to fill this map.
+        Map<String, Path> fileSystems = new HashMap<>();
+
+        groupIdToFileStore.forEach((key, val) -> fileSystems.putIfAbsent(val.getValue().name(), val.getKey()));
+
+        // Resulting collection should not be big. For most installations it will only consist of a single element.
+        for (Path path : fileSystems.values()) {
```
rpuch, line 108: I suggest to separate this method in two. First half (extracted to a different method) would form the collection that you obtain via `fileSystems.values()` from `updStores` (for example, `fileSystemsToSync()`); second half (again, a new method) would sync the filesystems in that collection (for example, `syncFileSystems()`). Current method is very long..

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsUtils.java
```diff
@@ -0,0 +1,146 @@
...
+                    List<String> messages = new ArrayList<>(2);
+
+                    if (syncFsErr != 0)
+                        messages.add(SyncFsLibC.strerror(syncFsErr));
+
+                    if (closeErr != 0)
+                        messages.add(SyncFsLibC.strerror(closeErr));
+
+                    //noinspection ThrowFromFinallyBlock
+                    throw new IgniteCheckedException("Failed to execute syncfs: " + messages);
```
rpuch, line 133: Do we really need to throw from a `finally` block? Aren't we going to lose some throwable thrown from the `try` block?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
```diff
@@ -196,6 +199,13 @@ public class CheckpointWorkflow {
         this.checkpointWriteOrder = checkpointWriteOrder;
         this.igniteInstanceName = igniteInstanceName;
         this.checkpointCollectPagesInfoPool = initializeCheckpointPool();
+
+        if (SyncFsUtils.SYNC_FS_ENABLED) {
+            if (SyncFsUtils.isEnabled())
```
rpuch, line 204: This looks weird. 'If it's enabled and it's enabled'. Also, `isEnabled()` is not about just being enabled, but about being enabled by the user and being available in the OS.

Can `isEnabled()` be renamed to `isActive()` or something else.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
```diff
@@ -618,6 +628,30 @@ public List<CheckpointListener> getRelevantCheckpointListeners(Collection<DataRe
             .collect(Collectors.toList());
     }
 
+    /**
+     * Tries to call {@code syncfs} on the stores if it's enabled and there are enough stores to exceed the threshold.
+     *
+     * @param updStores Map with updated page stores.
+     * @return {@code true} is {@code syncfs} as been called, {@code false} otherwise.
```
rpuch, line 635:
```suggestion
     * @return {@code true} if {@code syncfs} has been called, {@code false} otherwise.
```

modules/core/pom.xml
```diff
@@ -1,7 +1,7 @@
 <?xml version="1.0" encoding="UTF-8"?>
 
 <!--
- Copyright 2019 GridGain Systems, Inc. and Contributors.
+ Copyright 2024 GridGain Systems, Inc. and Contributors.
```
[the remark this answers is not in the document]
ibessonov: No way :)

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsLibC.java
```diff
@@ -0,0 +1,92 @@
...
+        boolean jnaAvailable0 = false;
+
+        if (Platform.isLinux()) {
+            try {
+                if (checkLinuxVersion()) {
+                    Native.register(Platform.C_LIBRARY_NAME);
+                    jnaAvailable0 = true;
+                }
+            }
+            catch (Throwable ignore) {
```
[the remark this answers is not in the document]
ibessonov: I've no logger here, but yeah, probably. I'll log it in checkpoint workflow constructor

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsUtils.java
ibessonov: 1024 per checkpoint, determining how many FSs there are requires some extra code as well. This number is almost random, I must say, because there's no way I can predict it precisely.
I see what you mean, but I'm not sure that such cases worth the effort

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsUtils.java
ibessonov: Ok

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/syncfs/SyncFsUtils.java
ibessonov: I'll think what I can do. If seems like I'll write my own try-with-resources

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointWorkflow.java
ibessonov: Good idea

rpuch (approved):

PR #3024, opened 2023-12-29
GG-38207 Lazy PartitionLogTree initialization

AMashenkov:


modules/core/src/main/java/org/apache/ignite/internal/Factory.java
```diff
@@ -0,0 +1,30 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ygerzhedovich, line 2: 2024 :smile:

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheOffheapManager.java
```diff
@@ -2522,7 +2587,7 @@ private Metas getOrAllocatePartitionMetas() throws IgniteCheckedException {
                             partMetastoreReuseListAllocated = true;
                         }
 
-                        if ((updateLogTreeRoot = io.getUpdateTreeRoot(pageAddr)) == 0) {
+                        if ((updateLogTreeRoot = io.getUpdateTreeRoot(pageAddr)) == 0 && preallocateUpdateLogTree) {
```
ygerzhedovich, line 2590: why we do it again? At line 2541 we already allocated page for case preallocateUpdateLogTree

ygerzhedovich: sorry, my bad.

PR #3033, opened 2024-01-09
GG-38251 Optimise fsync calls during WAL rollovers

ibessonov:
DON'T MERGE!
Experimental branch for benchmarking.
https://ggsystems.atlassian.net/browse/GG-38251

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
@@ -445,7 +445,7 @@ public class FileWriteAheadLogManager extends GridCacheSharedManagerAdapter impl
 
         switchSegmentRecordOffset = isArchiverEnabled() ? new AtomicLongArray(dsCfg.getWalSegments()) : null;
 
-        useAsyncRollover = mode != WALMode.FSYNC && mmap && IgniteSystemProperties.getBoolean(USE_ASYNC_ROLLOVER);;
+        useAsyncRollover = mode != WALMode.FSYNC && mmap && IgniteSystemProperties.getBoolean(USE_ASYNC_ROLLOVER);
 
         if (useAsyncRollover) {
             segmentStatus = new AtomicIntegerArray(dsCfg.getWalSegments());
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
@@ -164,7 +164,6 @@ import static org.apache.ignite.internal.processors.compress.CompressionProcesso
  */
 @SuppressWarnings("IfMayBeConditional")
 public class FileWriteAheadLogManager extends GridCacheSharedManagerAdapter implements IgniteWriteAheadLogManager {
-
     /** */
     private static final FileDescriptor[] EMPTY_DESCRIPTORS = new FileDescriptor[0];
 
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
@@ -393,7 +393,7 @@ public class FileWriteAheadLogManager extends GridCacheSharedManagerAdapter impl
     /** Pointer to the last successful checkpoint until which WAL segments can be safely deleted. */
     private volatile FileWALPointer lastCheckpointPtr = new FileWALPointer(0, 0, 0);
 
-    private final boolean useAsyncRollover = true;
+    private final boolean useAsyncRollover;
 
     private final AtomicIntegerArray segmentStatus;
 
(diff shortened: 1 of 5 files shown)


ibessonov:
First of all, I would like to raise a question here, because I don't remember how it works. Imagine the following scenario:
- you're writing a segment with MMAP;
- concurrently somebody opens standalone WAL iterator and starts scanning the WAL;
- eventually they would need to scan currently opened segment. But its content may not be persisted yet I guess? Does it work? If it does, than how?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -1491,12 +1536,19 @@ private FileWriteHandle initNextWriteHandle(FileWriteHandle cur) throws IgniteCh
 
             boolean interrupted = false;
 
+            long nextSegmentId = cur.getSegmentId() + 1;
+
+            int segmentStatus;
+            if (useAsyncRollover && (segmentStatus = getSegmentStatus(nextSegmentId)) != SEGMENT_CLOSED)
```
ibessonov, line 1542: Are you sure that we shouldn't wait here, and fail immediately instead?
By the way, I think braces are missing for this `if`

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -1741,6 +1799,50 @@ public long maxWalSegmentSize() {
...
+        boolean statusChanged = segmentStatus.compareAndSet(idx, expected, updated);
+
+        if (statusChanged)
+            log.info("Status for segment " + segmentId + " has been changed from " + segmentStatusText(expected)
+                    + " to " + segmentStatusText(updated));
+
+        return statusChanged;
+    }
+
+    private String segmentStatusText(int status) {
```
ibessonov, line 1820: Why not just use an `enum`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -3621,4 +3723,73 @@ static long minWalArchiveSize(DataStorageConfiguration dsCfg) {
     @Override public void startAutoReleaseSegments() {
         segmentAware.startAutoReleaseSegments();
     }
+
+    private class WalSegmentAsyncCloser extends GridWorker {
+        final ConcurrentLinkedQueue<IgniteBiTuple<FileWriteHandle, Boolean>> tasks = new ConcurrentLinkedQueue<>();
```
ibessonov, line 3728: Why don't you use a blocking queue? You wouldn't have to park the thread manually. Just a question, I'm not suggesting anything

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/filehandle/FileWriteHandleImpl.java
```diff
@@ -476,8 +476,6 @@ private static void fsync(MappedByteBuffer buf, int off, int len) throws IgniteC
                     if (segRecPtr != null) {
                         FileWALPointer filePtr = (FileWALPointer)segRecPtr;
 
-                        fsync(filePtr);
```
ibessonov, line 479: I think you can return these bad boys back

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
Erixonich: Initially it was enum, but code was too clunky with that.
I wanted to leave the same structure : AtomicIntegerArray. In this case, I have to convert int from enum and vice versa.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
Erixonich: If you reffering to the `ArrayBlockingQueue` then I believe it synchronized on same mutex on adding and extracting element. LinkedQueue on contrary doesn't use mutexes.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/filehandle/FileWriteHandleImpl.java
Erixonich: Yep, forgot about this.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
Erixonich: Unfortunately not. This method called during node start and node activation. In case of cluster deactivation segment status may be "SEGMENT_ACTIVE".

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -217,6 +221,10 @@ public class FileWriteAheadLogManager extends GridCacheSharedManagerAdapter impl
     /** @see IgniteSystemProperties#IGNITE_THRESHOLD_WAIT_TIME_NEXT_WAL_SEGMENT */
     public static final long DFLT_THRESHOLD_WAIT_TIME_NEXT_WAL_SEGMENT = 1000L;
 
+    @SystemProperty(value = "Enables asynchronous WAL segment rollover for non-mmap non-fsync WAL mode",
```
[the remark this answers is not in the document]
Erixonich: Thanks, I forgot how it worked.

PR #3044, opened 2024-01-12
GG-38289 Revert GG-37436 Default value of the statisticsEnabled property of the cache configuration was changed to true.

sk0x50:
https://ggsystems.atlassian.net/browse/GG-38289

This reverts commit 00d71d9b0cff55decb149ea7acf4bc709de096aa.


modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
```diff
@@ -531,6 +519,11 @@ public CacheConfiguration(CompleteConfiguration<K, V> cfg) {
         writeBehindFlushSize = cc.getWriteBehindFlushSize();
         writeBehindFlushThreadCnt = cc.getWriteBehindFlushThreadCount();
         writeSync = cc.getWriteSynchronizationMode();
+        storeConcurrentLoadAllThreshold = cc.getStoreConcurrentLoadAllThreshold();
+        maxQryIterCnt = cc.getMaxQueryIteratorsCount();
+        sqlOnheapCache = cc.isSqlOnheapCacheEnabled();
+        sqlOnheapCacheMaxSize = cc.getSqlOnheapCacheMaxSize();
+        evtsDisabled = cc.isEventsDisabled();
```
sergeyuttsel, line 526: In the reverted commit, we made this list in alphabetical order. We could save this change.

sergeyuttsel:
When we set the default value of the StatisticsEnabled to true then we discovered the bug that was fixed by https://github.com/gridgain/gridgain/commit/9cc3428ca788bbaf83d6fbe61c421d62bf64e776
This bug is reproduced in CacheOptimisticTransactionsWithFilterTest if the StatisticsEnabled is true. May be we need to set StatisticsEnabled=true explicitly in this test class to test this change.

modules/core/src/main/java/org/apache/ignite/configuration/CacheConfiguration.java
sk0x50: Yep, makes sense to me. Will do.

sk0x50:
I think creating a new ticket for this case would be the right way to handle it. Will create.

sergeyuttsel (approved):

PR #3070, opened 2024-02-01
GG-36475 Fix the illegal reflective access warnings issue

dy2285:
By adding some JVM parameters in bin/control.sh, the issue of "illegal reflective access warnings" was addressed.


iakkuratov:
Take a look at the file /bin/include/jvmdefaults.sh. I think most of the options have to be moved there

dy2285:
@iakkuratov added the jvm options to jvmdefaults.sh and corresponding changes in control.sh.

dy2285:
@iakkuratov please review the changes.

dy2285:
Undid the commits that added a wrong file name. Cleaned up the empty lines.

bin/controlsh
iakkuratov, line 1: This file should not exist.

bin/include/jvmdefaults.sh
```diff
@@ -82,10 +95,17 @@ getJavaSpecificOpts() {
...
           --add-opens=java.base/java.time=ALL-UNNAMED \
+          --add-opens=java.base/sun.security.ssl=ALL-UNNAMED \
           --add-opens=java.base/sun.security.x509=ALL-UNNAMED \
           --add-opens=java.sql/java.sql=ALL-UNNAMED \
           ${current_value}"
   fi
 
   echo $value
 }
+# Reflective access options
```
iakkuratov, line 106: Please look closer at the comment style for other methods.

bin/control.sh
```diff
@@ -138,6 +138,20 @@ fi
...
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.base/sun.nio.ch=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.management/com.sun.jmx.mbeanserver=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=jdk.internal.jvmstat/sun.jvmstat.monitor=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.base/sun.reflect.generics.reflectiveObjects=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=jdk.management/com.sun.management.internal=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.base/java.io=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.base/java.nio=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.base/java.util=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --add-opens=java.base/java.lang=ALL-UNNAMED"
+CONTROL_JVM_OPTS="${CONTROL_JVM_OPTS} --illegal-access=warn"
```
iakkuratov, line 154: If I get it right this changes have to be reverted since you moved options to a separate file

iakkuratov (changes requested):

bin/controlsh
dy2285: removed.

iakkuratov (approved):
Looks good to me. Please double-check spaces before strings and feel free to proceed with the core review.

bin/control.sh
```diff
@@ -40,12 +40,19 @@ fi
 # Set SCRIPTS_HOME - base path to scripts.
 #
 SCRIPTS_HOME="${IGNITE_HOME_TMP}/bin"
+. "${SCRIPTS_HOME}"/include/functions.sh
```
sk0x50, line 43: I am a bit confused about this change. IMHO, we need to add additional JVM options to the `CONTROL_JVM_OPTS` variable. Something like as follows:
```
CONTROL_JVM_OPTS=$(getJavaSpecificOptsForControlScript $version "$CONTROL_JVM_OPTS")
```
where `getJavaSpecificOptsForControlScript` should determine the required options based on the JVM version like the `getJavaSpecificOpts` does.

By the way, this fix should be applied to the `control.bat` script as well. Perhaps, it makes sense to update `snapshot-utility` scripts in the same way. WDYT?

bin/control.sh
```diff
@@ -112,7 +119,7 @@ fi
 # Assertions are disabled by default since version 3.5.
 # If you want to enable them - set 'ENABLE_ASSERTIONS' flag to '1'.
 #
-ENABLE_ASSERTIONS="1"
+ENABLE_ASSERTIONS="0"
```
sk0x50, line 122: :+1:

bin/control.sh
```diff
@@ -147,18 +160,18 @@ fi
...
-         -DIGNITE_UPDATE_NOTIFIER=false -DIGNITE_HOME="${IGNITE_HOME}" \
+          -DIGNITE_HOME="${IGNITE_HOME}" \
          -DIGNITE_PROG_NAME="$0" ${JVM_XOPTS:-} -cp "${CP}" ${MAIN_CLASS} "$@"
     ;;
     OS/390*)
         "$JAVA" ${CONTROL_JVM_OPTS} ${QUIET:-} \
-         -DIGNITE_UPDATE_NOTIFIER=false -DIGNITE_HOME="${IGNITE_HOME}" \
-         $(getIbmSslOpts $version) \
+          -DIGNITE_HOME="${IGNITE_HOME}" \
+         -Dcom.ibm.jsse2.overrideDefaultTLS=true -Dssl.KeyManagerFactory.algorithm=IbmX509 \
```
sk0x50, line 169: Could you please clarify why this change is needed? I would say that using `$(getIbmSslOpts $version)` is the right way.

bin/include/jvmdefaults.sh
```diff
@@ -103,3 +103,9 @@ getJavaSpecificOpts() {
...
+addReflectiveAccessOptions() {
```
sk0x50, line 107: I think the exact list of options should be based on the JVM version. For example, using these opts with JVM8 leads to 
```
Unrecognized option: --add-exports=java.base/jdk.internal.misc=ALL-UNNAMED
Error: Could not create the Java Virtual Machine.
Error: A fatal exception has occurred. Program will exit.
```

Perhaps, I am missing something.

sk0x50 (changes requested):

bin/include/jvmdefaults.sh
dy2285: Thank you for highlighting the compatibility issues with Java 8. I realize now that my updates to the jvmdefaults.sh script included Java 9+ specific options without checks for Java version compatibility. I will revise the script to dynamically adjust the JVM options based on the version it detects. 

Yes as recommended I will also conduct testing across these Java versions to confirm that the script functions correctly in each environment. 

Once you've confirmed my approach regarding the initial comment of the PR, I will make all necessary changes and conduct all tests at once.

bin/control.sh
sk0x50: If I got your point right, you propose updating `getJavaSpecificOpts` instead of introducing a specific method for `control.sh`. Well, the idea looks good at first glance.

dy2285:
Made changes based on above discussion. Included `control.bat` and 'jvmdefaults.bat' . Will make a separate pr for snapshot-utility scripts. Please review when you have time. Thank you.

PR #3098, opened 2024-02-29
GG-38536 Remove experimental flag from --meta command

odiszapc:


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/meta/MetadataCommand.java
```diff
@@ -123,7 +118,7 @@ public class MetadataCommand extends AbstractCommand<Object> {
 
     /** {@inheritDoc} */
     @Override public boolean experimental() {
```
vldpyatkov, line 120: The overrided method can be removed because it returns false by default.

vldpyatkov (approved):

PR #3104, opened 2024-03-06
GG-37810 added HEADER keyword for BulkLoadCsvFormat

myskov:
I added a new keyword to support CSV headers


vadimkolodingg (approved):
lgtm

PR #3107, opened 2024-03-11
GG-38457 Thin client: Add ClientCacheConfiguration.pluginConfigurations

ptupitsyn:
**Public API**
* Add `ClientCacheConfiguration.pluginConfigurations`, which allows plugin authors to attach additional properties to every cache configuration when starting caches from the thin client.
* Add `PluginProvider.readClientCachePluginConfiguration` to deserialize client-provided plugin config on the server.

**Protocol**
* Add `ProtocolBitmaskFeature.CACHE_PLUGIN_CONFIGURATIONS` feature.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/client/ClientCachePluginConfiguration.java
+++ b/modules/core/src/main/java/org/apache/ignite/client/ClientCachePluginConfiguration.java
@@ -14,17 +14,21 @@
  * limitations under the License.
  */
 
-package org.apache.ignite.plugin;
+package org.apache.ignite.client;
 
-import java.io.Serializable;
-
-import org.apache.ignite.binary.BinaryRawReader;
 import org.apache.ignite.binary.BinaryRawWriter;
-import org.apache.ignite.configuration.CacheConfiguration;
+import org.apache.ignite.plugin.CachePluginConfiguration;
 
 /**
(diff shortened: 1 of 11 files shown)


modules/core/src/main/java/org/apache/ignite/client/ClientCachePluginConfiguration.java
```diff
@@ -0,0 +1,32 @@
...
+/**
+ * Client cache plugin configuration. Maps to {@link CachePluginConfiguration} on the server side.
+ */
+public interface ClientCachePluginConfiguration {
+    /**
+     * Serializes the configuration on the client side.
+     *
+     * @param writer Writer.
+     */
+    void write(BinaryRawWriter writer);
```
isapego, line 31: Should we somehow make it clear that the first written field should always be the plugin name?

isapego (approved):

modules/core/src/main/java/org/apache/ignite/client/ClientCachePluginConfiguration.java
ptupitsyn: Good catch. This is not the right way. I've added `pluginName()` to the interface, so that the plugin does not have to deal with this, and ser/de logic is symmetrical.

isapego (approved):

PR #3112, opened 2024-03-13
GG-38603 Backport SQL System Views from Ignite

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/AtomicReferenceViewWalker.java
```diff
@@ -1,5 +1,5 @@
 /*
- * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ygerzhedovich, line 2: what reason? here and below

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/SqlQueryHistoryViewWalker.java
```diff
@@ -37,15 +37,15 @@ public class SqlQueryHistoryViewWalker implements SystemViewRowAttributeWalker<S
         v.accept(5, "durationMin", long.class);
         v.accept(6, "durationMax", long.class);
         v.accept(7, "lastStartTime", Date.class);
-        v.accept(8, "diskAllocationMax", long.class);
```
ygerzhedovich, line 40: why did you reorder them?

ygerzhedovich (commented):
The ticket not only about BINARY_METADATA, but also for TABLE_COLUMNS, VIEW_COLUMNS and VIEWS views.

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/AtomicReferenceViewWalker.java
zstan: it`s auto generated code, ok i revert it.

zstan:
@ygerzhedovich all other views are implemented, check H2IndexesSystemViewTest#testGeometryColumnPrecisionAndScale or use context search for founding particular test

modules/core/src/main/java/org/apache/ignite/internal/managers/systemview/walker/SqlQueryHistoryViewWalker.java
zstan: missed, thanks fixed

PR #3116, opened 2024-03-14
GG-37140 Downgrade to Debug level for job failures because of Inactive cluster 

josephdanrev:


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -4156,7 +4156,7 @@ private void unguard() {
      */
     private void checkClusterState() throws IgniteException {
         if (!ctx.state().publicApiActiveState(true)) {
-            throw new IgniteException("Can not perform the operation because the cluster is inactive. Note, that " +
+            throw new IllegalStateException("Can not perform the operation because the cluster is inactive. Note, that " +
```
rpuch, line 4159: The idea was to fix it not here, but check the cluster state before actually trying to do an operation (in `VisorSnapshots.snapshot()`, just before calling `ignite.plugin()`, in the same way it's made when checking that persistence and snapshots are activated).

`IllegalStateException` is used widely, it might be thrown by Java components, so it will not be very selective.

rpuch (changes requested):

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
josephdanrev: I'll narrow down my logic. Thanks for feedback.

PR #3123, opened 2024-03-20
GG-37066 Remove meta command need to remove marshaller data

zstan:


modules/core/src/main/java/org/apache/ignite/internal/MarshallerMappingFileStore.java
```diff
@@ -102,6 +104,20 @@ final class MarshallerMappingFileStore {
...
+
+        for (byte platformId : allPlatforms) {
+            String fileName = getFileName(platformId, typeId);
+
+            File file = new File(mappingDir, fileName);
+
+            if (log.isDebugEnabled()) {
+                log.debug("Try to remove: " + file.getAbsolutePath());
+            }
+        }
```
AMashenkov, line 118: Do you forget to call `file.delete()` ?

zstan: yep, done.

AMashenkov (approved):

PR #3125, opened 2024-03-29
GG-38421 The client node was hanging for >1 hour due to rejecting the incoming connection by the server because it was in the process of opening the outgoing connection.

Erixonich:


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionClientPool.java
```diff
@@ -338,6 +359,43 @@ public GridCommunicationClient reserveClient(ClusterNode node, int connIdx) thro
         }
     }
 
+    private GridFutureAdapter<GridCommunicationClient> handleConnectExceptionWhileIncomingConnectionOccurred(
+            ConnectionKey connKey,
+            ConnectFuture origFut,
+            Throwable e
+    ) {
+        log.info("Outgoing connection failed while incoming connection attempt was detected. ");
```
ibessonov, line 367:
```suggestion
        log.info("Outgoing connection failed while incoming connection attempt was detected.");
```

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionClientPool.java
```diff
@@ -338,6 +359,43 @@ public GridCommunicationClient reserveClient(ClusterNode node, int connIdx) thro
...
+
+        GridFutureAdapter<GridCommunicationClient> triggerFut = new GridFutureAdapter<>();
+
+        triggerFut.listen(f -> {
```
ibessonov, line 371: I'm not yet convinced that this code is thread-safe, so please add an explanation of why it should work. I see no tests that would prove it, so I have to trust your words.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionClientPool.java
```diff
@@ -271,13 +277,28 @@ public GridCommunicationClient reserveClient(ClusterNode node, int connIdx) thro
...
+                        if (e instanceof Error) {
+                            fut.onDone(e);
                             throw (Error)e;
+                        }
+
+                        ConnectFuture conFut = (ConnectFuture) (fut);
+
+                        if (conFut.incomingConnectionAttempts() > INCOMING_CONNECTION_ATTEMPTS_THRESHOLD) {
+                            try {
+                                fut = handleConnectExceptionWhileIncomingConnectionOccurred(connKey, conFut, e);
```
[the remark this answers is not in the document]
Erixonich: In `handleConnectExceptionWhileIncomingConnectionOccurred` fut is replaced in the map `clientFuts`. So remove in finally block would delete properly.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionClientPool.java
Erixonich: will fix it.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/ConnectionClientPool.java
Erixonich: AFAIK, listen guarantees one time execution on future completion.
Future will be completed either on line 393 in this class or in InboundConnectionHandler:711.
Map clientFuts is Concurrent

Erixonich:
@ibessonov, I've added necessary commentaries, take look again please.

ibessonov (approved):

PR #3143, opened 2024-04-14
GG-39004 Node will dump persistence files on data corruption error

Erixonich:
https://ggsystems.atlassian.net/browse/GG-39004


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1261,6 +1263,36 @@ public int pageSize() {
...
+
+            U.copy(new File(idxStore.getFileAbsolutePath()), new File(dumpDir, INDEX_FILE_NAME), false);
+
+            Set<Integer> parts = new HashSet<>();
+            for (long pageId : pageIds) {
+                parts.add(PageIdUtils.partId(pageId));
+            }
+
+            for (Integer part : parts) {
+                if (part >= pageStores.partStores.length)//skip index parts
```
ibessonov, line 1284: I don't understand this condition and this comment, can you elaborate?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -3236,6 +3236,57 @@ else if (file.exists()) {
...
+            boolean[] copied = new boolean[(int) (currIdx - lastCpIdx + 1)];
+
+            List<File> walFiles = new ArrayList<>();
+
+            if (archiver != null)
+                walFiles.addAll(Arrays.asList(walArchiveDir.listFiles(WAL_SEGMENT_COMPACTED_OR_RAW_FILE_FILTER)));
+
+            walFiles.addAll(Arrays.asList(walWorkDir.listFiles(WAL_SEGMENT_COMPACTED_OR_RAW_FILE_FILTER)));
+
+            if (archiver != null) {
```
ibessonov, line 3261: What if archiver is null?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -3236,6 +3236,57 @@ else if (file.exists()) {
...
+
+                            U.copy(file, new File(dumpDir, fileName(idx)), false);
+                        }
+                    }
+                }
+            }
+        } catch (Exception e) {
+            log.error("Failed to dump wal files", e);
+        }
+
```
ibessonov, line 3287:
```suggestion
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
Erixonich: Then we don't need to check archived segments, don't we?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
Erixonich: Sure, if error occurred in reading index file, then part = 65535, which outbound of data partitions values.

ibessonov (approved):

PR #3161, opened 2024-05-08
GG-39187 Introduced async version of IgniteCache.touch

sk0x50:
https://ggsystems.atlassian.net/browse/GG-39187

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/IgniteCache.java
+++ b/modules/core/src/main/java/org/apache/ignite/IgniteCache.java
@@ -1719,4 +1719,16 @@ public interface IgniteCache<K, V> extends javax.cache.Cache<K, V>, IgniteAsyncS
      */
     @IgniteExperimental
     public boolean touch(K key);
+
+    /**
+     * Updates the time to live value for the given {@code key}.
+     *
+     * @param key The key whose associated ttl value is to be updated in accrordance with the specified {@link ExpiryPolicy}.
+     * @return A future representing a pending time to live update. The returned future is completed with {@code true}
+     *          if the ttl value was updated, {@code false} if the key is not present in the cache.
+     * @throws NullPointerException If key is {@code null}.
+     * @see IgniteCache#withExpiryPolicy(ExpiryPolicy)
+     */
+    @IgniteExperimental
+    public IgniteFuture<Boolean> touchAsync(K key);
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteInternalCache.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteInternalCache.java
@@ -1864,4 +1864,15 @@ public interface IgniteInternalCache<K, V> extends Iterable<Cache.Entry<K, V>> {
      * @see IgniteCache#withExpiryPolicy(ExpiryPolicy)
      */
     public boolean touch(K key);
+
+    /**
+     * Updates the time to live value for the given {@code key}.
+     *
+     * @param key The key whose associated ttl value is to be updated in accrordance with the specified {@link ExpiryPolicy}.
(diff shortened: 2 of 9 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/expiry/IgniteCacheExpiryPolicyAbstractTest.java
```diff
@@ -1174,9 +1189,18 @@ public void testTouch() throws Exception {
         for (final Integer key : keys) {
             log.info("Update duration on access, key: " + key);
 
-            boolean updated = cache.touch(key);
+            if (async) {
+                IgniteFuture<Boolean> fut = cache.touchAsync(key);
+
+                assertTrue("touchAsync was not completed in 1 sec.", waitForCondition(() -> fut.isDone(), 1_000));
```
denis-chudov, line 1195: Why not `IgniteFuture#get(long, java.util.concurrent.TimeUnit)`?
By the way, isn't 1 sec too little? Any thread or JVM freeze and the test would become flaky. I think it's safe to put bigger timeout here, assuming that in most cases the future will be completed very fast anyway.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteInternalCache.java
```diff
@@ -1864,4 +1864,15 @@ public void localLoadCache(@Nullable IgniteBiPredicate<K, V> p, @Nullable Object
...
     public boolean touch(K key);
+
+    /**
+     * Updates the time to live value for the given {@code key}.
+     *
+     * @param key The key whose associated ttl value is to be updated in accrordance with the specified {@link ExpiryPolicy}.
+     * @return A future representing a pending time to live update. The returned future is completed with {@code true}
+     *          if the ttl value was updated, {@code false} if the key is not present in the cache.
+     * @throws NullPointerException If key is {@code null}.
+     * @see IgniteCache#withExpiryPolicy(ExpiryPolicy)
```
denis-chudov, line 1875: same: I would propose dots here and abobe in `touch()`

modules/core/src/main/java/org/apache/ignite/IgniteCache.java
```diff
@@ -1719,4 +1719,16 @@ public <T> IgniteFuture<Map<K, EntryProcessorResult<T>>> invokeAllAsync(Set<? ex
...
     public boolean touch(K key);
+
+    /**
+     * Updates the time to live value for the given {@code key}.
+     *
+     * @param key The key whose associated ttl value is to be updated in accrordance with the specified {@link ExpiryPolicy}.
+     * @return A future representing a pending time to live update. The returned future is completed with {@code true}
+     *          if the ttl value was updated, {@code false} if the key is not present in the cache.
+     * @throws NullPointerException If key is {@code null}.
+     * @see IgniteCache#withExpiryPolicy(ExpiryPolicy)
```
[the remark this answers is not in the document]
sk0x50: I checked GG and JDK Javadocs and could not find any periods when `@see` is used.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/IgniteInternalCache.java
sk0x50: Please see my comment above.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/expiry/IgniteCacheExpiryPolicyAbstractTest.java
sk0x50: This makes sense to me. will change.

denis-chudov (approved):

PR #3164, opened 2024-05-08
GG-37761 Provide capabilities for JDBC retry on server node restarts

xtern:
Basic issue description:

After node restart, when partition awareness is disabled the first attempt to execute query fails (see example below).
```java
    startGrids(2);

    // Connect with disabled partition awareness.
    try (Connection conn = DriverManager.getConnection(CONNECT_URL)) {
        stopGrid(0);
        startGrid(0);

        conn.createStatement().execute("SELECT 1"); // fails "Failed to communicate with Ignite cluster"
        conn.createStatement().execute("SELECT 1"); // But the second try is successful
    }
```

To fix this issue connection recovery logic was reworked a bit.

* Similar to what is done for partition awareness, the connection is recovered transparently only for idempotent requests.
* If the connection cannot be re-established, the current statement is forced to close and the `SQLException` with sql state '08006' will be raised, so the user may retry the operation at his own risk.
* The connection will not be re-established if `MVCC` is used (in which case the connection is forced to close).


modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinPartitionAwarenessReconnectionAndFailoverSelfTest.java
```diff
@@ -244,10 +299,10 @@ public void testReconnectionDelayIncreasing() throws Exception {
             assertEquals("Unexpected log records count.", 4, logHnd.records.size());
 
             String expRecordMsg = "Failed to connect to Ignite node " +
-                "[url=jdbc:ignite:thin://127.0.0.1:10800,127.0.0.1:10810]. address = [localhost/127.0.0.1:10810].";
+                "[url=jdbc:ignite:thin://127.0.0.1:10800,127.0.0.1:10810]. address = [";
```
xtern, line 302: Depending on OS and java version the local hostname may remain unresolved (in my case it prints `/127.0.0.1:10810` instead of `localhost/127.0.0.1:10810`).

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinPartitionAwarenessReconnectionAndFailoverSelfTest.java
```diff
@@ -244,10 +299,10 @@ public void testReconnectionDelayIncreasing() throws Exception {
             assertEquals("Unexpected log records count.", 4, logHnd.records.size());
 
             String expRecordMsg = "Failed to connect to Ignite node " +
-                "[url=jdbc:ignite:thin://127.0.0.1:10800,127.0.0.1:10810]. address = [localhost/127.0.0.1:10810].";
```
xtern, line 247: Depending on OS and java version the local hostname may remain unresolved (in my case it prints `/127.0.0.1:10810` instead of `localhost/127.0.0.1:10810`).

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinPartitionAwarenessReconnectionAndFailoverSelfTest.java
```diff
@@ -861,17 +867,23 @@ private void assertConnectionsCount(Map<UUID, JdbcThinTcpIo> ios, int expConnCnt
...
      */
-    private void invalidateConnectionToStoppedNode(Connection conn) {
+    private void invalidateConnectionToStoppedNode(Connection conn) throws SQLException {
+        Map<UUID, JdbcThinTcpIo> ios = GridTestUtils.getFieldValue(conn, "ios");
+
+        int initialSize = ios.size();
+
         while (true) {
-            try (Statement stmt = conn.createStatement()) {
-                stmt.execute("select ';';");
```
xtern, line 872: multi-statement query now reconnects silently and does not produces exceptions if there still live nodes in the cluster

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -87,7 +87,6 @@
 import static org.apache.ignite.internal.processors.cache.distributed.dht.topology.GridDhtLocalPartition.DFLT_ATOMIC_CACHE_DELETE_HISTORY_SIZE;
 import static org.apache.ignite.internal.processors.cache.mvcc.MvccCachingManager.DFLT_MVCC_TX_SIZE_CACHING_THRESHOLD;
 import static org.apache.ignite.internal.processors.cache.persistence.GridCacheDatabaseSharedManager.DFLT_DEFRAGMENTATION_REGION_SIZE_PERCENTAGE;
-import static org.apache.ignite.internal.processors.cache.persistence.GridCacheDatabaseSharedManager.DFLT_IGNITE_VALIDATE_CACHE_NAMES;
```
xtern, line 90: not related - compilation problem in master branch

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinConnectionFailoverTest.java
```diff
@@ -0,0 +1,252 @@
...
+                pstmt.setInt(2, 3);
+                pstmt.addBatch();
+
+                // Batch cannot be re-used after connection reset.
+                assertThrowsSql(pstmt::executeBatch, "Failed to communicate with Ignite cluster");
+                assertThrowsSql(() -> pstmt.setInt(1, 5), STATEMENT_CLOSED_ON_DISCONNECT_MESSAGE);
+
+                restartGrid(0);
+
+                PreparedStatement pstmt0 = conn.prepareStatement("INSERT INTO TEST VALUES (?, ?)");
```
ygerzhedovich, line 101: seems lost check  for the prepared statement after restart

xtern: PR reworked to re-establish connection only on idempotent requests. These tests reworked accordingly.

modules/clients/src/test/java/org/apache/ignite/jdbc/thin/JdbcThinConnectionMultipleAddressesTest.java
```diff
@@ -303,6 +309,157 @@ public void testClientConnectionMXBean() throws Exception {
...
+                // Statement is empty when disconnect detected.
+                try (Statement stmt = conn.createStatement()) {
+                    stopGrid(0);
+                    startGrid(0);
+
+                    stmt.execute("SELECT 1");
+                    stmt.execute("SELECT 1");
+                }
+
+                // Statement has closed resultset when disconnect detected.
```
ygerzhedovich, line 344: I don't see such check

ygerzhedovich (approved):

ygerzhedovich (approved):

modules/core/src/main/java/org/apache/ignite/internal/jdbc/thin/JdbcThinConnection.java
```diff
@@ -1785,31 +1818,48 @@ private void connectInCommonMode() throws SQLException {
...
             if (exceptions.size() == 1) {
                 Exception ex = exceptions.get(0);
 
-                if (ex instanceof SQLException)
+                if (ex instanceof SQLException) {
+                    if (disconnected) {
+                        throw new SQLException(
+                            ex.getMessage(), ((SQLException)ex).getSQLState(), new JdbcThinDisconnectException(ex));
+                    }
+
```
AMashenkov, line 1851: Do we need all these exception wrapping? 

Maybe a "if (exceptions.size() == 1) " branch can be dropped? 
Or left it as is and add it something like that
```
if (disconnected) {
   e = new JdbcThinDisconnectException(null);
    for (Exception ex : exceptions)
                e.addSuppressed(ex);
    throw e;
}
```

xtern: > Do we need all these exception wrapping?
:tired_face:  :thinking: 

The task description states that we need to somehow signal the client that a shutdown has occurred, but he can repeat the operation at his own risk.

But after some thinking I decided to drop `JdbcThinDisconnectException`, because it looks like the client can check the SQL error state to understand what happened, something like this.
```
   catch(SQLException e) {
     if (!conn.isClosed() && "08006".equals(ex.getSQLState()))
       // means that disconnection happened but connection can be re-used (operation can be retried
   }
```
states described here: https://www.gridgain.com/docs/latest/developers-guide/SQL/JDBC/error-codes
(`08006` mean `I/O error during communication`) :thinking: 

WDYT?

AMashenkov (approved):

solveme (commented):
I'm not familiar with code of components that were changed, so  I can't reason about correctness. But chosen approach from UX perspective looks good to me.

PR #3179, opened 2024-05-22
GG-39400 Do not complete foreign connection futures when failing to reserve recovery descriptor

rpuch:
https://ggsystems.atlassian.net/browse/GG-39400

When processing an incoming connection attempt, seeing a future created by another initiator of the same connection and failing to reserve the corresponding connection, we should not complete this future and remove it from the map of such futures. If we do it, we break the 'at most 1 physical connection per connection key' invariant.

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/InboundConnectionHandler.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/InboundConnectionHandler.java
@@ -711,25 +711,19 @@ public class InboundConnectionHandler extends GridNioServerListenerAdapter<Messa
                     ses.send(new RecoveryLastReceivedMessage(ALREADY_CONNECTED));
                 }
                 else {
-                    try {
-                        // The code below causes a race condition between shmem and TCP (see IGNITE-1294)
-                        boolean reserved = recoveryDesc.tryReserve();
+                    // The code below causes a race condition between shmem and TCP (see IGNITE-1294)
+                    boolean reserved = recoveryDesc.tryReserve();
 
-                        if (reserved) {
-                            GridTcpNioCommunicationClient client =
-                                connected(recoveryDesc, ses, rmtNode, msg0.received(), true, !hasShmemClient);
+                    if (reserved) {
+                        GridTcpNioCommunicationClient client =
+                            connected(recoveryDesc, ses, rmtNode, msg0.received(), true, !hasShmemClient);
 
-                            oldFut.onDone(client);
-                        }
-                        else {
-                            ses.send(new RecoveryLastReceivedMessage(ALREADY_CONNECTED));
+                        oldFut.onDone(client);
 
-                            oldFut.onDone();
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/InboundConnectionHandler.java
```diff
@@ -711,25 +711,19 @@ else if (discoverySpi instanceof IgniteDiscoverySpi)
                     ses.send(new RecoveryLastReceivedMessage(ALREADY_CONNECTED));
                 }
                 else {
-                    try {
```
sashapolo, line 714: Why did you remove the try-finally block?

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationConnectionExclusivenessTest.java
```diff
@@ -0,0 +1,453 @@
...
+    /**
+     * Tests a scenario of ensureNoDuplicatesWhenInboundCompetesWithOutbounds(boolean) where
+     * the order of B is higher than the order of A.
+     *
+     * @throws Exception If failed.
+     * @see #ensureNoDuplicatesWhenInboundCompetesWithOutbounds(boolean)
+     */
+    @Test
+    public void noDuplicatesWhenInboundFromHigherNodeCompetesWithOutbounds() throws Exception {
+        ensureNoDuplicatesWhenInboundCompetesWithOutbounds(true);
```
sashapolo, line 117: Can we use a parameterized test?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/InboundConnectionHandler.java
rpuch: Because it was incorrect to always remove `oldFut` from the map. We must only remove it if we managed to create a connection.

sashapolo: ok, so it was originally there just to save us from code duplication?

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationConnectionExclusivenessTest.java
rpuch: Rewrote to Parameterized runner

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/InboundConnectionHandler.java
rpuch: I think it was there *due* to code duplication, as a result of a copy-paste

sashapolo (approved):

PR #3185, opened 2024-05-27
GG-39428 added option to finish maintenance mode with custom exit code.

antkr:


modules/core/src/main/java/org/apache/ignite/internal/maintenance/MaintenanceProcessor.java
```diff
@@ -79,6 +81,9 @@ public class MaintenanceProcessor extends GridProcessorAdapter implements Mainte
     /** Node in maintenance mode will shut down after all active tasks would be completed. */
     private final boolean autoShutdown = getBoolean(IGNITE_MAINTENANCE_AUTO_SHUTDOWN_AFTER_RECOVERY);
 
+    /** Defragmentation regions size percentage of configured ones. */
```
rpuch, line 84: The javadoc does not match the field

modules/core/src/main/java/org/apache/ignite/internal/maintenance/MaintenanceProcessor.java
```diff
@@ -42,8 +42,7 @@
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
 
-import static org.apache.ignite.IgniteSystemProperties.IGNITE_MAINTENANCE_AUTO_SHUTDOWN_AFTER_RECOVERY;
-import static org.apache.ignite.IgniteSystemProperties.getBoolean;
+import static org.apache.ignite.IgniteSystemProperties.*;
```
rpuch, line 45: Do we allow wildcard imports?

modules/core/src/main/java/org/apache/ignite/internal/maintenance/MaintenanceProcessor.java
```diff
@@ -376,6 +381,10 @@ private void startShutdownThread() {
                 while (true) {
                     if (activeTasks.isEmpty() && Ignition.state(igniteInstanceName) == IgniteState.STARTED) {
                         G.stop(igniteInstanceName, false);
+
+                        if (mntcModeExitCode != 0)
+                            System.exit(mntcModeExitCode);
```
Erixonich, line 386: I am not sure, that it works fine. I would try to create an override for G#stop to have optional exit code. But I am ok to have as it is for now as experimental option.

rpuch (approved):

Erixonich (approved):

PR #3186, opened 2024-05-27
GG-39391 Added an ability to filter network interface to be used for TcpCommunication SPI

sk0x50:
https://ggsystems.atlassian.net/browse/GG-39391

This patch adds a new experimental property `networkInterfacesBlacklist` that allows specifying network interfaces that should not be used as a connection target by other nodes when a local address represents a wildcard.

It is possible to use ranges and wildcards in the list of interfaces. Wildcard symbol `*` represents a range of values between
`0` and `255`. For example, `12.12.12.*` refers to addresses from `12.12.12.0` to `12.12.12.255`. The range symbol `-` represents a range of values. For example, `12.12.12.12-24` refers to addresses from `12.12.12.12` to `12.12.12.24`.

Example of `xml` configuration:
```
        <property name="communicationSpi">
            <bean class="org.apache.ignite.spi.communication.tcp.TcpCommunicationSpi">
                <property name="localAddress" value="0.0.0.0"/>
                <property name="networkInterfacesBlacklist">
                    <list>
                        <value>192.168.1.4</value>
                        <value>10.10.1.*</value>
                        <value>172.27.176.10-100</value>
                    </list>
                </property>
            </bean>
        </property>
```

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
@@ -235,7 +235,7 @@ public class CommunicationTcpUtils {
          * @param addr Address to match.
          * @return {@code true} if the address matches the pattern, {@code false} otherwise.
          */
-        abstract boolean match(InetAddress addr);
+        abstract boolean matches(InetAddress addr);
 
         /** {@inheritDoc} */
         @Override public String toString() {
@@ -311,7 +311,7 @@ public class CommunicationTcpUtils {
         }
 
         /** {@inheritDoc} */
-        @Override boolean match(InetAddress addr) {
+        @Override boolean matches(InetAddress addr) {
             String[] segments = addr.getHostAddress().split("\\.");
 
             if (segments.length != 4)
@@ -345,10 +345,15 @@ public class CommunicationTcpUtils {
                 matchers.add(new IPv4Matcher(pattern));
         }
 
-        /** {@inheritDoc} */
+        /**
+         * Returns {@code true} if the given address is not in the blacklist and {@code false} otherwise.
+         *
+         * @param inetAddress Addres to be checked
+         * @return {@code true} if the given address is not in the blacklist and {@code false} otherwise.
+         */
         @Override public boolean apply(InetAddress inetAddress) {
             for (NetworkInterfaceMatcher m : matchers)
-                if (m.match(inetAddress))
+                if (m.matches(inetAddress))
                     return false;
 
             return true;
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
@@ -330,6 +330,9 @@ public class CommunicationTcpUtils {
      * Represents a filter basen on blacklist of network interfaces.
      */
     public static class BlacklistFilter implements IgnitePredicate<InetAddress> {
+        /** Serial version UID. */
+        private static final long serialVersionUID = 0L;
+
         /** List of matchers. */
         private final List<NetworkInterfaceMatcher> matchers;
 
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
@@ -26,14 +26,18 @@ import java.util.LinkedHashSet;
 import java.util.List;
 import java.util.Set;
 import java.util.function.Supplier;
+import java.util.regex.Pattern;
 import org.apache.ignite.IgniteCheckedException;
 import org.apache.ignite.IgniteLogger;
 import org.apache.ignite.IgniteSystemProperties;
 import org.apache.ignite.cluster.ClusterNode;
+import org.apache.ignite.internal.util.lang.gridfunc.AlwaysTruePredicate;
 import org.apache.ignite.internal.util.nio.GridNioException;
 import org.apache.ignite.internal.util.typedef.F;
 import org.apache.ignite.internal.util.typedef.X;
+import org.apache.ignite.internal.util.typedef.internal.S;
 import org.apache.ignite.internal.util.typedef.internal.U;
+import org.apache.ignite.lang.IgnitePredicate;
 import org.apache.ignite.lang.IgniteRunnable;
 import org.apache.ignite.spi.IgniteSpiContext;
 import org.apache.ignite.spi.IgniteSpiOperationTimeoutException;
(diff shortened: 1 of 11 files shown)


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
...
+        @Override public String toString() {
+            return S.toString(NetworkInterfaceMatcher.class, this);
+        }
+    }
+
+    /**
+     * Matcher for IPv4 addresses.
+     */
+    public static class IPv4Matcher extends NetworkInterfaceMatcher {
+        private static final Pattern DIGIT_PATTERN = Pattern.compile("\\d{1,3}");
```
rpuch, line 251: This pattern matches something that is not a decimal digit, but it can be understood as 256-cimal digit. Could it be renamed to `OCTET` to avoid confusion?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
...
+            try {
+                for (int s = 0; s < segments.length; ++s) {
+                    String segment = segments[s];
+
+                    if (DIGIT_PATTERN.matcher(segment).matches()) {
+                        int seg = Integer.parseInt(segment);
+                        if (seg < 0 || seg > IPV4_MAX_OCTET_VALUE)
+                            throw new IllegalArgumentException("Invalid IPv4 address: " + networkInterface());
+
+                        segmentPred[s] = addr -> segment.equals(addr);
```
rpuch, line 288: `addr` seems to be misleading, its a segment of an address

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
...
+                    }
+                    else if (RANGE_PATTERN.matcher(segment).matches()) {
+                        final String[] range = segment.split("-");
+                        if (range.length != 2)
+                            throw new IllegalArgumentException("Invalid IPv4 address: " + networkInterface());
```
rpuch, line 293: But it's not an IPv4 address, it's some generalized pattern of such address

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
...
+
+                        final int min = Integer.parseInt(range[0].trim());
+                        final int max = Integer.parseInt(range[1].trim());
+
+                        segmentPred[s] = addr -> {
```
rpuch, line 298: Segment, not address

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
...
+                if (!segmentPred[s].apply(segments[s]))
+                    return false;
+            }
+
+            return true;
+        }
+    }
+
+    /**
+     * Represents a filter basen on blacklist of network interfaces.
```
rpuch, line 335: Probably we should mention that it returns `true` if an address is not blacklisted, this is not obvious

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
             ", err=" + err +
             ", connectErrs=" + X.getSuppressedList(err) + ']');
     }
+
+    /**
+     * Base class for network interface matchers.
+     */
+    public abstract static class NetworkInterfaceMatcher {
+        /** Pattern of network interface. */
```
rpuch, line 213: Does it make sense to document the pattern syntax somewhere?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
             ", err=" + err +
             ", connectErrs=" + X.getSuppressedList(err) + ']');
     }
+
+    /**
+     * Base class for network interface matchers.
+     */
+    public abstract static class NetworkInterfaceMatcher {
```
rpuch, line 212: Why are these types internal to `CommunicationTcpUtils`?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/TcpCommunicationConfigInitializer.java
```diff
@@ -156,6 +159,39 @@ public TcpCommunicationSpi setLocalAddress(String locAddr) {
         return (TcpCommunicationSpi) this;
     }
 
+    /**
+     * Allows specifying network interfaces that should not be used by Ignite when local address represents a wildcard.
```
rpuch, line 163: That should not be used as a connection target by other nodes?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
```diff
@@ -201,4 +205,161 @@ public static void failNode(ClusterNode nodeToFail,
...
+                            throw new IllegalArgumentException("Invalid IPv4 address: " + networkInterface());
+
+                        segmentPred[s] = addr -> segment.equals(addr);
+                    }
+                    else if (RANGE_PATTERN.matcher(segment).matches()) {
+                        final String[] range = segment.split("-");
+                        if (range.length != 2)
+                            throw new IllegalArgumentException("Invalid IPv4 address: " + networkInterface());
+
+                        final int min = Integer.parseInt(range[0].trim());
```
rpuch, line 295: Should we check that `min < max`?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: Yep, good catch!

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: Renamed to `segmentToCheck`

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: Makes sense to me. Changed to `Invalid IPv4 address pattern: `

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: Changed to `segmentToCheck`

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: This is already mentioned here:
```
        /**
         * Returns {@code true} if the given address is not in the blacklist and {@code false} otherwise.
         *
         * @param inetAddress Address to be checked
         * @return {@code true} if the given address is not in the blacklist and {@code false} otherwise.
         */
        @Override public boolean apply(InetAddress inetAddress)
```
Anyway, it is worth mentioning in the javadoc of the class level too.

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: Updated as follows:
```
    /**
     * Base class for network interface matchers.
     *
     * An implementation of the interface should support using ranges and wildcards in the list of interfaces.
     * Wildcard symbol {@code *} represents a range of values between {@code 0} and {@code 255}. For example, {@code 12.12.12.*} refers to
     * addresses from {@code 12.12.12.0} to {@code 12.12.12.255}. Range symbol {@code -} represents a range of values. For example,
     * {@code 12.12.12.12-24} refers to addresses from {@code 12.12.12.12} to {@code 12.12.12.24}.
     */
    public abstract static class NetworkInterfaceMatcher {
```

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/CommunicationTcpUtils.java
sk0x50: Moved to separate files

rpuch (approved):

PR #3191, opened 2024-06-03
GG-39533 Fixed handling of empty and null network interface patterns

sk0x50:
https://ggsystems.atlassian.net/browse/GG-39533

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/BlacklistFilter.java
+++ b/modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/BlacklistFilter.java
@@ -43,7 +43,7 @@ public class BlacklistFilter implements IgnitePredicate<InetAddress> {
         matchers = new ArrayList<>(blacklist.size());
 
         for (String pattern : blacklist) {
-            if (pattern == null || pattern.isEmpty())
+            if (pattern == null || pattern.trim().isEmpty())
                 continue;
 
             matchers.add(new IPv4Matcher(pattern));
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/BlacklistFilter.java
```diff
@@ -42,8 +42,12 @@ public class BlacklistFilter implements IgnitePredicate<InetAddress> {
     public BlacklistFilter(Collection<String> blacklist) {
         matchers = new ArrayList<>(blacklist.size());
 
-        for (String pattern : blacklist)
+        for (String pattern : blacklist) {
+            if (pattern == null || pattern.isEmpty())
```
solveme, line 46: Let's also verify that `pattern.trim()` is also not empty. Maybe we have a shortcut for this in `IgniteUtils`, so eventually we can write something like:
`if (isNotEmpty(pattern))
    matchers.add(new IPv4Matcher(pattern));`

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/internal/BlacklistNetwokInterfacesTest.java
```diff
@@ -89,4 +89,18 @@ public void testBlacklistFilter() throws Exception {
             IgniteUtils.resetCachedLocalAddressAndHostNames();
         }
     }
+
```
solveme, line 92: let's add test-case for string like "    " (one or more whitespaces)

solveme (changes requested):
Please take a look to my comments. WDYT?

modules/core/src/main/java/org/apache/ignite/spi/communication/tcp/internal/BlacklistFilter.java
sk0x50: nice catch! fixed

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/internal/BlacklistNetwokInterfacesTest.java
sk0x50: test added

solveme (commented):
LGTM!

PR #3199, opened 2024-06-07
GG-39581 Fixed deadlock in Task Processor that can happen during internal task execution

sk0x50:
https://ggsystems.atlassian.net/browse/GG-39581

### DIFF
--- a/modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputeGridMonitorTest.java
+++ b/modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputeGridMonitorTest.java
@@ -17,15 +17,19 @@
 package org.apache.ignite.internal.processors.compute;
 
 import java.util.Collection;
+import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
 import java.util.Queue;
 import java.util.TreeMap;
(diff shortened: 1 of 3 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputeGridMonitorTest.java
```diff
@@ -440,6 +462,58 @@ public ComputeFullWithWaitTask(long timeout) {
         }
     }
 
+    /**
+     * Representa a long-running task that spawns another task.
+     */
+    @GridInternal
+    private static class InternalLongRunningTask extends ComputeTaskAdapter<Void, Void> {
+        /** Latch to indicate that job is started. */
```
tkalkirill, line 470:
```suggestion
        /** Latch to indicate that some first job is started locally. */
```

tkalkirill (approved):

modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputeGridMonitorTest.java
sk0x50: There is only one job that is mapped to the coordinator node.

PR #3219, opened 2024-07-09
GG-34315 OOM caught internally in H2 should not bring node into non operable state 

zstan:


modules/h2/src/main/java/org/gridgain/internal/h2/command/Command.java
```diff
@@ -213,7 +213,6 @@ public ResultInterface executeQuery(int maxrows, boolean scrollable) {
                         // the transaction may be applied partially
                         // in this case we need to panic:
                         // close the database
-                        database.shutdownImmediately();
```
ygerzhedovich, line 216: in previous line we do ``callStop=false;`` - should we remove it ?

zstan: as i can see:
```
                if (callStop) {
                    stop();
                }
                ...
                    public void stop() {
        session.setCurrentCommand(null, false);
        if (!isTransactional()) {
            session.commit(true);
        } else if (session.getAutoCommit()) {
            session.commit(false);
        } else {
            session.unlockReadLocks();
        }
        session.endStatement();
        if (trace.isInfoEnabled() && startTimeNanos > 0) {
            long timeMillis = (System.nanoTime() - startTimeNanos) / 1000 / 1000;
            if (timeMillis > Constants.SLOW_QUERY_LIMIT_MS) {
                trace.info("slow query: {0} ms", timeMillis);
            }
        }
    }
```
thus if we want to close current session - we need to remove this flag, but i don`t  think so. If some sql can`t be executed due to memory consumption - we still need possibility to execute another one.

ygerzhedovich (approved):

PR #3227, opened 2024-07-18
GG-39925 Add 'help' subcommand to --tracing-configuration

solveme:
Also was adjusted format of `--tracing-configuration` usage documentation and added few small optimizations for `TracingConfigurationCoordinates` spawning.


modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/SpanManager.java
```diff
@@ -48,7 +48,7 @@ default Span create(@NotNull SpanType spanType) {
      * @param serializedParentSpan Parent span as serialized bytes.
      * @return Created span.
      */
-    Span create(@NotNull SpanType spanType, @Nullable byte[] serializedParentSpan);
+    Span create(@NotNull SpanType spanType, byte @Nullable [] serializedParentSpan);
```
sk0x50, line 51: From the language point of view, it has the same meaning, right? So what is the reason for this change?

modules/core/src/main/java/org/apache/ignite/spi/tracing/Scope.java
```diff
@@ -63,6 +70,20 @@ public short idx() {
         return idx;
     }
 
+    /**
+     * @return Coordinates without label.
+     */
+    @NotNull public TracingConfigurationCoordinates coordinates() {
```
sk0x50, line 76: Please don't use `@NotNull` annotation. Our coding guidelines prohibit it https://cwiki.apache.org/confluence/display/IGNITE/Coding+Guidelines#CodingGuidelines-@Annotations

modules/core/src/main/java/org/apache/ignite/spi/tracing/Scope.java
```diff
@@ -63,6 +70,20 @@ public short idx() {
...
+        return coordinates;
+    }
+
+    /**
+     * @return Coordinates with optional label.
+     */
+    @NotNull public TracingConfigurationCoordinates coordinates(@Nullable String label) {
```
sk0x50, line 83: I would suggest removing this method and adding a new one to `TracingConfigurationCoordinates` as follows:

```
    /**
     * Creates a new tracing coordinates with the specified label. If the given {@code coordinates} already has the
     * specified label, the same instance is returned.
     *
     * @return Tracing coordinates with the specified label.
     */
    public static TracingConfigurationCoordinates of(TracingConfigurationCoordinates coordinates, String lb) {
        if (Objects.equals(coordinates.label(), lb))
            return coordinates;

        return new TracingConfigurationCoordinates.Builder(coordinates.scope()).withLabel(lb).build();
    }
```
and use it everywhere you need:
```
    TracingConfigurationCoordinates coordinates = TracingConfigurationCoordinates.of(scope.coordinates(), lb);
```

In that case, the code that creates new coordinates is a bit closer to the `TracingConfigurationCoordinates` itself. What do you think?

modules/core/src/main/java/org/apache/ignite/internal/processors/tracing/SpanManager.java
solveme: Just a suggestion of IDEA. I suppose it uses generic pattern for all arrays. In this case you're right it doesn't matter. But in case of, for example, Person[], moving `@Nullable` makes sence. Nullable array and array of nullables are indeed different.

In this particular case I agree, I overcomplicated it. Will fix it.

modules/core/src/main/java/org/apache/ignite/spi/tracing/Scope.java
solveme: Fixed

sk0x50 (approved):

PR #3230, opened 2024-07-21
GG-40027 Added average CP write speed to messages about CP finish

solveme:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointManager.java
```diff
@@ -203,9 +204,7 @@ public CheckpointManager(
 
         checkpointer = checkpointerProvider.get();
 
-        Long cfgCheckpointReadLockTimeout = persistenceCfg != null
-            ? persistenceCfg.getCheckpointReadLockTimeout()
-            : null;
+        Long cfgCheckpointReadLockTimeout = persistenceCfg.getCheckpointReadLockTimeout();
```
solveme, line 207: We call methods of `persistenceCfg` multiple times in this constructor before reaching line 207, that's why null check is redundant here

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointerTest.java
```diff
@@ -0,0 +1,46 @@
...
+        assertEquals("1.06", speedInKb(1024 + 64));     // 1.0625
+        assertEquals("0.88", speedInKb(896));           // 0.875
+        assertEquals("0.0625", speedInKb(64));          // 0.0625
+        assertEquals("0.0078", speedInKb(8));           // 0.0078125
+    }
+
+    private static String speedInKb(long kbs) {
+        return Checkpointer.WriteSpeedFormatter.formatWriteSpeed(U.KB * kbs);
+    }
+}
```
ibessonov, line 46: Please add an empty line at the end

ibessonov (approved):

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/checkpoint/CheckpointerTest.java
solveme: Done

PR #3235, opened 2024-07-23
GG-38496 Externally configure logging for control utility

myskov:
https://ggsystems.atlassian.net/browse/GG-38496

Added a possibility to use custom logging properties file by providing JVM options:

CONTROL_JVM_OPTS=-Djava.util.logging.config.file=<PATH_TO_CONFIG> ./control.sh

### DIFF
--- a/modules/control-utility/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerLoggingTest.java
+++ b/modules/control-utility/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerLoggingTest.java
@@ -24,6 +24,8 @@ import java.util.logging.Handler;
 import java.util.logging.LogManager;
 import java.util.logging.Logger;
 import java.util.logging.StreamHandler;
+import org.apache.ignite.testframework.junits.WithSystemProperty;
+import org.junit.After;
 import org.junit.Test;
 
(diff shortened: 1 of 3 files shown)


modules/control-utility/src/test/java/org/apache/ignite/internal/commandline/CommandHandlerLoggingTest.java
```diff
@@ -0,0 +1,94 @@
...
+        // check that custom logger has no handlers
+        Logger customLogger = LogManager.getLogManager().getLogger(CUSTOM_LOGGER_NAME);
+        assertEquals(0, customLogger.getHandlers().length);
+
+        // check that root logger has exactly one file handler
+        Logger rootLogger = LogManager.getLogManager().getLogger(ROOT_LOGGER_NAME);
+        assertEquals(1, rootLogger.getHandlers().length);
+        assertEquals(FileHandler.class, rootLogger.getHandlers()[0].getClass());
+
+        System.clearProperty("java.util.logging.config.file");
```
vmalin, line 65: Explicit property clearing implies that it won't be cleared automatically which can lead to unexpected consequences should the test fail before reaching this line. It's better to use @WithSystemProperty annotation on the test method

PR #3246, opened 2024-08-06
GG-39863 erase empty cache directory after cache deletion

odiszapc:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStore.java
```diff
@@ -291,8 +292,6 @@ private long checkFile(FileIO fileIO, File cfgFile) throws IOException {
 
             fileIO = null;
 
-            if (delete)
-                Files.delete(pathProvider.apply());
```
ibessonov, line 295: Can you please explain why you removed these two lines?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1093,8 +1095,13 @@ private IgniteCheckedException shutdown(Collection<CacheStoreHolder> holders, bo
      */
     private IgniteCheckedException shutdown(CacheStoreHolder holder, boolean cleanFile,
         @Nullable IgniteCheckedException aggr) {
+
         aggr = shutdown(holder.idxStore, cleanFile, aggr);
 
+        if (cleanFile) {
+            cleanParentDirectories(holder);
+        }
```
ibessonov, line 1103: What's the reason behind deleting a directory **before** you close page store instances for its content? This code looks broken, I don't understand your intentions

modules/core/src/test/java/org/apache/ignite/cache/CacheDirectoryCleanupTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.ignite.internal.processors.cache.persistence.file.FilePageStoreManager;
+import org.apache.ignite.testframework.junits.WithSystemProperty;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_DISABLE_WAL_DURING_REBALANCING;
+
+/** */
+@WithSystemProperty(key = IGNITE_DISABLE_WAL_DURING_REBALANCING, value = "false")
+public class CacheDirectoryCleanupTest extends GridCommonAbstractTest {
```
ibessonov, line 37: Please add your test to a suite, it won't run on TC right now, I believe you didn't check it

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1093,8 +1095,13 @@ private IgniteCheckedException shutdown(Collection<CacheStoreHolder> holders, bo
      */
     private IgniteCheckedException shutdown(CacheStoreHolder holder, boolean cleanFile,
         @Nullable IgniteCheckedException aggr) {
+
```
ibessonov, line 1098:
```suggestion
        @Nullable IgniteCheckedException aggr
    ) {
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1103,6 +1110,23 @@ private IgniteCheckedException shutdown(CacheStoreHolder holder, boolean cleanFi
         return aggr;
     }
 
+    private void cleanParentDirectories(CacheStoreHolder holder) {
+        Arrays.stream(holder.partStores).map(store -> ((FilePageStoreV2)store).getPath().getParent()).distinct().forEach(new Consumer<Path>() {
```
ibessonov, line 1114: Doing such class casts is a bad habit, maybe we should introduce new method to interface

ibessonov, line 1114: All files in a store should have the same parent, it should be enough to read it for index partition, for example.
Or, if you want to be maximally flexible, you should use `findAny` instead of `distinct` stream.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1103,6 +1110,23 @@ private IgniteCheckedException shutdown(CacheStoreHolder holder, boolean cleanFi
...
+            @Override public void accept(Path dir) {
+                if (Files.isDirectory(dir, LinkOption.NOFOLLOW_LINKS)) {
```
ibessonov, line 1116: I believe you can reuse `org.apache.ignite.internal.util.IgniteUtils#delete(java.nio.file.Path)`, it is recursive

modules/core/src/test/java/org/apache/ignite/cache/CacheDirectoryCleanupTest.java
```diff
@@ -0,0 +1,87 @@
...
+import org.apache.ignite.internal.IgniteEx;
+import org.apache.ignite.internal.processors.cache.persistence.file.FilePageStoreManager;
+import org.apache.ignite.testframework.junits.WithSystemProperty;
+import org.apache.ignite.testframework.junits.common.GridCommonAbstractTest;
+import org.junit.Test;
+
+import static org.apache.ignite.IgniteSystemProperties.IGNITE_DISABLE_WAL_DURING_REBALANCING;
+
+/** */
+@WithSystemProperty(key = IGNITE_DISABLE_WAL_DURING_REBALANCING, value = "false")
```
ibessonov, line 36: Why did you add this?

modules/core/src/test/java/org/apache/ignite/cache/CacheDirectoryCleanupTest.java
```diff
@@ -0,0 +1,87 @@
...
+        ignite1.createCache(new CacheConfiguration(DEFAULT_CACHE_NAME));
+        IgniteCache c = ignite1.cache(DEFAULT_CACHE_NAME);
+        c.put(30000, 30000);
+
+        c.destroy();
+
+        File workDir = ((FilePageStoreManager)ignite1.context().cache().context().pageStore()).workDir();
+        assertFalse(new File(workDir, "cache-" + DEFAULT_CACHE_NAME).exists());
+    }
+
```
ibessonov, line 86:
```suggestion
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStore.java
odiszapc: mistake. Reverted

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
odiszapc: mistake. fixed

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
```diff
@@ -1093,17 +1096,35 @@ private IgniteCheckedException shutdown(Collection<CacheStoreHolder> holders, bo
...
+            cleanParentDirectories(holder);
+
         return aggr;
     }
 
+    private void cleanParentDirectories(CacheStoreHolder holder) {
+        Arrays.stream(holder.partStores).map(store -> store.getPath().getParent()).findAny().ifPresent(dir -> {
+            if (Files.isDirectory(dir, LinkOption.NOFOLLOW_LINKS)) {
+                try {
+                    if (!Files.list(dir).findFirst().isPresent())
```
ibessonov, line 1118: I see, you only remove the directory if it's not empty. What happens to `cache_data.dat` file, do we remove it when we delete cache? I'm not sure I know where that happens.

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsDestroyCacheTest.java
```diff
@@ -394,4 +396,18 @@ public void cacheDestroyWithConcImplicitTx() throws Exception {
...
+        IgniteEx ignite = startGrids(1);
+        ignite.cluster().state(ClusterState.ACTIVE);
+
+        ignite.createCache(new CacheConfiguration(DEFAULT_CACHE_NAME));
+        IgniteCache cache = ignite.cache(DEFAULT_CACHE_NAME);
+        cache.put(30000, 30000);
+
+        cache.destroy();
+
+        assertFalse(new File(((FilePageStoreManager)ignite.context().cache().context().pageStore()).workDir(), "cache-" + DEFAULT_CACHE_NAME).exists());
```
ibessonov, line 411: This line seems too long

ibessonov: And just in case, please assert that this dir exists before destroying the cache, so that there's no doubt that there are no error in test itself

ibessonov (approved):
Please expand the test a little bit, and then everything's fine. Giving you an approval in advance

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java
odiszapc: Yes, because `cache_data.dat` is removed a little bit earlier as a part of entire cache decommission process here: https://github.com/gridgain/gridgain/blob/c942a4f899b5228a6c8084dfe58f34f9de6afdc7/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/file/FilePageStoreManager.java#L1166-L1166

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgnitePdsDestroyCacheTest.java
odiszapc: added additional assert

ibessonov (approved):

PR #3249, opened 2024-08-07
GG-39128 Fix lambda classname parsing in deployment for Java 21+

ibessonov:
https://ggsystems.atlassian.net/browse/GG-39128


Phillippko (approved):
Lgtm

PR #3250, opened 2024-08-12
GG-34115 Java thin: Add AtomicSequence

ptupitsyn:
* Add `ClientAtomicSequence` interface
  * Same methods as `IgniteAtomicSequence`
  * Client-size range reservation (increments are local within range)
* Add `IgniteClient#atomicSequence`


modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientAtomicSequenceImpl.java
```diff
@@ -0,0 +1,215 @@
...
+                in -> in.in().readLong());
+
+        upBound = locVal;
+    }
+
+    /** {@inheritDoc} */
+    @Override public long get() throws IgniteException {
+        checkRemoved();
+
+        return locVal;
```
isapego, line 74: What if the value was updated remotely?

isapego (approved):

PR #3254, opened 2024-08-16
GG-37635 Fix ThreadLocalAllocator buffer shrinking

AMashenkov:


modules/core/src/test/java/org/apache/ignite/internal/binary/streams/BinaryMemoryAllocatorTest.java
```diff
@@ -0,0 +1,163 @@
...
+    @Test
+    public void threadLocalBufferCompaction() throws Exception {
+        resetThreadLocalBuffer();
+
+        BinaryMemoryAllocatorChunk buffer = threadLocalMemoryAllocatorBuffer();
+
+        // Allocate large object.
+        allocateThenRelease(buffer, LARGE_OBJECT_SIZE);
+
+        assertTrue(bufferSize(buffer) >= LARGE_OBJECT_SIZE);
```
ygerzhedovich, line 47: Looks like the test can be flaky.
1. ``assertTrue(bufferSize(buffer) >= LARGE_OBJECT_SIZE);`` can fail due to buffer will be shrinked earlier
2. `` Thread.sleep(1_000);`` don't bring any guarantee that shrink will be done.

ygerzhedovich (approved):

PR #3258, opened 2024-08-22
GG-40026 Type for a column 'DT' is not compatible with index definition. Expected 'Timestamp', actual type 'Date'

AMashenkov:


ygerzhedovich (approved):

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/IgniteQueryConvertibleTypesValidationTest.java
```diff
@@ -0,0 +1,188 @@
...
+
+        grid(0).cache(DEFAULT_CACHE_NAME).put(1, new Data(1, testObj));
+
+        grid(0).cache(DEFAULT_CACHE_NAME).put(2, grid(0).binary().toBinary(new Data(2, testObj)));
+
+        List<List<?>> selectData = execute("SELECT data FROM DATA");
+
+        Object sqlObj = sqlTypeConverter.apply(testObj);
+
+        assertTrue(selectData.get(0).stream().allMatch(d -> Objects.equals(sqlObj, d)));
```
zstan, line 107: seems we need to check all results, not only "0" ?

zstan (approved):
overall looks good, plz fix the mentioned test

PR #3270, opened 2024-09-05
GG-36974 [IGNITE-19811] Continuous queries backup acknowledge message sending fails for expired entries

ygerzhedovich:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
```diff
@@ -1431,27 +1424,42 @@ private void sendBackupAcknowledge(final IgniteBiTuple<Map<Integer, Long>, Set<A
...
 
-                                    if (log.isDebugEnabled())
-                                        log.debug("Failed to send acknowledge message, node left " +
-                                            "[msg=" + msg + ", node=" + node + ']');
-                                }
-                                catch (IgniteCheckedException e) {
-                                    IgniteLogger log = ctx.log(CU.CONTINUOUS_QRY_LOG_CATEGORY);
+                        try {
+                            for (ClusterNode node : ctx.discovery().cacheGroupAffinityNodes(cctx.groupId(), topVer)) {
+                                if (!node.isLocal() && cctx.discovery().alive(nodeId)) {
```
zstan, line 1435: do we need additional check for nodeId nullability here ?

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryEntriesExpireTest.java
```diff
@@ -0,0 +1,76 @@
...
+    public void testBackupQOnEntriesExpire() throws Exception {
+        IgniteEx srv0 = startGrid(0);
+        IgniteEx srv1 = startGrid(1);
+
+        IgniteCache<Object, Object> cache = srv0.getOrCreateCache(new CacheConfiguration<>(DEFAULT_CACHE_NAME)
+                .setBackups(1)
+                .setAffinity(new RendezvousAffinityFunction().setPartitions(10)))
+            .withExpiryPolicy(new CreatedExpiryPolicy(new Duration(TimeUnit.MILLISECONDS, 1)));
+
+        srv1.cache(DEFAULT_CACHE_NAME).query(new ContinuousQuery<>().setLocalListener(evts -> {}).setIncludeExpired(true));
```
zstan, line 54: this test is not clear for me, why do e need to set "setIncludeExpired" here ?
also issue is about expired items, but fix is about limitation for message sending, did i miss smth ?

ygerzhedovich: Because in the original ticket, the patch fixed the issue with the expired entity and has the tests. 
The issue appears after implementing some optimization for CQ (IGNITE-13126), but we didn't port it yet. In any case the test can help us in the future if we meet an issue like was introduced earlier in AI2.
Meanwhile GG-36974 have a few comments about these.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
ygerzhedovich: no `nodeId` can't be null.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryHandler.java
```diff
@@ -20,14 +20,7 @@
...
 import java.io.ObjectOutput;
-import java.util.ArrayList;
-import java.util.Collection;
-import java.util.Collections;
-import java.util.Iterator;
-import java.util.List;
-import java.util.Map;
-import java.util.Set;
-import java.util.UUID;
+import java.util.*;
```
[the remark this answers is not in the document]
ygerzhedovich: fixed

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryEntriesExpireTest.java
zstan: Of course i have read this comments but as for me such approach is confusing, imagine that this test become flaky, someone will annotate code and found ... only these not relevant changes, btw did you check this test after IGNITE-13126 ? or we blindly port it ?

zstan (approved):

modules/core/src/test/java/org/apache/ignite/internal/processors/cache/query/continuous/CacheContinuousQueryEntriesExpireTest.java
ygerzhedovich: yes, I've checked it

PR #3276, opened 2024-09-11
GG-40379 CPP: Add BinaryReader::IsNull and BinaryRawReader::SkipIfNull

isapego:
https://ggsystems.atlassian.net/browse/GG-40379

- Added `BinaryReader::IsNull` and `BinaryRawReader::SkipIfNull` methods to check for `null`s;
- Modified tests to check new methods.

### DIFF
--- a/modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
+++ b/modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
@@ -66,9 +66,9 @@ namespace ignite
             BinaryRawReader(ignite::impl::binary::BinaryReaderImpl* impl);
 
             /**
-             * Skip next value if it is the null.
+             * Skip the next value if it is null.
              *
-             * @return True if the null value has been detected and skipped.
+             * @return @c true if a null value has been detected and skipped, and @c false otherwise.
              */
             bool SkipIfNull();
 
--- a/modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
+++ b/modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
@@ -64,7 +64,14 @@ namespace ignite
              * @param impl Implementation.
              */
             BinaryRawReader(ignite::impl::binary::BinaryReaderImpl* impl);
-                        
+
+            /**
+             * Skip next value if it is the null.
+             *
+             * @return True if the null value has been detected and skipped.
+             */
+            bool SkipIfNull();
+
             /**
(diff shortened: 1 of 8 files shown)


modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
```diff
@@ -64,7 +64,14 @@ namespace ignite
              * @param impl Implementation.
              */
             BinaryRawReader(ignite::impl::binary::BinaryReaderImpl* impl);
-                        
+
+            /**
+             * Skip next value if it is the null.
```
ptupitsyn, line 69:
```suggestion
             * Skip the next value if it is null.
```

modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
```diff
@@ -64,7 +64,14 @@ namespace ignite
...
+             *
+             * @return True if the null value has been detected and skipped.
```
ptupitsyn, line 71:
```suggestion
             * @return True if a null value has been detected and skipped.
```

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
```diff
@@ -777,6 +777,24 @@ namespace ignite
...
+
+                int32_t fieldId = idRslvr->GetFieldId(typeId, fieldName);
+                int32_t fieldPos = FindField(fieldId);
+
+                if (fieldPos <= 0)
+                    return true;
+
+                stream->Position(fieldPos);
+
+                return SkipIfNull();
```
ptupitsyn, line 795: `SkipIfNull` will advance the position, is that ok here?

ptupitsyn (changes requested):

modules/platforms/cpp/binary/src/impl/binary/binary_reader_impl.cpp
isapego: Yeah, there is a position guard in the beginning. It makes sure the position is going to be restored when the function is left.

modules/platforms/cpp/binary/include/ignite/binary/binary_raw_reader.h
isapego: Yeah, copied old comment without checking. My bad. Fixed.

ptupitsyn (approved):

PR #3284, opened 2024-09-18
GG-40372 Update links to gridgain documentation

sk0x50:
https://ggsystems.atlassian.net/browse/GG-40372

### DIFF
--- a/examples/config/servlet/README.txt
+++ b/examples/config/servlet/README.txt
@@ -5,4 +5,4 @@ This folder contains web.xml file that demonstrates how to configure any servlet
 to start a GridGain node inside a Web application.
 
 For more information on available configuration properties, etc. refer to our documentation:
-http://apacheignite.readme.io/docs/web-session-clustering
+https://www.gridgain.com/docs/latest/developers-guide/configuring-caches/configuration-overview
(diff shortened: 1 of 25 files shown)


examples/config/servlet/README.txt
```diff
@@ -5,4 +5,4 @@ This folder contains web.xml file that demonstrates how to configure any servlet
 to start a GridGain node inside a Web application.
 
 For more information on available configuration properties, etc. refer to our documentation:
-http://apacheignite.readme.io/docs/web-session-clustering
+https://www.gridgain.com/docs/latest/developers-guide/configuring-caches/configuration-overview
```
ygerzhedovich, line 8: looks like we have lost documentation about Web Session Clustering. Should we ask Doc team to make it and put the right link here?

sk0x50: Hi @ygerzhedovich ,

I have already asked the Doc team about this.
This legacy page was transformed into this one (that was the answer).

The initial page provides two following major things:
 - use case description
 - common cache configuration aspects (+ the required artifacts that should be added to the user project, such as ignite-web)

So, I would say that we only lost the first part. Honestly, I don't know it is a critical use case or not.
By the way, the official page of Apache Ignite does not mention this use case https://ignite.apache.org/features/.

ygerzhedovich (approved):

PR #3287, opened 2024-09-26
GG-39772 Avoid exceptions in page store metrics

ibessonov:
https://ggsystems.atlassian.net/browse/GG-39772

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/pagemem/store/PageStoreCollection.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/pagemem/store/PageStoreCollection.java
@@ -33,8 +33,7 @@ public interface PageStoreCollection {
 
     /**
      * @param grpId Cache group ID.
-     * @return Collection of related page stores.
-     * @throws IgniteCheckedException If failed.
+     * @return Collection of related page stores or {@code null} if it's not found.
      */
-    public Collection<PageStore> getStores(int grpId) throws IgniteCheckedException;
+    public Collection<PageStore> getStores(int grpId);
 }
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
@@ -2078,9 +2078,11 @@ public class GridCacheDatabaseSharedManager extends IgniteCacheDatabaseSharedMan
     }
 
     /**
+     * Returns accumulated results for all page stores. Ignores all exceptions that could happen.
+     *
      * @param gctx Group context.
      * @param f Consumer.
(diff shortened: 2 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/pagemem/store/PageStoreCollection.java
```diff
@@ -33,8 +33,7 @@ public interface PageStoreCollection {
 
     /**
      * @param grpId Cache group ID.
-     * @return Collection of related page stores.
-     * @throws IgniteCheckedException If failed.
+     * @return Collection of related page stores or {@code null} if it's not found.
```
sashapolo, line 36: Does it make sense to return an empty collection instead of `null`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -2080,24 +2080,25 @@ public PageStore getPageStore(int grpId, int partId) throws IgniteCheckedExcepti
     /**
      * @param gctx Group context.
      * @param f Consumer.
-     * @return Accumulated result for all page stores.
+     * @return Accumulated result for all page stores. Ignores all exceptions that could happen.
```
sashapolo, line 2083: I'd suggest to move this to the main part of the javadoc

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/GridCacheDatabaseSharedManager.java
```diff
@@ -2080,24 +2080,25 @@ public PageStore getPageStore(int grpId, int partId) throws IgniteCheckedExcepti
...
 
-            if (stores != null) {
-                for (PageStore store : stores)
+        if (stores != null) {
+            for (PageStore store : stores) {
+                try {
                     res += f.applyAsLong(store);
+                }
+                catch (Exception ignore) {
+                    // Intentionally ignore it.
```
sashapolo, line 2098: Do we want to log the exception at least? What kind of exceptions are we expecting here?

ibessonov: We might get occasional `FileNotFoundException`, very rare. I wouldn't want to log that

modules/core/src/main/java/org/apache/ignite/internal/pagemem/store/PageStoreCollection.java
ibessonov: I wouldn't do that. Empty collection means "there's something and it's empty", while null means "there's nothing". Also, we already use nulls there

sashapolo (approved):

PR #3290, opened 2024-09-30
GG-40472 Fix NPE in tx prepare on client node disconnect

isapego:
Fix the NPE in case the node was disconnected from the cluster.


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/near/GridNearPessimisticTxPrepareFuture.java
```diff
@@ -173,6 +175,10 @@ private MiniFuture miniFuture(int miniId) {
 
     /** {@inheritDoc} */
     @Override public void prepare() {
+        GridCacheMvccManager mvcc = cctx.mvcc();
+        if (mvcc == null)
+            throw new IgniteException("Locking manager is not available (probably, disconnected from a cluster)");
```
ptupitsyn, line 180:
```suggestion
            throw new IgniteException("Locking manager is not available (probably disconnected from the cluster)");
```

ptupitsyn (approved):

PR #3293, opened 2024-09-30
GG-39629 Disable WAL on rebalance even though the corresponding exchange relates to changing baseline and enabled transaction tracker

sk0x50:
https://ggsystems.atlassian.net/browse/GG-39629


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -2785,9 +2785,9 @@ public LocalPendingTransactionsTracker pendingTxsTracker() {
     }
 
     /**
-     * Enables pending transactions tracker.
+     * Enables pending transaction's tracker.
```
denis-chudov, line 2788:
```suggestion
     * Enables pending transactions' tracker.
```

denis-chudov (approved):

PR #3297, opened 2024-10-04
GG-40720 Fix logging tx markers when point-in-time recovery is enabled

sk0x50:
https://ggsystems.atlassian.net/browse/GG-40720

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
@@ -2846,7 +2846,7 @@ public class IgniteTxManager extends GridCacheSharedManagerAdapter {
      */
     @Nullable WALPointer logTxRecord(IgniteTxAdapter tx) {
         // Skip logging tx state change to WAL if required.
-        if (cctx.wal() == null || canSkipTxRecordLogging(tx))
+        if (cctx.wal() == null || !txRecordLoggingNeeded(tx))
             return null;
 
         Map<Short, Collection<Short>> nodes = tx.consistentIdMapper.mapToCompactIds(tx.topVer, tx.txNodes);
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -2873,8 +2868,22 @@ record = new TxRecord(tx.state(), tx.nearXidVersion(), tx.writeVersion(), nodes)
         }
     }
 
-    /** Checks if a record belongs to a persistent cache with WAL enabled. */
-    private boolean containsCacheWithEnabledWal(IgniteTxAdapter tx) {
+    /**
+     * Returns {@code true} if it is possible to skip logging of transaction record to WAL.
+     *
```
denis-chudov, line 2873: I would expect to see the criteria in javadoc: why tx logging can be omitted?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -2873,8 +2868,22 @@ record = new TxRecord(tx.state(), tx.nearXidVersion(), tx.writeVersion(), nodes)
...
+     * @param tx Transaction.
+     * @return {@code true} if it is possible to skip logging of transaction record to WAL.
+     */
+    private boolean canSkipTxRecordLogging(IgniteTxAdapter tx) {
```
denis-chudov, line 2877: I am not sure that semantics of this method's return value should have been inverted. Maybe leave it as something like `txRecordLoggingNeeded`?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
```diff
@@ -2873,8 +2868,22 @@ record = new TxRecord(tx.state(), tx.nearXidVersion(), tx.writeVersion(), nodes)
...
+        BaselineTopology baselineTop = cctx.kernalContext().state().clusterState().baselineTopology();
+
+        if (baselineTop == null || !baselineTop.consistentIds().contains(cctx.localNode().consistentId()))
+            return true;
+
+        // Point-in-time recovery is enabled, so we have to log all transaction markers.
+        if (cctx.snapshot().needTxReadLogging())
```
denis-chudov, line 2884: Also regarding the method semantics: does it mean that, if `needTxReadLogging`, then tx logging cannot be skipped? I think this invertion is not needed, it makes the code complicated

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
sk0x50: Please check the updated javadoc:
```
    /**
     * Returns {@code true} if the given transaction should be logged in WAL and {@code false} otherwise.
     * Transaction marker should be logged in WAL if the following conditions are met:
     * <ul>
     *     <li>local node is a part of baseline topology.</li>
     *     <li>point-in-time recovery is enabled,
     *     or there is a cache enlisted in the transaction with persistence and WAL enabled for the corresponding cache group.</li>
     * </ul>
     * @param tx Transaction.
     * @return {@code true} if the given transaction should be logged in WAL.
     */
    private boolean txRecordLoggingNeeded(IgniteTxAdapter tx) {
```

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
sk0x50: Renamed.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/transactions/IgniteTxManager.java
sk0x50: > Also regarding the method semantics: does it mean that, if needTxReadLogging, then tx logging cannot be skipped?

Yes, it does. 

> I think this invertion is not needed, it makes the code complicated

Ok. Fixed.

denis-chudov (approved):

PR #3300, opened 2024-10-08
GG-37424 remove support of TRIGGER from H2 parser

ygerzhedovich:


modules/h2/src/main/java/org/gridgain/internal/h2/command/Parser.java
```diff
@@ -5739,9 +5726,7 @@ private Prepared parseCreate() {
             return parseCreateSequence();
         } else if (readIf("USER")) {
             return parseCreateUser();
-        } else if (readIf("TRIGGER")) {
-            return parseCreateTrigger(force);
-        } else if (readIf("ROLE")) {
+        }else if (readIf("ROLE")) {
```
AMashenkov, line 5729:
```suggestion
        } else if (readIf("ROLE")) {
```

AMashenkov (approved):

modules/h2/src/main/java/org/gridgain/internal/h2/command/Parser.java
ygerzhedovich: thanks. fixed

PR #3316, opened 2024-10-30
GG-41023: Fix CVE-2024-38820 by updating spring.version to 5.3.41

p-hash:


p-hash:
5.3.41 is only available for spring enterprise

PR #3317, opened 2024-11-04
GG-39209 Implement logging of fingerprinting

vmalin:


vldpyatkov (approved):

modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2270,14 +2270,27 @@ private void ackNodeMetrics(DecimalFormat dblFmt,
 
             SB msg = new SB();
 
+            final ClusterNode crdNode = ctx.discovery().discoCache().oldestAliveServerNode();
+
+            String crdInfoStr;
+
+            if (crdNode != null) {
+                crdInfoStr = "    ^-- Coordinator [id=" + crdNode.id() + ", consistentId=" + crdNode.consistentId() +
+                    ", version=" + crdNode.version().toString() + "]";
```
vldpyatkov, line 2279: Here we can skip toString().

PR #3349, opened 2024-12-04
GG-41256 fixed transition from ACTIVE_READ_ONLY to ACTIVE on adding new node

antkr:
…ew node.


modules/core/src/test/java/org/apache/ignite/internal/processors/cache/persistence/IgniteBaselineAffinityTopologyActivationTest.java
```diff
@@ -380,6 +381,99 @@ public void testNodeWithOldBltIsAllowedToJoinCluster() throws Exception {
...
+
+        grid.cluster().state(ClusterState.ACTIVE_READ_ONLY);
+
+        startGridWithConsistentId("C");
+
+        long topologyVersion = grid.cluster().topologyVersion();
+
+        try {
+            GridTestUtils.runAsync(
+                    () -> grid.cluster().setBaselineTopology(topologyVersion)
```
sk0x50, line 405: Perhaps, it makes sense to check `void setBaselineTopology(Collection<? extends BaselineNode> baselineTop)` also.

vldpyatkov (approved):

sk0x50 (approved):

PR #3351, opened 2024-12-06
GG-41320 Fix NPE when running 'drop column' DDL from thick client node

solveme:


modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -939,11 +940,7 @@ else if (cmdH2 instanceof GridSqlAlterTableDropColumn) {
                 else {
                     assert tbl.rowDescriptor() != null;
 
-                    GridCacheContext cctx = tbl.cacheContext();
```
ygerzhedovich, line 942: For not started caches we have a special structure GridCacheContextInfo which can provide base information about the cache without starting cache itself. I see that you already use it in a new introduced method below. For consistency with the other parts I propose to introduce the `mvccEnabled` method in the GridCacheContextInfo and use it instead of current GridCacheContext

solveme: Agree with proposal. Please check out last commit.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/CommandProcessor.java
```diff
@@ -1001,6 +996,16 @@ else if (cmdH2 instanceof GridSqlAlterTableDropColumn) {
...
     }
 
+    /**
+     * Check if table has MVCC enabled.
+     *
+     * @param tbl Table.
+     * @return True if MVCC is enabled.
+     */
+    private static boolean isMvccEnabled(GridH2Table tbl) {
+        return tbl.cacheInfo().mvccEnabled();
```
ygerzhedovich, line 1006: the method looks redundant, but up to you

ygerzhedovich (approved):

PR #3354, opened 2024-12-11
GG-41421 Improve error handling of partition reconciliation tool

sk0x50:
https://ggsystems.atlassian.net/browse/GG-41421

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/AbstractPipelineProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/AbstractPipelineProcessor.java
@@ -24,6 +24,7 @@ import java.util.concurrent.Semaphore;
 import java.util.concurrent.TimeUnit;
 import java.util.concurrent.atomic.AtomicReference;
 import org.apache.ignite.IgniteCheckedException;
+import org.apache.ignite.IgniteException;
 import org.apache.ignite.IgniteLogger;
 import org.apache.ignite.cluster.ClusterGroup;
 import org.apache.ignite.cluster.ClusterNode;
@@ -32,6 +33,7 @@ import org.apache.ignite.internal.GridKernalContext;
 import org.apache.ignite.internal.IgniteEx;
 import org.apache.ignite.internal.processors.affinity.AffinityTopologyVersion;
 import org.apache.ignite.internal.processors.cache.GridCachePartitionExchangeManager;
+import org.apache.ignite.internal.processors.cache.IgniteInternalCache;
 import org.apache.ignite.internal.processors.cache.checker.objects.CachePartitionRequest;
 import org.apache.ignite.internal.processors.cache.checker.objects.ExecutionResult;
 import org.apache.ignite.internal.processors.cache.checker.util.DelayedHolder;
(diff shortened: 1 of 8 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/AbstractPipelineProcessor.java
```diff
@@ -210,12 +212,14 @@ protected PipelineWorkload pollTask(long timeout, TimeUnit unit) throws Interrup
...
      * @param lsnr Listener.
+     * @throws InterruptedException If task was interrupted.
+     * @throws IgniteException If failed to execute the task.
      */
     protected <T extends CachePartitionRequest, R> void compute(
         Class<? extends ComputeTask<T, ExecutionResult<R>>> taskCls,
         T workload,
         IgniteInClosure<? super R> lsnr
-    ) throws InterruptedException {
+    ) throws InterruptedException, IgniteException {
```
sanpwc, line 222: IgniteException is a RuntimeException right?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/AbstractPipelineProcessor.java
```diff
@@ -317,10 +331,16 @@ protected void printStatistics() {
...
      */
-    private ClusterGroup partOwners(String cacheName, int partId) {
-        Collection<ClusterNode> nodes = ignite.cachex(cacheName).context().topology().owners(partId, startTopVer);
+    private ClusterGroup partOwners(String cacheName, int partId) throws IgniteException {
+        IgniteInternalCache<?, ?> cachex = ignite.cachex(cacheName);
+
+        if (cachex == null)
+            throw new IgniteException("Cache not found (was stopped) [name=" + cacheName + ']');
+
+        Collection<ClusterNode> nodes = cachex.context().topology().owners(partId, startTopVer);
```
sanpwc, line 343: Do we handle cache owners changes during further task processing?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/ReconciliationResultCollector.java
```diff
@@ -101,12 +104,13 @@ public interface ReconciliationResultCollector {
      * @param partId Partition id.
      * @param conflicts Conflicted entries.
      * @param actualKeys Actual values.
+     * @throws IgniteException If cache with the given name does not exist.
```
sanpwc, line 107: It seems a bit too generic to consider that IgniteException will be thrown only if cache with the given name does not exist, even if it's true for now. I'd rather create CacheNotFound extends IgniteException or similar. Up to you.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/AbstractPipelineProcessor.java
sk0x50: That is right.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/AbstractPipelineProcessor.java
sk0x50: We calculate owners for the given topology. Changing the owners means changing the cluster topology. Therefore, this partition reconciliation session will be canceled.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/checker/processor/ReconciliationResultCollector.java
sk0x50: It might be a good idea, but IMHO, creating a new exception class for the sake of one or two methods seems overkill.

sanpwc (approved):

PR #3358, opened 2024-12-13
GG-41422 Fix a hanging of removeAll when there are lost partitions

sk0x50:
https://ggsystems.atlassian.net/browse/GG-41422


alievmirza (approved):
Discussed in private, patch looks good, thank you!

PR #3359, opened 2024-12-15
GG-36352 Enhance error handling in GridMapQueryExecutor with detailed logging

dy2285:
https://ggsystems.atlassian.net/browse/GG-36352

- Implemented logQueryDetails method for centralized error logging.
- Integrated logQueryDetails into key error-handling blocks in GridMapQueryExecutor.
- Added unit tests:
`testLogQueryDetailsDirectly`
`testLogQueryDetailsWithNullInputs`
`testLogQueryDetailsOnQueryFailure`

### DIFF
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -23,6 +23,11 @@
     <property name="charset" value="UTF-8"/>
 
     <property name="fileExtensions" value="java, properties, xml"/>
+    
+    <module name="SuppressionFilter">
+        <property name="file" value="checkstyle/checkstyle-suppressions.xml"/>
+    </module>
+
 
     <!-- Whitespaces Checks. See: http://checkstyle.sourceforge.net/config_whitespace.html -->
     <module name="FileTabCharacter">
--- a/checkstyle/checkstyle.xml
+++ b/checkstyle/checkstyle.xml
@@ -23,11 +23,6 @@
     <property name="charset" value="UTF-8"/>
 
     <property name="fileExtensions" value="java, properties, xml"/>
-    
-    <module name="SuppressionFilter">
-        <property name="file" value="checkstyle/checkstyle-suppressions.xml"/>
-    </module>
-
 
     <!-- Whitespaces Checks. See: http://checkstyle.sourceforge.net/config_whitespace.html -->
     <module name="FileTabCharacter">
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -210,7 +210,7 @@ public class GridMapQueryExecutor {
                 .append("\nLabel: ").append(label != null ? label : "N/A")
                 .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
                 .append("\nQueries: ").append(
-                        queries != null
+                        !F.isEmpty(queries)
                                 ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
                                 : "N/A")
                 .append("\nLocal Node ID: ").append(localNodeId)
@@ -225,7 +225,7 @@ public class GridMapQueryExecutor {
         }
 
         if (error != null)
-            logMessage.append("\nError: ").append(error.getMessage());
+            logMessage.append("\nError: ");
 
         return logMessage.toString();
 
@@ -388,7 +388,7 @@ public class GridMapQueryExecutor {
      * @param maxMem Query memory limit.
      * @param runningQryId Running query id.
      */
-    protected void onQueryRequest0(
+    private void onQueryRequest0(
         final ClusterNode node,
         final long reqId,
         final String label,
@@ -633,7 +633,15 @@ public class GridMapQueryExecutor {
                         if (qryRetryErr != null)
                             sendError(node, reqId, qryRetryErr);
                         else {
-                            String errMsg = buildQueryLogDetails(reqId, label, schemaName, qrys, params, e, node.id(),ctx.localNodeId());
+                            String errMsg = buildQueryLogDetails(
+                                    reqId,
+                                    label,
+                                    schemaName,
+                                    qrys,
+                                    params,
+                                    e,
+                                    node.id(),
+                                    ctx.localNodeId());
 
                             if (e instanceof Error) {
                                 U.error(log, errMsg, e);
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -193,7 +193,7 @@ public class GridMapQueryExecutor {
      * @param params Query parameters, if any.
      * @param error Exception that occurred during the query execution.
      */
-    String buildQueryLogDetails(
+    static String buildQueryLogDetails(
             long reqId,
             String label,
             String schemaName,
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -191,29 +191,30 @@ public class GridMapQueryExecutor {
      * @param error Exception that occurred during the query execution.
      */
     protected void logQueryDetails(
-            long reqId,
-            String label,
-            String schemaName,
-            Collection<GridCacheSqlQuery> queries,
-            Object[] params,
-            Throwable error
+        long reqId,
+        String label,
+        String schemaName,
+        Collection<GridCacheSqlQuery> queries,
+        Object[] params,
+        Throwable error
     ) {
         StringBuilder logMessage = new StringBuilder();
 
         logMessage.append("Query Execution Failed:")
-                .append("\nRequest ID: ").append(reqId)
-                .append("\nLabel: ").append(label != null ? label : "N/A")
-                .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
-                .append("\nQueries: ").append(
-                        queries != null
-                                ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
-                                : "N/A")
-                .append("\nParameters: ").append(params != null ? Arrays.toString(params) : "N/A");
+            .append("\nRequest ID: ").append(reqId)
+            .append("\nLabel: ").append(label != null ? label : "N/A")
+            .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
+            .append("\nQueries: ").append(
+                queries != null
+                    ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
+                    : "N/A")
+            .append("\nParameters: ").append(params != null ? Arrays.toString(params) : "N/A");
 
         if (error != null) {
             logMessage.append("\nError: ").append(error.getMessage());
         }
 
+        // Log the error with the constructed message
         log.error(logMessage.toString(), error);
     }
 
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -219,7 +219,6 @@ public class GridMapQueryExecutor {
         }
 
         // Log the error with the constructed message
-        System.out.println("🔥 ACTUAL LOGGER CALL: " + logMessage);// Jan 31 - Debugging statement
         log.error(logMessage.toString(), error);
 
     }
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -224,9 +224,6 @@ public class GridMapQueryExecutor {
             logMessage.append(secureParams ? "HIDDEN" : Arrays.toString(params));
         }
 
-        if (error != null)
-            logMessage.append("\nError: ");
-
         return logMessage.toString();
 
     }
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -190,7 +190,7 @@ public class GridMapQueryExecutor {
      * @param params Query parameters, if any.
      * @param error Exception that occurred during the query execution.
      */
-    protected void logQueryDetails(
+    String buildQueryLogDetails(
         long reqId,
         String label,
         String schemaName,
@@ -218,8 +218,8 @@ public class GridMapQueryExecutor {
             logMessage.append("\nError: ").append(error.getMessage());
         }
 
-        // Log the error with the constructed message
-        log.error(logMessage.toString(), error);
+        // Return the constructed message
+        return logMessage.toString();
 
     }
 
@@ -625,16 +625,15 @@ public class GridMapQueryExecutor {
                         if (qryRetryErr != null)
                             sendError(node, reqId, qryRetryErr);
                         else {
+                            String errMsg = buildQueryLogDetails(reqId, label, schemaName, qrys, params, e, node.id());
+
                             if (e instanceof Error) {
-                                // Log detailed query information for debugging.
-                                logQueryDetails(reqId, label, schemaName, qrys, params, e, node.id());
+                                log.error(errMsg, e);
                                 throw (Error)e;
                             } else {
-                                //Log only once for recoverable errors.
-                                logQueryDetails(reqId, label, schemaName, qrys, params, e,node.id());
+                                log.error(errMsg, e); // You can skip or keep this line based on verbosity level
                                 U.warn(log, "Failed to execute local query.", e);
                             }
-                            sendError(node, reqId, e);
                         }
                     }
                 }
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -218,11 +218,9 @@ public class GridMapQueryExecutor {
             logMessage.append("\nError: ").append(error.getMessage());
         }
 
-        // Return the constructed message
         return logMessage.toString();
 
     }
-
 
     /**
      * @param nodeId Node ID.
--- a/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
+++ b/modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
@@ -191,28 +191,28 @@ public class GridMapQueryExecutor {
      * @param error Exception that occurred during the query execution.
      */
     String buildQueryLogDetails(
-        long reqId,
-        String label,
-        String schemaName,
-        Collection<GridCacheSqlQuery> queries,
-        Object[] params,
-        Throwable error,
-        UUID remoteNodeId
+            long reqId,
+            String label,
+            String schemaName,
+            Collection<GridCacheSqlQuery> queries,
+            Object[] params,
+            Throwable error,
+            UUID remoteNodeId
     ) {
         StringBuilder logMessage = new StringBuilder();
 
         logMessage.append("Query Execution Failed:")
-            .append("\nRequest ID: ").append(reqId)
-            .append("\nLabel: ").append(label != null ? label : "N/A")
-            .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
-            .append("\nQueries: ").append(
-                queries != null
-                    ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
-                    : "N/A")
-            .append("\nParameters: ").append(params != null ? Arrays.toString(params) : "N/A")
-            .append("\nTimestamp: ").append(System.currentTimeMillis())
-            .append("\nLocal Node ID: ").append(ctx.localNodeId())
-            .append("\nRemote Node ID: ").append(remoteNodeId);
+                .append("\nRequest ID: ").append(reqId)
+                .append("\nLabel: ").append(label != null ? label : "N/A")
+                .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
+                .append("\nQueries: ").append(
+                        queries != null
+                                ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
+                                : "N/A")
+                .append("\nParameters: ").append(params != null ? Arrays.toString(params) : "N/A")
+                .append("\nTimestamp: ").append(System.currentTimeMillis())
+                .append("\nLocal Node ID: ").append(ctx.localNodeId())
+                .append("\nRemote Node ID: ").append(remoteNodeId);
 
         if (error != null) {
             logMessage.append("\nError: ").append(error.getMessage());
@@ -589,15 +589,15 @@ public class GridMapQueryExecutor {
         catch (Throwable e) {
             if (qryResults != null) {
                 nodeRess.remove(reqId, segmentId, qryResults);
+
                 qryResults.close();
 
                 // If a query is cancelled before execution is started partitions have to be released.
                 if (!lazy || !qryResults.isAllClosed())
                     qryResults.releaseQueryContext();
             }
-            else {
+            else
                 releaseReservations(qctx);
-            }
 
             if (e instanceof QueryCancelledException)
                 sendError(node, reqId, e);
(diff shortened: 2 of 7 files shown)


dy2285:
Hi Ilya @solveme 
Please review only the last commit (07e01a2).The previous commits were part of the development process and should not be considered separately.

Summary of Changes:
- Implemented `logQueryDetails` method for centralized error logging.
- Integrated `logQueryDetails` into query execution error handling.
- Addressed previous feedback by removing redundant logging.
- Added an integration test to verify logged messages.
- Successfully tested in both local IDE and TeamCity.

Notes:
- Branch blocker VISA issue unresolved (see Slack discussion)

checkstyle/checkstyle.xml
```diff
@@ -23,6 +23,11 @@
     <property name="charset" value="UTF-8"/>
 
     <property name="fileExtensions" value="java, properties, xml"/>
+    
+    <module name="SuppressionFilter">
```
stasusov, line 27: Why did you add this? What's the reason or issue you were fixing with this?

dy2285: This was unintentionally added while troubleshooting local checkstyle violations. It’s been reverted since — not part of the final change.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -176,6 +176,48 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+
+        logMessage.append("Query Execution Failed:")
+                .append("\nRequest ID: ").append(reqId)
+                .append("\nLabel: ").append(label != null ? label : "N/A")
+                .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
+                .append("\nQueries: ").append(
+                        queries != null
+                                ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
+                                : "N/A")
+                .append("\nParameters: ").append(params != null ? Arrays.toString(params) : "N/A")
```
solveme, line 208: I've just realized that dumping query arguments might not be desirable behavior (from security perspective). If user want to see params he should explicitly allow it.

We can use same approach that is used in `BinaryObjectImpl.toString()`, i.e.:

    if (params == null)
        .append("\nParameters: N/A")
    else {
        boolean secureParams = S.getSensitiveDataLogging() != PLAIN;
        .append("\nParameters: ")
        .append(secureParams ? "HIDDEN" : Arrays.toString(params))
    }

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -578,15 +620,18 @@ private void onQueryRequest0(
                         if (qryRetryErr != null)
                             sendError(node, reqId, qryRetryErr);
                         else {
+                            String errMsg = buildQueryLogDetails(reqId, label, schemaName, qrys, params, e, node.id());
+
                             if (e instanceof Error) {
-                                U.error(log, "Failed to execute local query.", e);
+                                log.error(errMsg, e);
```
solveme, line 626: Let's use construction like:

    U.error(log, errMsg, e);

To guarantee that we won't lose this message in case if logging was not properly configured by user

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+        String logMsg = executor.buildQueryLogDetails(
+            1L, "TestLabel", "TestSchema", queries, new Object[]{}, e, nodeId
+        );
+
+        assertTrue(logMsg.contains("Query Execution Failed"));
+        assertTrue(logMsg.contains("SELECT * FROM test"));
+        assertTrue(logMsg.contains("Error: Forced Failure"));
+    }
+}
```
solveme, line 167: Add an empty line at the end

solveme (changes requested):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Replaced the original `params` appending line with the suggested conditional block that checks `S.getSensitiveDataLogging()`. Thanks for highlighting the security consideration!

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -176,6 +176,48 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+        logMessage.append("Query Execution Failed:")
+                .append("\nRequest ID: ").append(reqId)
+                .append("\nLabel: ").append(label != null ? label : "N/A")
+                .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
+                .append("\nQueries: ").append(
+                        queries != null
+                                ? queries.stream().map(GridCacheSqlQuery::query).collect(Collectors.joining("; "))
+                                : "N/A")
+                .append("\nParameters: ").append(params != null ? Arrays.toString(params) : "N/A")
+                .append("\nTimestamp: ").append(System.currentTimeMillis())
```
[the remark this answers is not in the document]
dy2285: Removed the timestamp line — makes sense to rely on the logger’s built-in timestamp. Thanks!

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Updated both logging branches to use `U.error(log, errMsg, e)` as suggested — ensures message won’t be lost due to misconfigured logger.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,167 @@
...
+
+import java.sql.SQLException;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.UUID;
+import java.lang.reflect.Field;
+import static org.junit.Assert.assertTrue;
+import static org.mockito.Mockito.*;
+
+public class GridMapQueryExecutorTest {
```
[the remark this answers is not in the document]
dy2285: Refactored the test class to remove unnecessary mocks and dependency wiring. Since `buildQueryLogDetails()` is now a pure function,  I've simplified the setup and made all tests standalone.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
dy2285: Added the trailing newline as part of the test class cleanup. Thanks for the catch!

solveme (commented):
LGTM

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -582,15 +633,18 @@ private void onQueryRequest0(
                         if (qryRetryErr != null)
                             sendError(node, reqId, qryRetryErr);
                         else {
+                            String errMsg = buildQueryLogDetails(reqId, label, schemaName, qrys, params, e, node.id(),ctx.localNodeId());
```
ibelyakov, line 636: There should be a space between "node.id()," and "ctx.localNodeId()". Also, the line seems to be pretty long, please consider splitting it into two.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -337,7 +388,7 @@ public void onQueryRequest(final ClusterNode node, final GridH2QueryRequest req)
      * @param maxMem Query memory limit.
      * @param runningQryId Running query id.
      */
-    private void onQueryRequest0(
+    protected void onQueryRequest0(
```
ibelyakov, line 391: I believe we can keep it "private" since there are no usages for this method by any other class besides GridMapQueryExecutor.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,106 @@
+/*
+ * Copyright 2019 GridGain Systems, Inc. and Contributors.
```
ibelyakov, line 2: Should be 2025

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +183,54 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+            UUID localNodeId
+    ) {
+        StringBuilder logMessage = new StringBuilder();
+
+        logMessage.append("Query Execution Failed:")
+                .append("\nRequest ID: ").append(reqId)
+                .append("\nLabel: ").append(label != null ? label : "N/A")
+                .append("\nSchema: ").append(schemaName != null ? schemaName : "N/A")
+                .append("\nQueries: ").append(
+                        queries != null
```
ibelyakov, line 213: I suppose that the `queries` collection could be empty, so let's use `!F.isEmpty(queries)` here instead

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +183,54 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+            boolean secureParams = S.getSensitiveDataLogging() != PLAIN;
+            logMessage.append("\nParameters: ");
+            logMessage.append(secureParams ? "HIDDEN" : Arrays.toString(params));
+        }
+
+        if (error != null)
+            logMessage.append("\nError: ").append(error.getMessage());
+
+        return logMessage.toString();
+
```
ibelyakov, line 231: Let's remove the empty string.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -582,15 +633,18 @@ private void onQueryRequest0(
...
                         else {
+                            String errMsg = buildQueryLogDetails(reqId, label, schemaName, qrys, params, e, node.id(),ctx.localNodeId());
+
                             if (e instanceof Error) {
-                                U.error(log, "Failed to execute local query.", e);
+                                U.error(log, errMsg, e);
 
                                 throw (Error)e;
                             }
+                            else {
```
ibelyakov, line 643: There is no point in adding "else" here. Also, let's change the logging level back to WARN here, since it was intentionally downgraded sometime ago as part of GG-34383.

ibelyakov (changes requested):
See comments.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +183,54 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+        if (params == null)
+            logMessage.append("\nParameters: N/A");
+        else {
+            boolean secureParams = S.getSensitiveDataLogging() != PLAIN;
+            logMessage.append("\nParameters: ");
+            logMessage.append(secureParams ? "HIDDEN" : Arrays.toString(params));
+        }
+
+        if (error != null)
+            logMessage.append("\nError: ").append(error.getMessage());
```
ibelyakov, line 228: Do we need to add error.getMessage() into the message here? Won't it be printed in the log anyway as a part of U.error(log, errMsg, e); execution?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Thanks! Switched to !F.isEmpty(queries) as suggested. Much cleaner and safer.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Makes sense — removed the explicit getMessage() call since it's already included in U.error(...). Thanks!

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Removed the dangling empty string after dropping getMessage() — thanks for pointing that out.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Reverted to `private` — It was changed to protected earlier when this was extracted as a method. After inlining back to a function inside the same class, `private` is sufficient again.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
dy2285: Added a space and split long line for clarity as suggested.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
dy2285: Updated year to 2025.

ibelyakov (approved):
LGTM

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +183,54 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+        if (params == null)
+            logMessage.append("\nParameters: N/A");
+        else {
+            boolean secureParams = S.getSensitiveDataLogging() != PLAIN;
+            logMessage.append("\nParameters: ");
+            logMessage.append(secureParams ? "HIDDEN" : Arrays.toString(params));
+        }
+
+        if (error != null)
+            logMessage.append("\nError: ");
```
ashapkin, line 228: Something is broken here. Why do we need "error" obj in this method? 
Do tests pass? It looks like here we are expecting to pring the error message.

dy2285: This if (error != null) block was originally meant to include error.getMessage(), but we removed that earlier due to redundancy (since the full throwable is logged right after). Since you pointed it out, I re-evaluated and agree that logging just the label without content isn’t useful. For clean and clear logging, it's better to rely on the logging framework’s built-in exception handling — so I’ve removed the block entirely for clarity.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +184,66 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+    ) {
+        SensitiveDataLogging sensitivity = S.getSensitiveDataLogging();
+
+        String queriesStr = F.isEmpty(queries)
+                ? "N/A"
+                : queries.stream()
+                .map(GridCacheSqlQuery::query)
+                .collect(Collectors.joining("; "));
+
+        String paramsStr;
```
solveme, line 215: Let's extract paramString resolution to separate utility method (i.e. `String stringifyParams(Object[] params)`). This will help us to get rid of `paramsStr` variable because use it only once in final formatting. So this method can be a pure function like:

    String stringifyParams(Object[] params) {
        if (params == null || params.length == 0)
            return "N/A";    // It's safe to tell that there were no params

        switch (S.getSensitiveDataLogging()) {
            case NONE:
                return "HIDDEN";
            case HASH:
                return String.valueOf(IgniteUtils.hash(Arrays.hashCode(params)));
            case PLAIN:
            default:
                return Arrays.toString(params);
        }    
    }

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +184,66 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+            long reqId,
+            String label,
+            String schemaName,
+            Collection<GridCacheSqlQuery> queries,
+            Object[] params,
+            Throwable error,
+            UUID remoteNodeId,
+            UUID localNodeId
+    ) {
+        SensitiveDataLogging sensitivity = S.getSensitiveDataLogging();
```
solveme, line 207: Let's move it closer to the place where it used. Probably we can even inline it, see my post about separate method below.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +184,66 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+                .map(GridCacheSqlQuery::query)
+                .collect(Collectors.joining("; "));
+
+        String paramsStr;
+        if (params == null || params.length == 0) {
+            paramsStr = "params=HIDDEN";
+        } else {
+            switch (sensitivity) {
+                case NONE:
+                    paramsStr = "params=HIDDEN";
```
solveme, line 221: We apready have `" ... params=%s ..."` in the final formatting, no need to add `params=` here

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +184,66 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+                    break;
+                case HASH:
+                    paramsStr = "params=" + Arrays.stream(params)
```
solveme, line 224: We can't hash each element because if has code of some types like integers is the number itself, so I can try to infer original param value from `Integer.toHexString(Objects.hashCode(p))` especially if it's clear from the query string that iven column is integer || short || byte. 

So what we should do is to calculate the hash of the whole params array. See my previous comment with sample. As we discussed the goal of having hash code is to evaluate whether given query always fail with same params or not. So proposed technique is enough for that.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutor.java
```diff
@@ -180,6 +184,66 @@ public void onCancel(ClusterNode node, GridQueryCancelRequest msg) {
...
+                            .map(p -> Integer.toHexString(Objects.hashCode(p)))
+                            .collect(Collectors.toList());
+                    break;
+                case PLAIN:
+                default:
+                    paramsStr = "params=" + Arrays.toString(params);
```
solveme, line 230: Again redundant "params="

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,118 @@
...
+                null,
+                remoteNodeId,
+                localNodeId
+        );
+
+        assertTrue(logContent.contains("reqId=123"));
+        assertTrue(logContent.contains("label=N/A"));
+        assertTrue(logContent.contains("schema=N/A"));
+        assertTrue(logContent.contains("queries=N/A"));
+        assertTrue(logContent.contains("params=params=HIDDEN"));
```
solveme, line 85: `params=params=HIDDEN` :smile:

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,118 @@
...
+        );
+
+        assertTrue(logMsg.contains("reqId=1"));
+        assertTrue(logMsg.contains("label=TestLabel"));
+        assertTrue(logMsg.contains("schema=TestSchema"));
+        assertTrue(logMsg.contains("queries=SELECT * FROM test"));
+        assertTrue(logMsg.contains("localNodeId=" + localNodeId));
+        assertTrue(logMsg.contains("remoteNodeId=" + nodeId));
+        assertTrue(logMsg.contains("params=HIDDEN"));
+    }
```
solveme, line 117: Let's add one more tests where params array has some content. Let's say `(int) 42, null, "foo", (double)3.14`, etc..

solveme (changes requested):
@dy2285 let's extract params stringification ot separate method in order to simplify the code a bit and to get rid of `"params="` duplications.

Also let's add one more test that I mentioned.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,117 @@
...
+                "TestSchema",
+                queries,
+                params,
+                new RuntimeException("Test exception"),
+                UUID.randomUUID(),
+                UUID.randomUUID()
+        );
+
+        // Depending on system config, assert format
+        assertTrue(log.contains("params="));
```
solveme, line 93: The goal of this test was to check that we properly dump params depending on sensitivity value, so best way is to   wrap body of this ticket to separate method which accepts Sensitivity value and expected params ie.. 

    private static testBuildQueryLogDetailsWithSampleParams(String sensitivity, String expectedParams) {
        System.setProperty(IgniteSystemProperties.IGNITE_SENSITIVE_DATA_LOGGING, sensitivity);
        
        Collection<GridCacheSqlQuery> queries = Collections.singletonList(
            new GridCacheSqlQuery("SELECT * FROM test")
        );
        
        ....
        
        assertTrue(log.contains("params=" + expectedParams));
    }

Then you can reuse it for testing all sensitivity cases, i.e.:

    @Test
    public testBuildQueryLogDetailsWithPlainSensitivity(String sensitivity, String expectedParams) {
        testBuildQueryLogDetailsWithSampleParams("plain", "[...]")
    }

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,117 @@
...
+
+import java.util.Collection;
+import java.util.Collections;
+import java.util.UUID;
+
+import static org.junit.Assert.assertTrue;
+
+public class GridMapQueryExecutorTest {
+
+    private final GridMapQueryExecutor executor = new GridMapQueryExecutor();
```
solveme, line 31: It looks like this variable is not used anywhere, if that so, let's remove it

solveme (changes requested):
Left a comment about test adjustments

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,90 @@
...
+import java.util.Arrays;
+import java.util.Collection;
+import java.util.Collections;
+import java.util.UUID;
+
+import static org.junit.Assert.assertTrue;
+
+public class GridMapQueryExecutorTest {
+
+    public void testBuildQueryLogDetailsWithSampleParams(String sensitivity, String expectedParams) {
```
solveme, line 34: It's better to not hardcode params value in the method body, but rather accept it as parameter, because it will make things much more obvious at the callsite. I.e. we pass following params and we expect to see following result.

P.S. I remember that It was me, who suggested such signature. Sorry if I confused you, but don't threat my code samples as perfect :smile: by default.

modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/h2/twostep/GridMapQueryExecutorTest.java
```diff
@@ -0,0 +1,90 @@
...
+    }
+
+    @Test
+    public void testBuildQueryLogDetailsWithPlainSensitivity() {
+        testBuildQueryLogDetailsWithSampleParams("plain", "[42, john, 345]");
+    }
+
+    @Test
+    public void testBuildQueryLogDetailsWithHashSensitivity() {
+        int hash = Arrays.hashCode(new Object[]{42, "john", 345});
```
solveme, line 81: As was highlighted above, it's better to have something like

    Object[] givenParams = new Object[]{42, "john", 345};
    String expectedOutput = String.valueOf(Arrays.hashCode(givenParams));
    testBuildQueryLogDetailsWithSampleParams("hash", givenParams, expectedOutput);

Please adjust sensitivity-related methods in this way.

solveme (changes requested):

solveme (approved):
LGTM

PR #3361, opened 2024-12-16
GG-41429: Added support for returning score field for VectorQuery

manini-zs:
Added support for similarity function and returning scores.
This is sub optimally implemented using Maps.
Based on feedback from the python side, will use either T3 or the new IgniteBiVectorTuple eventually


modules/core/src/main/java/org/apache/ignite/cache/query/VectorQuery.java
```diff
@@ -98,11 +102,13 @@ public VectorQuery(Class<?> type, String field, float[] clauseVector, int k) {
...
      * @param clauseVector Search string.
      */
-    public VectorQuery(String type, String field, float[] clauseVector, int k) {
+    public VectorQuery(String type, String field, float[] clauseVector, int k, float threshold) {
         setType(type);
         setField(field);
         setClauseVector(clauseVector);
         setK(k);
+        setThreshold(threshold);
+
```
PakhomovAlexander, line 111: empty line could be removed

modules/core/src/main/java/org/apache/ignite/cache/query/VectorQuery.java
```diff
@@ -159,6 +165,27 @@ public int getK() {
         return k;
     }
 
+    /**
+     * Sets k (the number of vectors to return).
```
PakhomovAlexander, line 169: The javadoc does not reflect the method.

modules/core/src/main/java/org/apache/ignite/cache/query/VectorQuery.java
```diff
@@ -159,6 +165,27 @@ public int getK() {
...
+
+        return this;
+    }
+
+    /**
+     * Gets k (the number of vectors to return).
+     *
+     * @return K.
+     */
+    public float getThreshold() {
```
PakhomovAlexander, line 185: Same for javadoc here.

PakhomovAlexander (changes requested):

modules/core/src/main/java/org/apache/ignite/cache/query/VectorQuery.java
manini-zs: Fixed

modules/core/src/main/java/org/apache/ignite/cache/query/VectorQuery.java
manini-zs: Fixed

manini-zs:
I have removed the similarity function for now, will add it as a part of a new improvement.
Just retained the score feature. Please review and let me know.

PakhomovAlexander (approved):

manini-zs:
This feature has been moved to a new ticket and the corresponding branch

PR #3362, opened 2024-12-18
GG-40982 Plugin information added to metrics message

ibelyakov:


modules/core/src/main/java/org/apache/ignite/internal/IgniteKernal.java
```diff
@@ -2308,6 +2309,11 @@ private void ackNodeMetrics(DecimalFormat dblFmt,
                     msg.nl().a("    ^-- ").a(createExecutorDescription(entry.getKey(), entry.getValue()));
             }
 
+            String pluginInfo = pluginInfoReport();
+
+            if (!F.isEmpty(pluginInfo))
+                msg.a(pluginInfo);
```
vmalin, line 2315: In this implementation we write nothing for the community edition. Let's write it to the log in case no pluginInfo was collected. And also add it to org.apache.ignite.internal.GridNodeMetricsLogSelfTest#testNodeMetricsLog. This test is currently designed in a way that it doesn't fail if new information is added to the metrics message so it's easy to forget about it.

PR #3368, opened 2025-01-09
GG-41725 Update the reference output provided by the control utility with the 2025 year

solveme:


solveme:
VISA acquired (see ticket)

PR #3381, opened 2025-02-07
GG-41610 Improve partition eviction logs

ibessonov:
https://ggsystems.atlassian.net/browse/GG-41610

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionsEvictManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionsEvictManager.java
@@ -26,6 +26,7 @@ import java.util.StringJoiner;
 import java.util.concurrent.ConcurrentHashMap;
 import java.util.concurrent.ConcurrentLinkedDeque;
 import java.util.concurrent.ConcurrentMap;
+import java.util.concurrent.atomic.AtomicBoolean;
 import java.util.concurrent.atomic.AtomicInteger;
 import java.util.concurrent.atomic.AtomicReference;
 import java.util.concurrent.locks.ReadWriteLock;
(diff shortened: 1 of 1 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionsEvictManager.java
```diff
@@ -657,13 +664,20 @@ private PartitionEvictionTask(
                 return;
             }
 
-            BooleanSupplier stopClo = () -> grpEvictionCtx.shouldStop() || (state.get() == Boolean.FALSE);
+            BooleanSupplier stopClo = () -> {
+                boolean stop = grpEvictionCtx.shouldStop() || (state.get() == Boolean.FALSE);
+
+                if (!stop)
+                    showProgress();
```
rpuch, line 671: It seems a bit esoteric to use a 'whether we should stop' closure to print progress. Can an explicit abstraction like `ProgressListener` be used here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionsEvictManager.java
```diff
@@ -241,10 +245,11 @@ public PartitionEvictionTask scheduleEviction(
             return task;
         }
 
-        if (log.isDebugEnabled())
-            log.debug("The partition has been scheduled for clearing [grp=" + grp.cacheOrGroupName()
+        if (log.isInfoEnabled()) {
```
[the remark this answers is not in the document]
ibessonov: Yes. It is supposed to be judged by a number of lines, not statements, as far as I know. Of course, I may be wrong, but bracing multi-lined statements makes more sense to me

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/distributed/dht/topology/PartitionsEvictManager.java
ibessonov: It is possible, but I felt like it would add unnecessary noise to the `clearAll` call

rpuch (approved):

PR #3391, opened 2025-02-24
GG 41675 Added vector similarity function (optional) & score functionality

manini-zs:
…ard compatibility)


modules/core/src/main/java/org/apache/ignite/cache/QueryIndex.java
```diff
@@ -295,6 +310,43 @@ public QueryIndex setInlineSize(int inlineSize) {
         return this;
     }
 
+    /**
+     * Sets the Vector Similarity Function for VECTOR Index based on the int value received.
+     *
+     * @param similarityFunctionInt Vector Similarity Function as integer.
+     * @return {@code this} for chaining.
+     */
+    public QueryIndex setSimilarityFunction(int similarityFunctionInt) {
```
PakhomovAlexander, line 319: Im not sure we really need this function in our public API. Why should the user know integer representation of `SimilarityFunction`? I propose removing this method from public API.

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ClientUtils.java
```diff
@@ -526,7 +526,13 @@ ClientCacheConfiguration cacheConfiguration(BinaryInputStream in, ProtocolContex
                                             LinkedHashMap::new
                                     ));
 
-                                    return new QueryIndex(fields, type).setName(name).setInlineSize(inlineSize);
+                                    QueryIndex queryIndex = new QueryIndex(fields, type).setName(name).setInlineSize(inlineSize);
+                                    if(protocolCtx.isFeatureSupported(ProtocolBitmaskFeature.QUERY_INDEX_VECTOR_SIMILARITY)){
```
PakhomovAlexander, line 530: Apply formatter here, please.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
```diff
@@ -1325,8 +1326,18 @@ protected void runQuery(GridCacheQueryInfo qryInfo) {
...
                                 continue;
                         }
-                        else
-                            data.add(new T2<>(key, val));
+                        else {
+
+                            if (type == VECTOR && row instanceof IgniteBiVectorTuple) {
+                                Double score = ((IgniteBiVectorTuple<K, V>) row).getScore();
+                                Map<Object, Double> value = new HashMap<>();
+                                value.put(val,score);
```
PakhomovAlexander, line 1334: Could you configure an IDE to apply formatter to changed lines automatically?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheConfigurationSerializer.java
```diff
@@ -559,8 +559,16 @@ public static QueryEntity readQueryEntity(BinaryRawReader in, ClientProtocolCont
         if (cnt > 0) {
             Collection<QueryIndex> indexes = new ArrayList<>(cnt);
 
-            for (int i = 0; i < cnt; i++)
-                indexes.add(PlatformConfigurationUtils.readQueryIndex(in));
+            for (int i = 0; i < cnt; i++) {
+                QueryIndex queryIndex = null;
+                //check if similarity function feature is supported
```
PakhomovAlexander, line 564:
```suggestion
```

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/utils/PlatformConfigurationUtils.java
```diff
@@ -689,11 +689,25 @@ public static QueryEntity readQueryEntity(BinaryRawReader in) {
...
+     * @param hasVectorSimilarity whether the similarity function is defined.
+     * @return Query index.
+     */
+    public static QueryIndex readQueryIndex(BinaryRawReader in, boolean hasVectorSimilarity) {
         QueryIndex res = new QueryIndex();
 
         res.setName(in.readString());
         res.setIndexType(QueryIndexType.values()[in.readByte()]);
         res.setInlineSize(in.readInt());
+        if(hasVectorSimilarity){
```
PakhomovAlexander, line 708: formatter

modules/core/src/main/java/org/apache/ignite/internal/processors/query/QueryIndexDescriptorImpl.java
```diff
@@ -59,6 +60,11 @@ public class QueryIndexDescriptorImpl implements GridQueryIndexDescriptor {
     /** */
     private final int inlineSize;
 
+    /**
+     * Vector Similarity Function for VECTOR index.
+     * */
```
PakhomovAlexander, line 65:
```suggestion
     */
```

modules/core/src/main/java/org/apache/ignite/cache/QueryIndex.java
manini-zs: This is needed by the PlatformConfigurationUtils and ClientUtils
They both need to set the Similarity Function as Integer

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/query/GridCacheQueryManager.java
manini-zs: I will do that

modules/core/src/main/java/org/apache/ignite/lang/IgniteBiVectorTuple.java
```diff
@@ -0,0 +1,45 @@
...
+        return score;
+    }
+
+    /**
+     * Sets score.
+     *
+     * @param score Score.
+     */
+    public void setScore(double score) {
+        this.score = score;
```
ygerzhedovich, line 42: Do we have cases when the`score` can be changed? Can it be a final field?

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheConfigurationSerializer.java
```diff
@@ -559,8 +559,16 @@ public static QueryEntity readQueryEntity(BinaryRawReader in, ClientProtocolCont
...
+                if (protocolCtx.isFeatureSupported(ClientBitmaskFeature.QUERY_INDEX_VECTOR_SIMILARITY)) {
```
ygerzhedovich, line 565: It will be better to introduce a new variable like a `vectorSearchSupported` and pass the value as a parameter to the utility method

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheQueryCursor.java
```diff
@@ -84,6 +88,17 @@ void writePage(BinaryRawWriterEx writer) {
         while (cnt < pageSize && iter.hasNext()) {
             T e = iter.next();
 
+            boolean isScoreSupported = getProtocolContext().isFeatureSupported(ClientBitmaskFeature.QUERY_INDEX_VECTOR_SIMILARITY);
+            if (!isScoreSupported && e instanceof Cache.Entry
```
ygerzhedovich, line 92: it looks wrong. You try to check that value is double (seems score) for the case when the score is not supported.

modules/core/src/main/java/org/apache/ignite/cache/QueryIndex.java
ygerzhedovich: Totally agree with @PakhomovAlexander . It looks incorrect to have it as part of public API. Let's think about how it can be avoided.

modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QueryVectorField.java
```diff
@@ -33,5 +33,17 @@
...
 public @interface QueryVectorField {
-    // No-op.
+    /**
+     * Specifies similarityFunction for a vector field.
+     *
+     * @return similarityFunction for a vector field.
+     */
+    SimilarityFunction similarityFunction() default SimilarityFunction.COSINE;
+
+    public static enum SimilarityFunction {
```
ygerzhedovich, line 43: I think we should have constant id's for all values to prevent issues in the future when somebody decides to resort it or add a new type into the middle of the list.

ygerzhedovich (changes requested):
All changes must be covered by tests

modules/core/src/main/java/org/apache/ignite/cache/QueryIndex.java
manini-zs: I need to set the similarity function that I receive as an integer from the python thin client here via ClientUtils.
In ClientUtils the configuration received from the client is processed and the object of the QueryIndex is created, so it made sense to put this here.
If there is a better way to do it, I'd be happy to improve this.

modules/core/src/main/java/org/apache/ignite/cache/query/VectorQuery.java
```diff
@@ -210,7 +210,7 @@ public VectorQuery<K, V> setThreshold(float threshold) {
     }
 
     /**
-     * Gets k (the number of vectors to return).
+     * Gets threshold for cosine similarity.
      *
      * @return K.
```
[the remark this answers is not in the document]
manini-zs: I was correcting the mistake in the previous java doc for threshold here, and missed the return type, I will correct this

modules/core/src/main/java/org/apache/ignite/internal/client/thin/ProtocolBitmaskFeature.java
```diff
@@ -76,7 +76,10 @@ public enum ProtocolBitmaskFeature {
     QRY_LABEL(16),
 
     /** Cache plugin configurations. GG-specific, use higher id to avoid conflicts. */
-    CACHE_PLUGIN_CONFIGURATIONS(32);
+    CACHE_PLUGIN_CONFIGURATIONS(32),
+
+    /** Vector Similarty function for VECTOR INDEX. */
+    QUERY_INDEX_VECTOR_SIMILARITY(47);
```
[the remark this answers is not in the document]
manini-zs: It can very well be 33.
I tried to keep a gap similar to the one between QRY_LABEL and QUERY_INDEX_VECTOR_SIMILARITY
/** Custom Query label. */
QRY_LABEL(16),

/** Cache plugin configurations. GG-specific, use higher id to avoid conflicts. */
CACHE_PLUGIN_CONFIGURATIONS(32),

/** Vector Similarty function for VECTOR INDEX. */
QUERY_INDEX_VECTOR_SIMILARITY(47);

modules/core/src/main/java/org/apache/ignite/lang/IgniteBiVectorTuple.java
manini-zs: No it does not change once queried, unless the data is somehow changed. It can be made final.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheConfigurationSerializer.java
manini-zs: You're right, it looks cleaner that way. I will do it.

modules/core/src/main/java/org/apache/ignite/internal/processors/platform/client/cache/ClientCacheQueryCursor.java
manini-zs: By default the score is always sent in the results for the vector search.
I am supposed to remove the score value from the data being sent to the python thin client, if the score is not supported in the version of the python thin client with vector search (pygridgain 1.5.0) for backward compatibility

modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QueryVectorField.java
manini-zs: Sure, will do that

modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QueryVectorField.java
```diff
@@ -33,5 +33,17 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.TYPE})
 public @interface QueryVectorField {
-    // No-op.
+    /**
+     * Specifies similarityFunction for a vector field.
```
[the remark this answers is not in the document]
manini-zs: returns. I will fix this.

manini-zs:
This pull request is on the older branch. A new pull request has been created and approved in its place.
https://github.com/gridgain/gridgain/pull/3428

PR #3394, opened 2025-02-27
GG-42378 Index reader - Additional B plus tree check [triangle invariant]

vmalin:


modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -574,6 +581,9 @@ private Map<Integer, List<Throwable>> checkParts(Map<String, TreeTraversalInfo>
 
         ProgressPrinter progressPrinter = new ProgressPrinter(System.out, "Checking partitions", partCnt);
 
+        // Map<partNum, partItemsCnt>
```
ibessonov, line 584: I'd rather see something like `{partNum -> partItemCnt}` in the comment. Type parameters should contain types, and their alternative usage looks inappropriate

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -638,6 +650,10 @@ private Map<Integer, List<Throwable>> checkParts(Map<String, TreeTraversalInfo>
                 res.put(partId, errors);
         }
 
+        long totalItemsCnt = itemsCntMap.values().stream().mapToLong(Long::longValue).sum();
+
+        print("Cache items number: " + totalItemsCnt);
```
ibessonov, line 655: Why do we need it?

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -1219,6 +1237,42 @@ private TreeNode getTreeNode(long pageId, TreeTraverseContext nodeCtx) {
...
     }
 
+    /**
+     * @param nodeCtx tree traverse context.
+     * @param pageContent inner page content.
+     * @param io inner page io.
+     * @throws IgniteCheckedException
+     */
+    private void checkTriangleInvariant(TreeTraverseContext nodeCtx, PageContent pageContent,
+        PageIO io) throws IgniteCheckedException {
```
ibessonov, line 1247: Formatting looks off

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -1219,6 +1237,42 @@ private TreeNode getTreeNode(long pageId, TreeTraverseContext nodeCtx) {
...
+        if (io instanceof BPlusInnerIO && pageContent.linkedPageIds != null) {
+            long rightChildPageId = -1;
+
+            // linkedPageIds are populated in InnerPageIOProcessor.getContent
+            for (Long linkedPageId: pageContent.linkedPageIds) {
```
ibessonov, line 1252:
```suggestion
            for (Long linkedPageId : pageContent.linkedPageIds) {
```

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -1219,6 +1237,42 @@ private TreeNode getTreeNode(long pageId, TreeTraverseContext nodeCtx) {
...
+                triangleChecksNum++;
+
+                if (rightChildPageId != -1 && linkedPageId != rightChildPageId)
+                    triangleErrorsCounter++;
+
+                final ByteBuffer innerBuf = allocateBuffer(pageSize);
```
ibessonov, line 1258: I think it's too wasteful to allocate and deallocate buffer for every linked page, please reuse it where possible

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -196,6 +196,10 @@ public class IgniteIndexReader implements AutoCloseable {
     /** */
     private final PageIOProcessor metaPageIOProcessor = new MetaPageIOProcessor();
 
+    private static long triangleErrorsCounter;
+
+    private static long triangleChecksNum;
```
ibessonov, line 201: Why are these fields static? Looks like a mistake, please

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
vmalin: fixed

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
vmalin: Fixed

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
vmalin: Fixed

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
vmalin: Fixed

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -1219,6 +1241,44 @@ private TreeNode getTreeNode(long pageId, TreeTraverseContext nodeCtx) {
...
+
+            try {
+                // linkedPageIds are populated in InnerPageIOProcessor.getContent
+                for (Long linkedPageId : pageContent.linkedPageIds) {
+                    triangleChecksNum++;
+
+                    innerBuf.rewind();
+
+                    if (rightChildPageId != -1 && linkedPageId != rightChildPageId)
+                        triangleErrorsCounter++;
```
ibessonov, line 1265: Would it make sense to preserve these `pageId`s and print them, like we do with orphans?

ibessonov (commented):
I don't see a `--data-consistency-check` option, as proposed in the description

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
```diff
@@ -1219,6 +1241,44 @@ private TreeNode getTreeNode(long pageId, TreeTraverseContext nodeCtx) {
...
+
+                    final long childAddr = bufferAddress(innerBuf);
+
+                    final PageIO childIO = PageIO.getPageIO(childAddr);
+
+                    rightChildPageId = ((BPlusIO)childIO).getForward(childAddr);
+                }
+            }
+            finally {
+                freeBuffer(innerBuf);
```
ibessonov, line 1277: Is it possible to allocate a single buffer per `IgniteIndexReader`?

vmalin: I tried to do it. A single buffer is definitely not enough. After shooting my foot a couple of times I did some optimizations with class-wize buffer. But the usage of such buffer seems very error prune - you have to be always aware of who writes there, who rewinds, where the addresses of this buffer are still in use. So I propose to leave it as it is - use a separate buffer for every method that needs it. It will lead to several extra allocations but will protect from some non-trivial buffer misuses.

modules/control-utility/src/main/java/org/apache/ignite/internal/commandline/indexreader/IgniteIndexReader.java
vmalin: fixed

vmalin:
>I don't see a --data-consistency-check option, as proposed in the description
The proposal to hide the check under an option was made by @ashapkin "just in case". I don't think we should add it. It will have to be added to the public API and will just contribute to confusion. 
But if you think that the extra check should be hidden behind an option I'll add it.

ibessonov (approved):

PR #3396, opened 2025-03-04
GG-41844 ODBC: Add timeout logging

isapego:
https://ggsystems.atlassian.net/browse/GG-41844

- Added logging of timeouts for query execution and any network operation;
- Removed checks for `_DEBUG` for logging, meaning, we now log all the information even in `Release` build;
- Fixed CMake a bit to work with the latest OpenSSL.


arcolight (approved):
LGTM

PR #3422, opened 2025-04-23
GG-39574 Improve inline index feature

Phillippko:
https://ggsystems.atlassian.net/browse/GG-39574


ivanzlenko (changes requested):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -252,6 +250,12 @@ public H2Tree(
         this.affinityKey = affinityKey;
         this.mvccEnabled = mvccEnabled;
 
+        int itemOverhead = Long.BYTES + (mvccEnabled ? Long.BYTES + Long.BYTES + Integer.BYTES : 0);
```
sashapolo, line 253: Where does `Long.BYTES + Long.BYTES + Integer.BYTES` come from?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -252,6 +250,12 @@ public H2Tree(
...
+        this.maxAllowedInlineSize = Math.min(
+                PageIO.MAX_PAYLOAD_SIZE,
+                (pageMem.realPageSize(grpId) - BPlusIO.ITEMS_OFF - 3 * AbstractDataPageIO.LINK_SIZE) / 2 - itemOverhead
```
sashapolo, line 256: Please add comments explaining what is happening here

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -252,6 +250,12 @@ public H2Tree(
         this.affinityKey = affinityKey;
         this.mvccEnabled = mvccEnabled;
 
+        int itemOverhead = Long.BYTES + (mvccEnabled ? Long.BYTES + Long.BYTES + Integer.BYTES : 0);
+        this.maxAllowedInlineSize = Math.min(
```
sashapolo, line 254: There exists a method called `H2TreeIndexBase#computeInlineSize`, I'm pretty sure that we should move this calculation there or at least modify it to not use `PageIO.MAX_PAYLOAD_SIZE`

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -850,7 +862,21 @@ private void inlineSizeRecomendation(SearchRow row) {
             colNames.add(index.columnName());
         }
 
-        if (newSize > inlineSize()) {
+        String idxType = pk ? "PRIMARY KEY" : affinityKey ? "AFFINITY KEY (implicit)" : "SECONDARY";
+
+        if (newSize > maxAllowedInlineSize) {
+            String warn = "Size of indexed columns of the row exceeds maximum index tree item size, " +
```
sashapolo, line 868: What is this warning about? How should clients react to it?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -144,8 +143,6 @@ static int computeInlineSize(
         }
 
         if (cfgInlineSize != -1) {
-            cfgInlineSize = Math.min(PageIO.MAX_PAYLOAD_SIZE, cfgInlineSize);
-
             if (fixedSize && size < cfgInlineSize) {
```
sashapolo, line 146: I see, you've actually modified this method. Why do we still have the warning below then? And why do we compute the max allowed size inside the tree and not here?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -157,7 +154,7 @@ static int computeInlineSize(
             return cfgInlineSize;
         }
 
-        return Math.min(PageIO.MAX_PAYLOAD_SIZE, size);
+        return size;
```
sashapolo, line 157: I don't understand this change. What if the computed inline size (not the one set through configuration) will exceed the maximum allowed inline size?

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Config/cache-query.xml
```diff
@@ -87,7 +87,6 @@
                                                 </map>
                                             </property>
                                             <property name="indexType" value="SORTED"/>
-                                            <property name="inlineSize" value="2345"/>
```
sashapolo, line 90: Why is this change needed?

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -1984,6 +1988,61 @@ public void testCreateSystemIndexWithSpecifiedInlineSizeByDdl() throws Exception
         assertTrue(lsnr.check());
     }
 
+    /**
+     *
+     */
+    @Test
+    public void testCreateIndexFailsWhenInlineSizeExceedsMax() throws Exception {
```
sashapolo, line 1995: These tests are not enough, I think. We need to check the following scenarios:
1. Too big inline size was set through the configuration.
2. We set nothing in configuration and create an index with a lot of columns, and we check that the computed inline size is capped at the maximum inline size.

Phillippko: If too big inline size is set through configuration, start of the node should fail, correct? Added tests

modules/platforms/dotnet/Apache.Ignite.Core.Tests/Config/cache-query.xml
Phillippko: This inline size exceeds the maximum. Changed to "1234"

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
Phillippko: This warning means that configured size exceeds the total size of index columns, not the maximum allowed size

Computing max allowed size requires some knowledge of the tree (i.e. realPageSize), and for ```H2TreeClientIndex``` we can't calculate this value (it has no page memory afaik, but we calculate inlineSize for it)

Also we need this computed value in ```H2Tree.inlineSizeRecomendation()``` for "Fix the “Indexed columns of a row cannot be fully inlined into index” warning to recommend at most the maximum allowed inline size."

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
Phillippko: Size of indexed columns is too big for inlining, and it can't be fixed by increasing inline size. I assumed client might use this knowledge, (i.e. think about reducing the scope of the index or just understand why index is slow)

Removed this warning

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
Phillippko: Answered in another thread why it is calculated here

Why we shouldn't use MAX_PAYLOAD_SIZE? Should I remove ```Math.min()``` and just use the calculated value?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
Phillippko: Created constants in ```H2IOUtils``` (storing of this overhead happens there) and method to get item overhead

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
sashapolo: > If too big inline size is set through configuration, start of the node should fail, correct? 

For new nodes - yes, for nodes with existing PDS based on this config - no. I would also recommend to notify @ashapkin about this change and confirm that the CS team approve this change or we should implement a different behavior.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -343,6 +362,16 @@ public H2Tree(
...
     }
 
+    private int maxAllowedInlineSize() {
+        // To avoid performance degradation, at least two items should fit into one page.
+        // So maximum payload size equals: P = (PS - H - 3L) / 2 - X , where P - Payload size, PS - page size, H - page
+        // header size, L - size of the child link, X - overhead per item.
+        int calculatedMaxItemSize = (pageMem.realPageSize(grpId) - BPlusIO.ITEMS_OFF - 3 * AbstractDataPageIO.LINK_SIZE)
+                / 2 - H2IOUtils.itemOverhead(mvccEnabled);
+
+        return Math.min(PageIO.MAX_PAYLOAD_SIZE, calculatedMaxItemSize);
```
sashapolo, line 372: How is it possible that `calculatedMaxItemSize` will be greater than `MAX_PAYLOAD_SIZE`?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
sashapolo: > Should I remove Math.min() and just use the calculated value?

Yes, I don't know how MAX_PAYLOAD_SIZE will be useful here.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeClientIndex.java
```diff
@@ -80,7 +81,7 @@ public static H2TreeClientIndex createIndex(
         List<InlineIndexColumn> inlineCols = getAvailableInlineColumns(false, ccfg.getName(),
             idxName, log, pk, tbl, cols, new InlineIndexColumnFactory(tbl.getCompareMode()), true);
 
-        inlineSize = computeInlineSize(idxName, inlineCols, inlineSize, ccfg.getSqlIndexMaxInlineSize(), log);
+        inlineSize = computeInlineSize(idxName, inlineCols, inlineSize, ccfg.getSqlIndexMaxInlineSize(), PageIO.MAX_PAYLOAD_SIZE, log);
```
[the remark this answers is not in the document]
Phillippko: Asked @AMashenkov , seems like we can use 0 for them. Fixed

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
Phillippko: Asked, he agrees that it is OK

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -1984,6 +1988,61 @@ public void testCreateSystemIndexWithSpecifiedInlineSizeByDdl() throws Exception
...
+                        + "AFFINITY_KEY=ID_AFF,"
+                        + "PK_INLINE_SIZE=" + inlineSize + ","
+                        + "AFFINITY_INDEX_INLINE_SIZE=" + inlineSize
+                        + "\""
+                ),
+                IgniteSQLException.class,
+                "Failed to complete exchange process"
+        );
+
+        assertTrue(X.hasCause(e, expMsg, IgniteCheckedException.class));
```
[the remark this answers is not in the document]
Phillippko: assertj is not available(

sashapolo (approved):

ivanzlenko (approved):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -319,6 +326,14 @@ public H2Tree(
                 upgradeMetaPage(inlineObjSupported);
         }
         else {
+            if (configuredInlineSize > maxAllowedInlineSize) {
```
AMashenkov, line 329: This validation should be done on index createion.
We should never fail when start node with existing index, just because we change the algo for inline size calculation.

AMashenkov (changes requested):
H2Tree constructor may fails due to new logic.
I see no compatibility tests that existed indexes are not affected even if they don't match the new conditions.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
Phillippko: It is calculated only if initNew = true, so it should run only for new indexes, no?

It would fail on start only if user pre-configured new index with incorrect value

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -329,8 +344,14 @@ public H2Tree(
...
                 table, cols, factory, true);
 
-            inlineSize = computeInlineSize(idxName, inlineIdxs, configuredInlineSize,
-                    cctx.config().getSqlIndexMaxInlineSize(), log);
+            inlineSize = computeInlineSize(
+                    idxName,
+                    inlineIdxs,
+                    configuredInlineSize,
+                    cctx.config().getSqlIndexMaxInlineSize(),
+                    maxAllowedInlineSize,
```
AMashenkov, line 352: Why do we have both parameters?
Can we validate `cctx.config().getSqlIndexMaxInlineSize()` against maxAllowedInlineSize, then use here a minimum of them?

Also, we could rewrite a `sqlIndexMaxInlineSize` in the config with `maxAllowedInlineSize` on the server-side, when index is created. After this, on client side, `sqlIndexMaxInlineSize` value will be aligned with `maxAllowedInlineSize` and system view will return correct information.

Phillippko: There is different logic for  ```sqlIndexMaxInlineSize```, I think it's better not to extract it

```
        int propSize = maxInlineSize == -1 ? IgniteSystemProperties.getInteger(IgniteSystemProperties.IGNITE_MAX_INDEX_PAYLOAD_SIZE,
            IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT) : maxInlineSize;
```

About writing to config - it seems that client index is actually created before server one

sashapolo (approved):

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheContextInfo.java
```diff
@@ -51,9 +55,10 @@ public class GridCacheContextInfo<K, V> {
      * Constructor of full cache context.
      *
      * @param cctx Cache context.
+     * @param kctx Kernel context.
      * @param clientCache Client cache or not.
      */
-    public GridCacheContextInfo(GridCacheContext<K, V> cctx, boolean clientCache) {
+    public GridCacheContextInfo(GridCacheContext<K, V> cctx, GridKernalContext kctx, boolean clientCache) {
```
AMashenkov, line 61: You don't need to pass KernalContext here.
It is available via  `cctx.kernalContext()`.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheContextInfo.java
```diff
@@ -62,20 +67,22 @@ public GridCacheContextInfo(GridCacheContext<K, V> cctx, boolean clientCache) {
...
      */
-    public GridCacheContextInfo(DynamicCacheDescriptor cacheDesc) {
+    public GridCacheContextInfo(DynamicCacheDescriptor cacheDesc, GridKernalContext ctx) {
         config = cacheDesc.cacheConfiguration();
         dynamicDeploymentId = cacheDesc.deploymentId();
         groupId = cacheDesc.groupId();
         cacheId = CU.cacheId(config.getName());
 
         clientCache = true;
+        this.kctx = ctx;
```
AMashenkov, line 85: You don't need to pass `KernalContext` here.
`KernalContext` can be get from `DynamicCacheDescriptor`,  there is a unused parameter in  `DynamicCacheDescriptor` constructor.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -90,17 +96,35 @@ protected H2TreeIndexBase(GridH2Table tbl, String name, IndexColumn[] cols, Inde
...
     }
 
+    /**
+     * To avoid tree corruption, at least two items should fit into one page.
+     * So maximum payload size equals: P = (PS - H - 3L) / 2 - X , where P - Payload size, PS - page size, H - page
+     * header size, L - size of the child link, X - overhead per item.
+     */
+    static int maxAllowedInlineSize(IgniteConfiguration cfg, boolean mvccEnabled) {
+        int configuredPageSize = cfg.getDataStorageConfiguration() == null
+                ? cfg.getDataStorageConfiguration().getPageSize() : DataStorageConfiguration.DFLT_PAGE_SIZE;
```
AMashenkov, line 106: I bet this will fails with NPE

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/DynamicCacheDescriptor.java
```diff
@@ -157,6 +161,13 @@ public DynamicCacheDescriptor(GridKernalContext ctx,
         this.cacheCfgEnrichment = cacheCfgEnrichment;
     }
 
+    /**
+     * @return Kernal context.
+     */
+    public GridKernalContext kctx() {
```
ygerzhedovich, line 167: We don't need the method at all. And shouldn't add a new field in the class

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheContextInfo.java
```diff
@@ -138,6 +144,13 @@ public boolean mvccEnabled() {
         return cctx;
     }
 
+    /**
+     * @return Kernal context.
+     */
+    public GridKernalContext kctx() {
```
ygerzhedovich, line 150: the same

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -90,17 +96,35 @@ protected H2TreeIndexBase(GridH2Table tbl, String name, IndexColumn[] cols, Inde
         return true;
     }
 
+    /**
+     * To avoid tree corruption, at least two items should fit into one page.
```
ygerzhedovich, line 100: Why does a single item lead a tree corruption?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -90,17 +96,35 @@ protected H2TreeIndexBase(GridH2Table tbl, String name, IndexColumn[] cols, Inde
...
+     * So maximum payload size equals: P = (PS - H - 3L) / 2 - X , where P - Payload size, PS - page size, H - page
+     * header size, L - size of the child link, X - overhead per item.
+     */
+    static int maxAllowedInlineSize(IgniteConfiguration cfg, boolean mvccEnabled) {
```
ygerzhedovich, line 104: IgniteConfix available here trought table 
'''tbl.cacheContext().gridConfig()'''

The same for mvcc flag

Phillippko: cache context is nullable, it is not set on client nodes - but we need to calculate inline size even there for ```SYSTEM.INDEXES``` view. Maybe for something else

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
ibessonov: https://support.gridgain.com/a/tickets/20077

Such trees are unbalanced, and they can overflow the max height of 128 layers

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -319,6 +324,14 @@ public H2Tree(
                 upgradeMetaPage(inlineObjSupported);
         }
         else {
+            if (configuredInlineSize > maxAllowedInlineSize) {
```
ygerzhedovich, line 327: Let's consider a case when user has a cluster on a previous version GG with an index with huge inline size. User upgrade the cluster to a new version and add a new node where the index tree should be created consequently. What is our expectation for such scenario? 
I propose to implement such test

Phillippko: There is actually such a test (testCreateSystemIndexFailsWhenConfiguredInlineSizeExceedsMax), and it is discussed that Alexander Shapkin that it is OK to fail starting _*new*_ node with incorrect index

sashapolo: There already exists a ticket: https://ggsystems.atlassian.net/browse/GG-40161

Phillippko: Naming is incorrect, my bad. Test should be called something like "testStartNewNodeFails...", but I believe it covers this case - start of the new node fails if its configuration contained index with too big inline

ygerzhedovich: The cases are different when you already have such an index on multiple nodes and when you try to create such an index for the first time. But anyway, if you have approval for the current approach, that's fine

ygerzhedovich (approved):

modules/indexing/src/test/java/org/apache/ignite/internal/processors/cache/index/BasicIndexTest.java
```diff
@@ -1984,6 +1988,100 @@ public void testCreateSystemIndexWithSpecifiedInlineSizeByDdl() throws Exception
...
+        assertThrows(
+                log,
+                () -> sql("CREATE INDEX FAILING_IDX ON TEST (VAL) inline_size " + idxInlineSize),
+                IgniteSQLException.class,
+                "Inline size is too big [cacheName=TEST" +
+                        ", tableName=SQL_PUBLIC_TEST" +
+                        ", idxName=FAILING_IDX" +
+                        ", configuredInlineSize=" + idxInlineSize +
+                        ", maxAllowedInlineSize="
+        );
```
AMashenkov, line 2011: Also, 
- let's check the cluster is operable after the error.
- let's check system view shows correct inline size value on server and client (and create a ticket to fix this)
- let's check client and server works fine, when page size on server side is larger than default 4k. E.g. 8k.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeClientIndex.java
```diff
@@ -75,12 +75,20 @@ public static H2TreeClientIndex createIndex(
...
-        CacheConfiguration ccfg = tbl.cacheInfo().config();
+        GridCacheContextInfo cacheInfo = tbl.cacheInfo();
+
+        CacheConfiguration ccfg = cacheInfo.config();
 
         List<InlineIndexColumn> inlineCols = getAvailableInlineColumns(false, ccfg.getName(),
-            idxName, log, pk, tbl, cols, new InlineIndexColumnFactory(tbl.getCompareMode()), true);
+                idxName, log, pk, tbl, cols, new InlineIndexColumnFactory(tbl.getCompareMode()), true);
+
+        int maxAllowedInlineSize = H2TreeIndexBase.maxAllowedInlineSize(
```
AMashenkov, line 85: It is incorrect to calculate value on client, because page size is unknown, EncryptionSpi can? be unavailable on client.
Maybe better always return -1 ?

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
AMashenkov: You can add a method for calculating maxInlineSize that will return max allowed value regarding the system property, cache configuration and memory configuration.

AMashenkov:
After short call decided
- do not fail if desired inline exceeds max allowed size and use max allowed size instead
- print warning if we changed desired inline size
- allow '-1' inline size in system view, when max allowed size can't be calculated (non-affinity or client node)

The last one is a regression. If we are not ok with this, let's create a ticket to fix this.
The proper fix is to calculate inline size before creating cache/index. On non-affinity nodes it may be impossible to calculate inline size from page size correctly. Possibly, we could limit minimal page size and introduce a constant for max inline size, which would be valid for page of minimal size.

AMashenkov (changes requested):
Please, make sure the code is correct and all the tests are passed.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -332,6 +341,8 @@ public H2Tree(
             inlineSize = computeInlineSize(idxName, inlineIdxs, configuredInlineSize,
                     cctx.config().getSqlIndexMaxInlineSize(), log);
 
+            assert inlineSize <= MAX_INLINE_SIZE;
```
AMashenkov, line 344: This should be a warning, otherwise user may failed to start a node.

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -110,10 +113,26 @@ static int computeInlineSize(
...
 
-        if (propSize == 0)
+            if (propSize > MAX_INLINE_SIZE)
+                log.warning("System property IGNITE_MAX_INDEX_PAYLOAD_SIZE exceeds maximum allowed size. Ignoring" +
+                        "[index=" + name + ", propertySize=" + propSize + ", maxAllowedInlineSize=" + MAX_INLINE_SIZE + ']');
+
+            maxSize = Math.min(propSize, MAX_INLINE_SIZE);
+        }
+
+        if (maxSize == 0)
```
AMashenkov, line 135:
```suggestion
        if (maxSize == 0 || cfgInlineSize == 0)
```

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -110,10 +113,26 @@ static int computeInlineSize(
...
             return 0;
 
         int size = 0;
```
AMashenkov, line 138: computedInlineSize

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2Tree.java
```diff
@@ -252,6 +253,14 @@ public H2Tree(
         this.affinityKey = affinityKey;
         this.mvccEnabled = mvccEnabled;
 
+        if (configuredInlineSize > MAX_INLINE_SIZE)
+            U.warn(log, "Configured inline size is too big [cacheName=" + cacheName +
+                    ", tableName=" + tblName +
+                    ", idxName=" + idxName +
+                    ", configuredInlineSize=" + configuredInlineSize +
+                    ", maxAllowedInlineSize=" + MAX_INLINE_SIZE + ']'
```
AMashenkov, line 261:
```suggestion
                    ", recommended=" + calculate(inlineIdxs0) + ']'
```

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/h2/database/H2TreeIndexBase.java
```diff
@@ -154,10 +169,21 @@ static int computeInlineSize(
...
+            if (cfgInlineSize > maxSize)
+                log.warning("Explicit INLINE_SIZE exceeds maximum size. Ignoring " +
+                        "[index=" + name + ", explicitInlineSize=" + cfgInlineSize + ", maxInlineSize=" + maxSize + ']');
+
+            return Math.min(cfgInlineSize, maxSize);
+        }
+
+        if (size > maxSize) {
+            log.warning("Calculated inline size exceeds maximum size. Ignoring " +
+                    "[index=" + name + ", calculatedInlineSize=" + size + ", maxInlineSize=" + maxSize + ']');
```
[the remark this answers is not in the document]
Phillippko: Fixed. Btw, we discussed breaking iterating indexcolumns, but we can't to it - because we should also check all columns for fixed sizes

AMashenkov (approved):

PR #3428, opened 2025-05-13
GG-43272 Add vector similarity function

manini-zs:


modules/core/src/main/java/org/apache/ignite/cache/QueryIndex.java
```diff
@@ -55,6 +56,11 @@ public class QueryIndex implements Serializable {
     /** */
     private int inlineSize = DFLT_INLINE_SIZE;
 
+    /**
+     * Vector Similarity Function for VECTOR index.
+     */
+    private QueryVectorField.SimilarityFunction similarityFunction;
```
ptupitsyn, line 62: Public API issue: `SimilarityFunction` should not be in `annotations` package. It is not only used for annotations, but also here in `QueryIndex`. 

There are two ways to define "Query Entities" - via annotations, or by building a `QueryEntity` instance manually. So the new enum should be in the `org.apache.ignite.cache` package. While `QueryVectorField` remains in `annotations`.

ptupitsyn (changes requested):
Protocol changes look good to me.

modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QueryVectorField.java
```diff
@@ -33,5 +33,52 @@
...
+         */
+        public int getSimilarityFunctionId() {
+            return similarityFunctionId;
+        }
+
+        /**
+         * Converts an integer ID to the corresponding SimilarityFunction.
+         * @param id The similarity function ID.
+         * @return The SimilarityFunction corresponding to the ID.
+         * @throws IllegalArgumentException if the ID doesn't match any similarity function.
```
ygerzhedovich, line 73: javadoc does not correspond to the implementation

modules/core/src/main/java/org/apache/ignite/cache/query/annotations/QueryVectorField.java
```diff
@@ -33,5 +33,52 @@
 @Retention(RetentionPolicy.RUNTIME)
 @Target({ElementType.FIELD, ElementType.TYPE})
 public @interface QueryVectorField {
-    // No-op.
-}
+    /**
+     * Returns similarityFunction for a vector field.
+     * @return similarityFunction for a vector field.
+     */
+    SimilarityFunction similarityFunction() default SimilarityFunction.COSINE;
```
ygerzhedovich, line 40: should be default value reflected in javadoc?

modules/core/src/main/java/org/apache/ignite/cache/QueryIndex.java
manini-zs: Done

modules/core/src/main/java/org/apache/ignite/cache/SimilarityFunction.java
```diff
@@ -0,0 +1,47 @@
...
+     * @return The SimilarityFunction corresponding to the ID, or COSINE if no match is found
+     */
+    public static SimilarityFunction fromId(int id) {
+        for (SimilarityFunction func : values()) {
+            if (func.getSimilarityFunctionId() == id) {
+                return func;
+            }
+        }
+        return COSINE;
+    }
```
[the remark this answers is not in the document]
manini-zs: Done

ptupitsyn (approved):

ygerzhedovich (approved):

PR #3430, opened 2025-05-13
GG-41981 Execute kill map SQL fragments on management pool

ygerzhedovich:
https://ggsystems.atlassian.net/browse/GG-41981


modules/indexing/src/test/java/org/apache/ignite/internal/processors/query/KillQueryTest.java
```diff
@@ -228,6 +237,10 @@ else if (delegate instanceof SchemaProposeDiscoveryMessage) {
             }
         }.setIpFinder(IP_FINDER));
 
+        // It is important to have just single thread for the query execution pool
+        // to check the ability to cancel queries, even query pool is busy.
+        cfg.setQueryThreadPoolSize(1);
```
zstan, line 242: i think i understand why you change this option for ALL existing tests, cause grids are starts from 
KillQueryTest#beforeTestsStarted
I think it`s not clear, can you implement concrete mentioned case from issue and use probably different test class and use equal approach as in IgniteQueryDedicatedPoolTest ?

zstan (approved):

PR #3431, opened 2025-05-16
GG-40393 Restore binary context after client reconnect.

AMashenkov:


modules/core/src/test/java/org/apache/ignite/internal/IgniteClientReconnectCustomBinaryContextTest.java
```diff
@@ -0,0 +1,296 @@
...
+                    ReflectionUtils.setField(objectField, obj, dateTime);
+                }
+            }
+        }
+    }
+
+    /**
+     * Discovery SPI, that makes a node stop sending messages when {@code block} is set to {@code true}.
+     */
+    private static class BlockedDiscoverySpi extends TcpDiscoverySpi {
```
zstan, line 252: can we reuse it from org.apache.ignite.internal.processors.cache.binary.ClientReconnectTest.BlockedDiscoverySpi ?

zstan (approved):

korlov42 (approved):

PR #3440, opened 2025-06-02
GG-42730 Fix PartitionReconciliationFullFixStressTest

sk0x50:
https://ggsystems.atlassian.net/browse/GG-42730


sk0x50:
The change is trivial, so a review is not necessary.

PR #3444, opened 2025-06-10
GG-37211 Implement ClientCacheChangeDummyDiscoveryMessage processing on inactive cluster.

ibelyakov:


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCachePartitionExchangeManager.java
```diff
@@ -405,8 +406,20 @@ private void processEventInactive(DiscoveryEvent evt, DiscoCache cache) {
...
-            log.debug("Ignore event, cluster is inactive: " + evt);
+        if (evt.type() == EVT_DISCOVERY_CUSTOM_EVT &&
+                (((DiscoveryCustomEvent)evt).customMessage() instanceof ClientCacheChangeDummyDiscoveryMessage)) {
+
+            ClientCacheChangeDummyDiscoveryMessage dummyDiscoveryMsg =
+                    (ClientCacheChangeDummyDiscoveryMessage)((DiscoveryCustomEvent)evt).customMessage();
+
+            IgniteException err = new IgniteException("Can not perform the operation because the cluster is inactive.");
+
+            cctx.cache().completeClientCacheChangeFuture(dummyDiscoveryMsg.requestId(), err);
```
sk0x50, line 417: In general, the rule of thumb is to process such messages within `exchange worker`'s thread, but in this particular case, it seems to be safe. In any case, we can improve this later if we find some corner-case scenarios.

sk0x50 (approved):

PR #3447, opened 2025-06-16
GG-43819 Improved log output in case of potential index inconsistency

sk0x50:
https://ggsystems.atlassian.net/browse/GG-43819


DirectXceriD (commented):
left one coment

PR #3448, opened 2025-06-18
GG-43080 VisorConnectivityTask execution leads to nodes failure

tkalkirill:
https://ggsystems.atlassian.net/browse/GG-43080

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
@@ -2095,7 +2095,7 @@ public class GridNioServer<T> {
                                 SelectionKey key = ch.keyFor(selector);
 
                                 if (key != null)
-                                    ses = (GridSelectorNioSessionImpl)key.attachment();
+                                    ses = sessionForCancelConnect(key);
 
                                 if (ses == null) {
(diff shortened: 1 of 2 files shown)


modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -4507,4 +4507,16 @@ interface SessionChangeRequest {
...
+    private static @Nullable GridSelectorNioSessionImpl sessionForCancelConnect(SelectionKey key) {
+        Object a = key.attachment();
+
+        if (a instanceof GridSelectorNioSessionImpl)
+            return (GridSelectorNioSessionImpl) a;
+        else if (a instanceof NioOperationFuture && ((NioOperationFuture<?>) a).operation() == NioOperation.CONNECT)
+            // Canceled immediately after NioOperation.CONNECT without waiting NioOperation.REGISTER.
+            return null;
+
+        throw new IgniteException("No session found when closing connection");
```
ibessonov, line 4520: Should we simply return `null` as well? There's this code after, it'll handle everything I suppose
```
if (ses == null) {
    // Session was not created yet.
    if (key != null)
        key.cancel();

    U.closeQuiet(ch);
}
```

ibessonov (approved):

PR #3453, opened 2025-06-23
GG-43386 Fixed process crash during execution GridCommandLineLoaderTest

sk0x50:
https://ggsystems.atlassian.net/browse/GG-43386


DirectXceriD (commented):
looks good

PR #3457, opened 2025-07-07
GG-37744 Improve logging related to enabling/disabling WAL

Phillippko:
https://ggsystems.atlassian.net/browse/GG-37744


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/CacheGroupContext.java
```diff
@@ -260,6 +260,13 @@ public CacheGroupContext(
 
         metrics = new CacheGroupMetricsImpl(this);
 
+        if (log.isInfoEnabled())
+            log.info("WAL settings on startup [grpId=" + grpId +
```
rpuch, line 264: The issue doesn't seem to request this change. Why is it introduced?

rpuch (approved):

PR #3473, opened 2025-08-05
GG-38135 Add possibility to change priority through SqlFieldsQuery

zstan:


modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
```diff
@@ -421,6 +435,39 @@ public boolean isLazy() {
         return lazy;
     }
 
+    /**
+     * Sets priority for a query.
+     * Queries with higher priority will be processed faster.
```
ygerzhedovich, line 440: It will be good to describe what is higher and what is lower.

modules/core/src/main/java/org/apache/ignite/internal/processors/pool/PoolProcessor.java
```diff
@@ -462,7 +465,39 @@ public PoolProcessor(GridKernalContext ctx) {
...
+                        rightPriority = (byte)-right.priority();
+                        rightTs = right.initializedTs();
+                    }
+
+                    if (leftPriority != null && rightPriority != null) {
+                        int comp = Byte.compare(leftPriority, rightPriority);
+                        return comp == 0 ? Long.compare(leftTs, rightTs) : comp;
+                    }
+                    else if (leftPriority != null || rightPriority != null) {
+                        return leftPriority != null ? leftPriority : rightPriority;
```
ygerzhedovich, line 492: It doesn't make a sense. You always return a positive number, which means the first argument is less than the second.

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
```diff
@@ -50,6 +50,16 @@ public class SqlFieldsQuery extends Query<List<?>> {
     /** Default value of the update internal batch size. */
     private static final int DFLT_UPDATE_BATCH_SIZE = 1;
 
+    /**
+     * The minimum priority that a query can have.
+     */
+    public static final int MIN_PRIORITY = 0;
```
AMashenkov, line 56: Let's use full byte range [-128,127]

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/PriorityWrapper.java
```diff
@@ -0,0 +1,52 @@
...
+
+package org.apache.ignite.internal.managers.communication;
+
+/** Holder for priority processing. */
+public class PriorityWrapper implements Runnable {
+    private final Runnable clo;
+
+    private final Byte priority;
+
+    private final long initializedTs;
```
AMashenkov, line 25:
```suggestion
    /** This value is used to preserve natural ordering for tasks with the same priority. Because PriorityQueue doesn't guarantee FIFO order for the same priority. */
    private final long initializedTs;
```

modules/core/src/main/java/org/apache/ignite/cache/query/SqlFieldsQuery.java
zstan: ho ho, i don`t like such a great granularity

AMashenkov: [-10,10] ?
I just think the one may want to enqueue low-priority task, which will be executed after tasks with default priority.

modules/core/src/main/java/org/apache/ignite/internal/processors/pool/PoolProcessor.java
zstan: thanks, this part is totally refactored.

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/GridIoManager.java
```diff
@@ -1303,6 +1305,16 @@ private void processRegularMessage(
                 }
             }
 
+            if (plc == GridIoPolicy.QUERY_POOL) {
+                Message initMsg = msg.message();
+                byte priority = SqlFieldsQuery.MIN_PRIORITY;
```
ygerzhedovich, line 1310: Let's introduce DEFAULT_PRIORITY and set here insted og MINIMUM. I think we should take 5 as the default to have the ability to run any other query with a bigger and lower priority

modules/core/src/main/java/org/apache/ignite/internal/managers/communication/PriorityWrapper.java
ygerzhedovich: I don't see the change

ygerzhedovich (approved):

PR #3474, opened 2025-08-08
GG-44530 Revert change of IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT

Phillippko:
https://ggsystems.atlassian.net/browse/GG-44530

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
+++ b/modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
@@ -134,8 +134,8 @@ import static org.apache.ignite.internal.util.GridLogThrottle.DFLT_LOG_THROTTLE_
 import static org.apache.ignite.internal.util.GridReflectionCache.DFLT_REFLECTION_CACHE_SIZE;
 import static org.apache.ignite.internal.util.GridUnsafe.DFLT_MEMORY_PER_BYTE_COPY_THRESHOLD;
 import static org.apache.ignite.internal.util.IgniteExceptionRegistry.DEFAULT_QUEUE_SIZE;
-import static org.apache.ignite.internal.util.IgniteUtils.*;
 import static org.apache.ignite.internal.util.IgniteUtils.DFLT_MBEAN_APPEND_CLASS_LOADER_ID;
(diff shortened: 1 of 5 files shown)


modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -134,6 +134,7 @@
 import static org.apache.ignite.internal.util.GridReflectionCache.DFLT_REFLECTION_CACHE_SIZE;
 import static org.apache.ignite.internal.util.GridUnsafe.DFLT_MEMORY_PER_BYTE_COPY_THRESHOLD;
 import static org.apache.ignite.internal.util.IgniteExceptionRegistry.DEFAULT_QUEUE_SIZE;
+import static org.apache.ignite.internal.util.IgniteUtils.*;
```
rpuch, line 137: Please don't use wildcard imports

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
```diff
@@ -1141,7 +1142,7 @@ public final class IgniteSystemProperties {
      * index store is disabled.
      */
     @SystemProperty(value = "Maximum payload size in bytes for H2TreeIndex. " +
-        "0 means that inline index store is disabled", type = Integer.class, defaults = "" + IgniteUtils.MAX_INLINE_SIZE)
+        "0 means that inline index store is disabled", type = Integer.class, defaults = "" + IGNITE_MAX_INDEX_PAYLOAD_SIZE_DEFAULT)
```
rpuch, line 1145: I already asked this question but I forgot the answer. Before introducing the change you are reverting it was 10, not 64. Why is it 64 now?

modules/core/src/main/java/org/apache/ignite/IgniteSystemProperties.java
Phillippko: In AI3 /GG 9 project IDEA doesn't do that, but in AI2 and GG8 it uses wildcards... Some day I will finally fix it :) 

Fixed

rpuch (approved):

PR #3483, opened 2025-08-19
GG-41778 Register local metadata during startup on a node with enabled persistence

xtern:
**Issue**:

During persistence-enabled node startup, we miss the local registration of metadata for the types specified in QueryEntity for statically configured caches. Because at the moment when we try to register metadata, there are no registered caches.

**Fix description**:

For a node with persistence enabled, an additional call to register metadata has been added after registering cache descriptors (GridCacheProcessor -> CacheRecoveryLifecycle -> onReadyForRead)

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
@@ -5507,6 +5507,8 @@ public class GridCacheProcessor extends GridProcessorAdapter {
             CacheJoinNodeDiscoveryData data = locCfgMgr.restoreCacheConfigurations();
 
             cachesInfo.onStart(data);
+
+            ctx.query().registerMetadataForRegisteredCaches(false);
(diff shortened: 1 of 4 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -5507,6 +5507,8 @@ private class CacheRecoveryLifecycle implements MetastorageLifecycleListener, Da
             CacheJoinNodeDiscoveryData data = locCfgMgr.restoreCacheConfigurations();
 
             cachesInfo.onStart(data);
+
+            ctx.query().registerMetadataForRegisteredCaches(false);
```
ibessonov, line 5511: Judging by your description, fix appears to be correct. Unfortunately, I'm also not the best person to say that, you know this code better than I do now :)

ibessonov (approved):

ygerzhedovich (approved):

PR #3492, opened 2025-09-04
GG-40285 [IGNITE-22562] Fixed notification message when Ignite Statistics usage state changes (#11494)

zstan:


AMashenkov (approved):

modules/indexing/src/main/java/org/apache/ignite/internal/processors/query/stat/IgniteStatisticsManagerImpl.java
```diff
@@ -202,8 +202,10 @@ else if (db == null)
 
         ctx.internalSubscriptionProcessor().registerDistributedConfigurationListener(dispatcher -> {
             usageState.addListener((name, oldVal, newVal) -> {
-                if (log.isInfoEnabled())
-                    log.info(String.format("Statistics usage state was changed from %s to %s", oldVal, newVal));
+                if (log.isInfoEnabled() && newVal != null) {
```
AMashenkov, line 205: Why do we ignore `null`?
Should we print default instead as for `oldVal`?

zstan: cause in such a case (without null ignoring) we will get strange : **Statistics usage state was changed from null to null** message, this was the main goal of this issue )

PR #3500, opened 2025-09-12
GG-45026 NPE on checkpoint write when cache with encryption is destroyed

vldpyatkov:
https://ggsystems.atlassian.net/browse/GG-45026


sk0x50:
I think that the data race is still here, unfortunately. Let's consider this scenario: 
 - two server nodes + encryption
 - create an encrypted cache. The encryption key is written to the metastore.
 - destroy this cache from the coordinator, for example (not important, to be honest).
 - the coordinator node completes its exchange and sends a full message to the second node. At this point, the encryption key is deleted from the metastore.
 - the second node is slow in processing the full message for some reason.
 - the coordinator's exchange thread is free to process a new exchange.
 - start creating a new cache with the same name and group from the coordinator, again.
 - `init()` on the coordinator executes and triggers the writing of a new key to the metastore.
 - the second node then wakes up and processes the full message from the destroy. As a result, the encryption key is deleted again.

If a checkpoint occurs on the first node (coordinator) before the second one processes a new exchange and writes the key back to the metastore, we get an NPE.

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/GridCacheProcessor.java
```diff
@@ -3822,26 +3823,34 @@ public IgniteInternalFuture<Boolean> dynamicStartCachesByStoredConf(
...
             Iterator<byte[]> grpKeysIter = F.isEmpty(grpKeys) ? null : grpKeys.iterator();
 
+            HashMap<String, byte[]> grpKeyMap = new HashMap<>();
+
             for (StoredCacheData ccfg : storedCacheDataList) {
-                assert grpKeysIter == null || !ccfg.config().isEncryptionEnabled() || grpKeysIter.hasNext();
+                CacheConfiguration<?, ?> cacxheCfg = ccfg.config();
 
-                byte[] encCacheKey = grpKeysIter != null && ccfg.config().isEncryptionEnabled() ? grpKeysIter.next() : null;
+                byte[] encCacheKey = grpKeyMap.get(cacxheCfg.getGroupName());
```
sk0x50, line 3831: It seems to me that using `getGroupName` here might be a problem. 
Let's assume the user starts two new caches without specifying a cache group. It means that two different cache groups should be created, right? On the other hand, after your change, we will use the same encryption key for both.

WDYT? Perhaps I am missing something?

vldpyatkov: You are right because in the case when the cache is self-grouped, we have a null in the configuration for the group name. I fixed it. using acache name instead of a group.
A test of this case was also added.

sk0x50 (approved):

PR #3520, opened 2025-10-21
GG-43971 Make name/version REST API available during node start

ibessonov:
https://ggsystems.atlassian.net/browse/GG-43971


modules/core/src/main/java/org/apache/ignite/internal/processors/rest/protocols/tcp/GridTcpRestNioListener.java
```diff
@@ -414,7 +411,7 @@ public void onSessionClosed(GridNioSession ses) {
         if (msg instanceof GridClientAuthenticationRequest) {
             GridClientAuthenticationRequest req = (GridClientAuthenticationRequest)msg;
 
-            restReq = new GridRestAuthenticationRequest();
+            restReq = new GridRestTaskRequest();
```
rpuch: Why is this changed like this?

modules/control-utility/src/test/java/org/apache/ignite/util/GridCommandHandlerTest.java
```diff
@@ -3270,20 +3270,20 @@ public void testSuccessStopWarmUp() throws Exception {
...
      *
      * @throws Exception If failed.
      */
     @Test
-    public void testFailStopWarmUp() throws Exception {
+    public void testStopCompletedWarmUp() throws Exception {
         startGrid(0);
 
-        assertEquals(EXIT_CODE_UNEXPECTED_ERROR, execute("--warm-up", "--stop", "--yes"));
+        assertEquals(EXIT_CODE_OK, execute("--warm-up", "--stop", "--yes"));
```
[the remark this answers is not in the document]
ibessonov: Old behavior made no real sense. We should never write a test that catches "unexpected error", that's oxymoron. I believe that this PR is an appropriate place to fix it, otherwise we'll never fix it

modules/core/src/main/java/org/apache/ignite/internal/processors/rest/protocols/tcp/GridTcpRestNioListener.java
ibessonov: I thought that `GridRestAuthenticationRequest` is not needed, but on the other hand I might have gone too far and should revert this particular change.

modules/clients/src/test/java/org/apache/ignite/internal/processors/rest/JettyRestBeforeNodeStartTest.java
```diff
@@ -62,7 +66,53 @@ public void testBasicCommands() throws Exception {
...
+     * @throws Exception If failed.
+     */
+    @Test
+    public void testBasicDuringPluginStart() throws Exception {
+        CountDownLatch startLatch = new CountDownLatch(1);
+        CountDownLatch stopLatch = new CountDownLatch(1);
+
+        AbstractTestPluginProvider provider = new AbstractTestPluginProvider() {
+            @Override public String name() {
+                return "BlockNodeStartPluginProvider";
```
rpuch: Is this a magic constant or just a name which can be changed freely?

rpuch (approved):

PR #3561, opened 2026-01-14
GG-43637 Fix deadlock on handshake retry

sashapolo:
https://ggsystems.atlassian.net/browse/GG-43637

In this PR we decided to do the following:

1. Disable the connection check message by default. There's a suspicion that this code doesn't do anything useful, I've created a ticket to remove it after a while: https://ggsystems.atlassian.net/browse/GG-46827
2. In order to fix the deadlock problem, when a connection check message is sent, we check the outgoing message queue and only add the message to it if the queue is empty.


modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -653,15 +654,29 @@ GridNioFuture<?> send(GridNioSession ses,
     /**
      * @param ses Session.
      * @param req Request.
-     * @param sys System message flag.
      * @throws IgniteCheckedException If session was closed.
      */
-    private void send0(GridSelectorNioSessionImpl ses, SessionWriteRequest req, boolean sys) throws IgniteCheckedException {
+    private void send0(GridSelectorNioSessionImpl ses, SessionWriteRequest req) throws IgniteCheckedException {
```
rpuch: Before your changes, it was the 'implementation' method, which was indicated by that zero in the name; but now it's a method of a higher level which 'sends a non-system message', so it doesn't seem to need that zero in the name anymore

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridSelectorNioSessionImpl.java
```diff
@@ -301,6 +301,23 @@ int offerSystemFuture(SessionWriteRequest writeFut) {
...
+    /**
+     * Tries to write a ping message to check this session's availability. If the outgoing queue of the
+     * session is not empty, then no additional messages are written.
+     */
+    // TODO: consider removing this code, https://ggsystems.atlassian.net/browse/GG-46827.
+    int checkConnection(SessionWriteRequest writeFut) {
+        int size = queue.sizex();
+
+        // There is a race condition here because we check the queue size and add the future without synchronization.
+        // This may result in a redundant ping message being sent, which we don't really care about.
```
rpuch: So the deadlock is not eliminated completely?

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -720,15 +735,30 @@ public void sendSystem(GridNioSession ses,
...
+     */
+    // TODO: consider removing this code, https://ggsystems.atlassian.net/browse/GG-46827.
+    public void checkConnection(GridNioSession ses) throws IgniteCheckedException {
+        SessionWriteRequest req = new WriteRequestSystemImpl(ses, new ConnectionCheckMessage());
+
+        GridSelectorNioSessionImpl impl = (GridSelectorNioSessionImpl)ses;
+
+        int msgCnt = impl.checkConnection(req);
+
+        handleSend(impl, req, msgCnt);
```
rpuch: It seems that we'll invoke `handleSend()` even if we didn't send anything. Why should we do it?

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationSpiWriteSemaphoreDeadlockTest.java
```diff
@@ -0,0 +1,261 @@
...
+
+        // Block sending messages from the server to the client.
+        GridNioSession serverSession = getSession(server, client);
+
+        insertBlockingFilter((GridNioSessionImpl) serverSession);
+
+        // Force the client to abandon the existing connection and issue a bunch of handshakes to establish a new one.
+        breakConnection(client, server);
+
+        assertEquals("1", cache.get(1));
```
rpuch: Does it hang when we encounter a deadlock? How about explicitly asserting that the `get` gets executed in, say, 10 seconds, to make the test code more readable?

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationSpiWriteSemaphoreDeadlockTest.java
```diff
@@ -0,0 +1,261 @@
...
+        }
+
+        @Override public void onExceptionCaught(GridNioSession ses, IgniteCheckedException ex) throws IgniteCheckedException {
+            proceedExceptionCaught(ses, ex);
+        }
+
+        @Override public GridNioFuture<?> onSessionWrite(GridNioSession ses, Object msg, boolean fut, IgniteInClosure<IgniteException> ackC) throws IgniteCheckedException {
+            if (msg instanceof ConnectionCheckMessage) {
+                try {
+                    // Block all selector threads.
```
rpuch: Does it block all selector threads or just current one?

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridSelectorNioSessionImpl.java
sashapolo: Theoretically, no. But it is nearly impossible to get it, because a lot of things need to happen at once:
1. Every existing selector thread reads the size of the queue (which is 0) and halts;
2. Some threads add QUEUE_SIZE messages into the queue (which is several thousand);
3. All selector threads resume operation and get blocked.

I really don't want to add some specific code to the queue implementation, but I think we can send this message as a system message instead, which bypasses the semaphore altogether

modules/core/src/test/java/org/apache/ignite/spi/communication/tcp/GridTcpCommunicationSpiWriteSemaphoreDeadlockTest.java
```diff
@@ -0,0 +1,261 @@
...
+ *     <li>We have a cluster with two nodes: a server and a client;</li>
+ *     <li>We then emulate a situation when it is possible to send messages from the client to the server but
+ *     <b>not</b> in the opposite direction;</li>
+ *     <li>We then force the client to abandon the existing connection, which makes it to try to open a new one;</li>
+ *     <li>The client sends a series of handshakes because messages from the server do not reach the client;</li>
+ *     <li>We unblock the server-to-client connection and expect normal operations afterward. Previously there
+ *     has been a bug that resulted in the selector threads on the server side to be deadlocked.</li>
+ * </ol>
+ */
+// TODO: consider removing this code, https://ggsystems.atlassian.net/browse/GG-46827.
```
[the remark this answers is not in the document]
sashapolo: This class tests a very specific scenario that relies on the logic of sending ConnectionCheckMessages. If we remove this logic, I don't know how to adapt this test and what should it even check

Erixonich (approved):

rpuch (approved):

modules/core/src/main/java/org/apache/ignite/internal/util/nio/GridNioServer.java
```diff
@@ -720,12 +735,29 @@ public void sendSystem(GridNioSession ses,
...
+    // TODO: consider removing this code, https://ggsystems.atlassian.net/browse/GG-46827.
+    public void checkConnection(GridNioSession ses) throws IgniteCheckedException {
+        SessionWriteRequest req = new WriteRequestImpl(ses, new ConnectionCheckMessage(), false, null);
+
+        GridSelectorNioSessionImpl impl = (GridSelectorNioSessionImpl)ses;
+
+        int msgCnt = impl.checkConnection(req);
+
+        if (msgCnt != -1) {
+            handleSend(impl, req, msgCnt);
```
ibessonov: I am a little bit confused by this code. What would be the difference between "offer" (in `checkConnection`) and "send"?

ibessonov (approved):

PR #3565, opened 2026-01-23
GG-43973 Fire management events on CC actions

nva:


shnus (commented):
I think the PR is not full.
I don't see event firing logic

shnus (approved):

PR #3566, opened 2026-02-02
GG-43957 Print messages about WAL read progress on startup

Phillippko:
https://ggsystems.atlassian.net/browse/GG-43957

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
@@ -951,13 +951,14 @@ public class FileWriteAheadLogManager extends GridCacheSharedManagerAdapter impl
 
     /** {@inheritDoc} */
     @Override public WALIterator replay(WALPointer start) throws IgniteCheckedException, StorageException {
-        return replay(start, null);
+        return replay(start, null, null);
     }
 
     /** {@inheritDoc} */
     @Override public WALIterator replay(
         WALPointer start,
-        @Nullable IgniteBiPredicate<WALRecord.RecordType, WALPointer> recordDeserializeFilter
+        @Nullable IgniteBiPredicate<WALRecord.RecordType, WALPointer> recordDeserializeFilter,
+        @Nullable IterationReason reason
     ) throws IgniteCheckedException, StorageException {
         assert start == null || start instanceof FileWALPointer : "Invalid start pointer: " + start;
 
@@ -982,7 +983,8 @@ public class FileWriteAheadLogManager extends GridCacheSharedManagerAdapter impl
             log,
             segmentAware,
             segmentRouter,
-            lockedSegmentFileInputFactory
+            lockedSegmentFileInputFactory,
+            reason
         );
 
         try {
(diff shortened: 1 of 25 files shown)


modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -951,13 +951,14 @@ private FileWriteHandle closeBufAndRollover(
 
     /** {@inheritDoc} */
     @Override public WALIterator replay(WALPointer start) throws IgniteCheckedException, StorageException {
-        return replay(start, null);
+        return replay(start, null, null);
```
rpuch: Is it useful at all if we pass null here?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -2804,20 +2806,24 @@ private static class RecordsIterator extends AbstractWalRecordsIterator {
...
-         * @param start Optional start pointer.
-         * @param end Optional end pointer.
-         * @param dsCfg Database configuration.
-         * @param serializerFactory Serializer factory.
-         * @param archiver File Archiver.
-         * @param decompressor Decompressor.
-         * @param log Logger  @throws IgniteCheckedException If failed to initialize WAL segment.
-         * @param segmentAware Segment aware.
-         * @param segmentRouter Segment router.
+         * @param cctx                    Shared context.
```
rpuch: Does the code style dictates all those indents?

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -2966,9 +2975,14 @@ private void init() throws IgniteCheckedException {
                 try {
                     fd = segmentRouter.findSegment(curWalSegmIdx);
 
-                    if (log.isDebugEnabled()) {
-                        log.debug("Reading next file [absIdx=" + curWalSegmIdx +
-                            ", file=" + fd.file.getAbsolutePath() + ']');
+                    if (log.isDebugEnabled() || log.isInfoEnabled()) {
```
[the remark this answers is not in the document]
rpuch: I screwed up with the suggestion, but still: isDebugEnabled() always implies isInfoEnabled(), so it's enough to check just for `isInfoEnabled()`. You cannot enable DEBUG and not enable INFO at the same time

modules/core/src/main/java/org/apache/ignite/internal/processors/cache/persistence/wal/FileWriteAheadLogManager.java
```diff
@@ -2975,11 +2982,11 @@ private void init() throws IgniteCheckedException {
                 try {
                     fd = segmentRouter.findSegment(curWalSegmIdx);
 
-                    if (log.isDebugEnabled() || log.isInfoEnabled()) {
+                    if (log.isInfoEnabled()) {
```
rpuch: Same here

rpuch (approved):

PR #3598, opened 2026-03-23
GG-43609 Improve PriorityQueueCollisionSpi Performance Under High Concurrency

Phillippko:
https://ggsystems.atlassian.net/browse/GG-43609

Problem was that dozens of onCollision methods were run in parallel every time job is submitted / completed. Every time we created slice of all wait jobs, and if there jobs to activate sorted them

compute().executeAsync is sync wrt to onCollision method. I didn't change it as 1. it requires to fix unknown number of tests as they expect sync behavior 2. makes code much messier 3. no benefits outside of synthethic stress test with huge number of short lived tasks 

There was also a problem with changing priorities of jobs that were sorted in other thread, leading to comparator exceptions

I made onCollision method sync to make it run sequentially + added fast exit route

For huge number of short lived tasks it still not great, because we always need to activate new jobs, but this is an extreme case that I don't expect to happen in real life

### DIFF
--- a/modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
+++ b/modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
@@ -102,6 +102,7 @@ import org.apache.ignite.spi.systemview.view.ComputeJobView;
 import org.apache.ignite.spi.systemview.view.ComputeJobView.ComputeJobState;
 import org.jetbrains.annotations.NotNull;
 import org.jetbrains.annotations.Nullable;
+import org.jetbrains.annotations.TestOnly;
 import org.jsr166.ConcurrentLinkedHashMap;
 
 import static java.util.concurrent.TimeUnit.MILLISECONDS;
@@ -918,8 +919,21 @@ public class GridJobProcessor extends GridProcessorAdapter {
         }
     }
 
-    private void handleCollisionsInternal() {
+    @TestOnly
+    /** Synchronised version of {@link #handleCollisions()} for testing purposes. */
+    public void handleCollisionsSync() {
         if (!rwLock.tryReadLock()) {
+            if (log.isDebugEnabled())
+                log.debug("Skipping handling collisions because node is stopping");
+
+            handlingCollisionFut.completeExceptionally(new NodeStoppingException("Node is stopping"));
+
+            return;
+        }
+
+        if (stopping) {
+            rwLock.readUnlock();
+
             if (log.isDebugEnabled())
                 log.debug("Skipping handling collisions because node is stopping");
 
@@ -1107,7 +1121,7 @@ public class GridJobProcessor extends GridProcessorAdapter {
             if (shouldHandleCollision && (handlingCollisionFut == null || handlingCollisionFut.isDone())) {
                 handlingCollisionFut = new CompletableFuture<>();
 
-                ctx.pools().getManagementExecutorService().submit(this::handleCollisionsInternal);
+                ctx.pools().getManagementExecutorService().submit(this::handleCollisionsSync);
 
                 handlingCollisionFut.whenComplete((res, err) -> {
                     if (err != null) {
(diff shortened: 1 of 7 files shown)


modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputePriorityQueueSpiConcurrencyTest.java
```diff
@@ -0,0 +1,145 @@
...
+
+        stopAllGrids();
+    }
+
+    /**
+     * {@inheritDoc}
+     */
+    @Override protected IgniteConfiguration getConfiguration(String igniteInstanceName) throws Exception {
+        return super.getConfiguration(igniteInstanceName)
+            .setFailureHandler(new StopNodeFailureHandler())
```
rpuch: Why do we need this?

modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputePriorityQueueSpiConcurrencyTest.java
```diff
@@ -0,0 +1,145 @@
...
+    @Test
+    public void testSubmitManyJobs() {
+        Ignite ignite = grid(0);
+
+        // When we submit new task we wait for collision SPI to start it if there are available slots.
+        // Tasks are short-lived, it happens often, so many threads are used to submit enough tasks fast.
+        ExecutorService executor = Executors.newFixedThreadPool(100);
+
+        try {
+            int taskCount = 100000;
```
rpuch:
```suggestion
            int taskCount = 100_000;
```

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
```diff
@@ -1069,22 +1126,43 @@ void advance() {
...
+
+    private void scheduleHandleCollisionsIfNeeded() {
+        synchronized (handleCollisionMutex) {
+            if (shouldHandleCollision && (handlingCollisionFut == null || handlingCollisionFut.isDone())) {
+                handlingCollisionFut = new CompletableFuture<>();
+
+                ctx.pools().getManagementExecutorService().submit(this::handleCollisionsInternal);
+
+                handlingCollisionFut.whenComplete((res, err) -> {
+                    if (!(err instanceof NodeStoppingException))
```
rpuch: Let's log an exception at least on debug level

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
```diff
@@ -1069,22 +1126,43 @@ void advance() {
...
+                        scheduleHandleCollisionsIfNeeded();
```
rpuch: Why is this recursion needed?

modules/core/src/main/java/org/apache/ignite/internal/processors/job/GridJobProcessor.java
```diff
@@ -898,21 +909,38 @@ private boolean removeFromPassive(GridJobWorker job) {
...
+                log.debug("Skipping handling collisions because node is stopping");
+
+            handlingCollisionFut.completeExceptionally(new NodeStoppingException("Node is stopping"));
 
             return;
         }
 
-        handlingCollision.set(Boolean.TRUE);
+        if (stopping) {
+            rwLock.readUnlock();
```
rpuch: Why do we need to have this `readUnlock()` here? A classic `take lock; try ... finally { unlock }` seems safer (as it is what everyone expects)

modules/core/src/test/java/org/apache/ignite/internal/processors/compute/ComputeJobChangePriorityTest.java
```diff
@@ -263,6 +260,10 @@ private static class NoopTask extends ComputeTaskAdapter<Void, Void> {
         }
     }
 
+    @Override protected long getTestTimeout() {
+        return 30_000;
```
rpuch: Why is this change needed?

rpuch: Oh, it's actually a decrease and not an increase. Should a comment explain this?

rpuch (approved):

PR #3613, opened 2026-04-10
GG-47843 [IGNITE-24224] [IGNITE-24913] Fixed message deserialization with SSL enabled

ibessonov:
https://ggsystems.atlassian.net/browse/GG-47843


Phillippko (approved):
Didn't dive into original fix logic, checked that nothing was lost, LGTM

PR #3624, opened 2026-04-23
GG-42930 DR. Control "dr pause" command only stops one sender from sending datа to replicated cluster

AMashenkov:


solveme (approved):
LGTM

PR #3627, opened 2026-04-29
GG-38584 [IGNITE-20730] Fix IgniteLock cannot be acquired after release on JDK 17

sk0x50:
https://ggsystems.atlassian.net/browse/GG-38584


ashapkin (approved):
LGTM

PR #3646, opened 2026-05-14
GG-47034 DR: Implement a run-time action to wipe out DR tree

AMashenkov:


modules/core/src/main/java/org/apache/ignite/internal/visor/dr/VisorDrResetTreeTask.java
```diff
@@ -0,0 +1,345 @@
...
+        }
+
+        /**
+         * Process task arguments and extract cache group id.
+         *
+         * @param arg      Task arguments.
+         * @param groupIds Group id collector.
+         * @param failures Failures collector.
+         */
+        private void extractGroupIds(VisorDrResetTreeTaskArgs arg, Consumer<Integer> groupIds,
```
solveme: This method seems to be unused, do we need it?

solveme (approved):

PR #3667, opened 2026-05-29
GG-48454 Java Thin Client: DataStreamer

tmgodinho:
https://ggsystems.atlassian.net/browse/GG-48454


ptupitsyn (changes requested):

ptupitsyn (changes requested):
@tmgodinho looks like backpressure mechanism is different from `DataStreamerImpl`. I would prefer to use the exact same logic, it is battle-tested.

Ideally we should reuse as much as we can with `DataStreamerImpl` (like GG9 reuses streamer logic across client and embedded), the only difference will be the batch send mechanism.

PR #3668, opened 2026-06-01
GG-42091 Add Http REST endpoint for k8s startup probe

valepakh:
## Summary

Add Kubernetes liveness/readiness/startup probe support to the `cmd=probe` REST
endpoint, dispatched on a new `kind` query parameter:

- **`kind=readiness`** — returns **503** while a (re)joining node is still pulling
  its partitions (inbound rebalance/demand) and **200** once its initial rebalance
  completes, then stays **200 (latched)** for the JVM lifetime so later
  topology-change rebalances don't flap an already-loaded node out of the Service
  endpoints. An INACTIVE cluster reports ready (nothing to rebalance); maintenance
  mode reports 503. The same endpoint doubles as a k8s `startupProbe`.
- **`kind=liveness`** and bare **`cmd=probe`** (unchanged, backward-compatible) remain
  the kernel-started check: 200 once the grid has started, 503 otherwise. A
  rebalancing node is alive, so these never block on rebalance/PME/state/maintenance.
- An unknown `kind` returns `STATUS_FAILED` with an explicit message.

This lets a rolling restart wait until each pod has finished inbound rebalance before
the next pod (a supplier) is bounced, avoiding the Lost-Partitions window that a
kernel-only probe leaves open.

## Jira

[GG-42091](https://ggsystems.atlassian.net/browse/GG-42091)

## Test plan

- [x] mtcga verification
- [ ] code review

🤖 Generated with [Claude Code](https://claude.com/claude-code)

[GG-42091]: https://ggsystems.atlassian.net/browse/GG-42091?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/modules/clients/src/test/java/org/apache/ignite/internal/client/rest/GridProbeCommandTest.java
+++ b/modules/clients/src/test/java/org/apache/ignite/internal/client/rest/GridProbeCommandTest.java
@@ -16,33 +16,49 @@
 
 package org.apache.ignite.internal.client.rest;
 
-import com.fasterxml.jackson.core.type.TypeReference;
-import com.fasterxml.jackson.databind.ObjectMapper;
 import java.io.IOException;
-import java.io.InputStreamReader;
-import java.net.HttpURLConnection;
-import java.net.URL;
 import java.util.Map;
 import java.util.concurrent.CountDownLatch;
+import org.apache.ignite.cluster.ClusterState;
 import org.apache.ignite.configuration.ConnectorConfiguration;
+import org.apache.ignite.configuration.DataRegionConfiguration;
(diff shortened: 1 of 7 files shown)


modules/clients/src/test/java/org/apache/ignite/internal/client/rest/GridProbeCommandTest.java
sk0x50: Could I ask you to fix a typo here `seperate` -> `separate`?

